PR #19, opened 2015-01-26
Made S3BotoStorage deconstructible so that Django can migrate it.

ramprabhuj:
Without this, Django 1.7 migrations fail with 
`ValueError: Cannot serialize: <storages.backends.s3boto.S3BotoStorage object at 0x1050a1358>
There are some values Django cannot serialize into migration files.
`

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -193,7 +193,7 @@ class S3BotoStorageFile(File):
                 self._multipart.cancel_upload()
         self.key.close()
 
-
+@deconstructible
 class S3BotoStorage(Storage):
     """
     Amazon Simple Storage Service using Boto


jschneier:
Hi and thanks for the pull request. Could you explain the use case for this a bit? As you can see [S3BotoStorageFile is already deconstructible](https://github.com/jschneier/django-storages-redux/blob/master/storages/backends/s3boto.py#L76-L77) and the way of using it that I am familiar with in models is to use that as your field. How do you use the ``S3BotoStorage`` object in your models file?

ramprabhuj:
You can specify 
`s3 = S3BotoStorage(params,...)
 image = models.ImageField(storage=s3,...)
`
Basically the Django FileField has a storage parameter and the user can pass any object that extends `Storage`

Without this PR the above code will fail in migration.

jschneier:
Ah interesting, wasn't familiar with that use case, thanks for the explanation.

PR #26, opened 2015-02-09
Fix exists and add modified_time in FTP storage

skirsdeda:
Apparently various FTP servers are inconsistent in NLST command output. Some output only filenames, others output paths. This fix supports both.

### DIFF
--- a/storages/backends/ftp.py
+++ b/storages/backends/ftp.py
@@ -185,7 +185,7 @@ class FTPStorage(Storage):
     def exists(self, name):
         self._start_connection()
         try:
-            if os.path.basename(name) in self._connection.nlst(
+            if name in self._connection.nlst(
                 os.path.dirname(name) + '/'
             ):
                 return True
--- a/storages/backends/ftp.py
+++ b/storages/backends/ftp.py
@@ -185,9 +185,10 @@ class FTPStorage(Storage):
     def exists(self, name):
         self._start_connection()
         try:
-            if name in self._connection.nlst(
+            nlst = self._connection.nlst(
                 os.path.dirname(name) + '/'
-            ):
+            )
+            if name in nlst or os.path.basename(name) in nlst:
                 return True
             else:
                 return False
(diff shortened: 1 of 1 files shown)


skirsdeda:
Also added **modified_time** function

storages/backends/ftp.py
```diff
@@ -165,6 +166,20 @@ def _get_dir_details(self, path):
...
 
+    def modified_time(self, name):
+        self._start_connection()
+        resp = self._connection.sendcmd('MDTM ' + name)
+        if resp[:3] == '213':
+            s = resp[3:].strip()
+            # workaround for broken FTP servers returning responses
+            # starting with e.g. 1904... instead of 2004...
+            if len(s) == 15 and s[:2] == '19':
+                s = repr(1900 + int(s[2:5])) + s[5:]
```
jschneier: let's you ``str`` instead of ``repr`` here.

storages/backends/ftp.py
```diff
@@ -236,7 +252,6 @@ def read(self, num_bytes=None):
         if not self._is_read:
             self._storage._start_connection()
             self.file = self._storage._read(self._name)
-            self._storage._end_connection()
```
jschneier: what was the logic of removing this?

jschneier:
Hi and thanks for your contribution. I'm excited that someone is using the FTP backend of this library. I've left a few comments inline and I'd be happy to merge this after they are addressed.

storages/backends/ftp.py
skirsdeda: function _end_connection() doesn't exist. With current code read() doesn't work at all because of this.

storages/backends/ftp.py
skirsdeda: Yeah, **str** would probably be a better option. I copied this from some path to python's ftplib which was never accepted.

storages/backends/ftp.py
jschneier: of course.

skirsdeda:
I'm actually using S3, SFTP and FTP. Was using some fork of django-storages-py3 but it's so messed up. Good that I finally found your repo. I'll send some improvements to SFTP backend as well when I get to it.

skirsdeda:
Changed repr() to str(). Should be good to pull now.

PR #29, opened 2015-02-17
Remove extra space

streeter:


jschneier:
That backend has now been deprecated and will be removed shortly. Thanks for looking out though!

PR #35, opened 2015-03-23
Fix libcloud provider type for GCS in Docs

pirsquare:
Otherwise we'll get `Unable to create libcloud driver type libcloud.storage.types.GOOGLE_STORAGE: Invalid module path` ...


jschneier:
thanks!

PR #36, opened 2015-03-31
Django 1.8+ compatibility

maryokhin:
As per the [deprecation note](https://docs.djangoproject.com/en/dev/releases/1.8/#support-for-the-max-length-argument-on-custom-storage-classes).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,15 @@ env:
   - TOX_ENV=py27-django15
   - TOX_ENV=py27-django16
   - TOX_ENV=py27-django17
+  - TOX_ENV=py27-django18
   - TOX_ENV=py33-django15
   - TOX_ENV=py33-django16
   - TOX_ENV=py33-django17
+  - TOX_ENV=py33-django18
   - TOX_ENV=py34-django15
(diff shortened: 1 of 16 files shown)


jschneier:
Thanks for this it looks great! My only concern is whether the ``super`` calls with no arguments will work on Python2.

maryokhin:
Yes, it should work because it is a [new style class](https://docs.python.org/2/reference/datamodel.html#new-style-and-classic-classes) + the build would have failed otherwise.

mbarrien:
Wrong, argumentless super() was introduced in Python 3, and never back ported. From a Python 2.7 REPL session:
```
>>> class Foo(object):
...     def func(self):
...         print "Foo"
...
>>> class Bar(Foo):
...     def func(self):
...         super().func()
...
>>> Bar().func()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in func
TypeError: super() takes at least 1 argument (0 given)
>>>
```

maryokhin:
Ok, true. I don't use Python 2, didn't expect it to still be such a mess. I'll fix it.

PR #37, opened 2015-03-31
Use cached properties where applicable

maryokhin:
Django already [provides](https://docs.djangoproject.com/en/1.7/ref/utils/#django.utils.functional.cached_property) a cached property implementation, so there is no need to do it yourself.


jschneier:
Thanks for your contribution. There is a difference between `cached_property` and `property`. One is a python builtin (the builtin use case of descriptors) and one is a django extension of it so that it's only evaluated once for expensive operations.

jschneier:
For things like `size` I don't think we want to use a `cached_property` because the file can change and then the new size won't be looked up.

maryokhin:
True, didn't think this through enough.

PR #39, opened 2015-04-06
rackspace cloud files updates for libcloud

django-emerge:
this pull request updates the libcloud interface.  

the specific updates are:

1. update liblcoud docs to specify correct rackspace driver `libcloud.storage.types.Provider.CLOUDFILES`
2. modify the url method to generate a temporary URL if container is private (as well as some new config options to support this)

hope this helps!

### DIFF
--- a/storages/backends/apache_libcloud.py
+++ b/storages/backends/apache_libcloud.py
@@ -126,12 +126,12 @@ class LibCloudStorage(Storage):
         obj = self._get_object(name)
         timeout = temporary_url_timeout if temporary_url_timeout else self.temporary_url_timeout
         if self.is_private_bucket or temporary_url:
-            return self.driver.ex_get_object_temp_url(obj, timeout)
+            return self.driver.ex_get_object_temp_url(obj, timeout=timeout)
         else:
             try:
                 return self.driver.get_object_cdn_url(obj)
             except ContainerDoesNotExistError:
-                return self.driver.ex_get_object_temp_url(obj, timeout)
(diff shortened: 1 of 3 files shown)


storages/backends/apache_libcloud.py
```diff
@@ -50,6 +50,8 @@ def __init__(self, provider_name=None, option=None):
                 "Unable to create libcloud driver type %s: %s" % \
                 (self.provider.get('type'), e))
         self.bucket = self.provider['bucket']   # Limit to one container
+        self.is_private_bucket = self.provider['is_private_bucket'] if 'is_private_bucket' in self.provider else None
```
jschneier: let's use `.get` for this

storages/backends/apache_libcloud.py
```diff
@@ -120,9 +122,16 @@ def size(self, name):
         else:
             return -1
 
-    def url(self, name):
+    def url(self, name, temporary_url=False, temporary_url_timeout=None):
```
jschneier: can we change these parameters to `temporary` and `timeout`? we're already part of the `url` method.

jschneier:
Thanks for your contribution, I've left some small code quality feedback but it looks good. Sorry for the delay in response I'd love to get this merged.

django-emerge:
Good feedback.  And those should be in now.

I also added a default for the `temporary_url_timeout` config, as if you omit that setting while working with a private container, the url method breaks trying to generate a temporary url.

litchfield:
We'd like to use django-storages, but need this-- any change we can get this merged?

PR #45, opened 2015-04-19
Improving azure storage

erlingbo:


storages/backends/azure_storage.py
```diff
@@ -57,9 +57,20 @@ def size(self, name):
         return properties["content-length"]
 
     def _save(self, name, content):
+        if hasattr(content.file, 'content_type'):
```
jschneier: This assumes we have a `ContentFile` right? I know that's what we return from `_open` but is this going to be the case generally?

jschneier:
Thanks for your contribution! I've left one concern of mine in line but looks nice other than that.

storages/backends/azure_storage.py
erlingbo: As i have interpreted the Django docs[1] a custom storage should handle File objects [2]. Thus it should always have a file attribute. This code checks whether the file has a content_type, if not it tries to guess it.

Please correct me if I have not interpreted the docs correctly ;)

[1] https://docs.djangoproject.com/en/1.8/howto/custom-file-storage/
[2] https://docs.djangoproject.com/en/1.8/ref/files/file/

jschneier: You are correct, sorry about that.

storages/backends/azure_storage.py
```diff
@@ -57,9 +57,20 @@ def size(self, name):
...
+            content_type = content.file.content_type
+        else:
+            content_type = mimetypes.guess_type(name)[0]
```
tbartelmess: I think you forgot to `import mimetypes`

erlingbo: Nice spot. I'll fix :)

pwwebdev:
Can someone merge this pull request?

Hanan-Natan:
Merge please.

cveilleux:
+1

pwwebdev:
Anyone with write access who can merge this?

jschneier:
Sorry for the delay folks, I have merged this. I realize that Azure storage probably wasn't really usable without that `url` fix in particular.

PR #51, opened 2015-06-12
Fixed error with GZIP settings for Google Storage

stmos:
GSBotoStorage inherits from S3BotoStorage.

In S3BotoStorage the gzip variable is called 'gzip' while GSBotoStorage was using is_gzipped.


jschneier:
Good catch! It's very annoying that this library was written with all of the settings vars at the class level, that makes it quite hard to mock things out to say the least.

I've opened up a PR at #60 with the same fix as you and a test, hope that's alright!

PR #54, opened 2015-06-24
Bugfix gs connection

patgmiller:
The `S3BotoStorage` class is overriding connection variables, this causes subclasses to use  `s3.amazonaws.com` as the host connection and fail authentication. 

Also reordered `Storage` class so it finds the overridden `get_available_name` instead of the one for super class.

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -278,14 +278,8 @@ class S3BotoStorage(Storage):
     def connection(self):
         if self._connection is None:
             self._connection = self.connection_class(
-                self.access_key,
-                self.secret_key,
-                is_secure=self.use_ssl,
+                self.access_key, self.secret_key,
                 calling_format=self.calling_format,
-                host=self.host,
-                port=self.port,
-                proxy=self.proxy,
-                proxy_port=self.proxy_port
             )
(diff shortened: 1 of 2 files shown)


patgmiller:
@jschneier have you looked at this yet? Is there anything you would like me to do in order to get it accepted?

jschneier:
Hi @patgmiller I understand the issue but this fix doesn't seem to actually use many of the `AWS_*` settings anymore, thus shattering backwards compat. Maybe what we need is to override the `@property connection`.

patgmiller:
@jschneier would you then prefer something more like this?

```python
@deconstructible
class GoogleStorage(S3BotoStorage): 

    @property
    def connection(self):
        if self._connection is None:
            self._connection = self.connection_class(
                self.access_key, self.secret_key,
                calling_format=self.calling_format,
            )
        return self._connection
```

dcgoss:
@patgmiller @jschneier I opened issue #124 not realizing it was a duplicate of this issue. My PR #125 should resolve the problem, in only one line.

jschneier:
Closed in favor of #125 (thanks both @dcgoss  and @patgmiller)

PR #55, opened 2015-06-25
Libcloud fixes

kaedroho:
This fixes a few issues I've encountered with the ``apache_libcloud`` backend (using Rackspace Cloud Files). These fixes follow the implemlementation of the S3 backend.

 - ``file.read(n)`` was returning the whole file regardless of the num_bytes parameter
 - Accessing ``file.name`` no longer gives an ``AttributeError``
 - Initialising the ``.file`` attribute to an empty ``BytesIO`` object caused errors when trying to load the file into ``Pillow``


jschneier:
urgh who wrote this garbage, thanks.

PR #63, opened 2015-07-14
Reverse mixin order, otherwise StorageMixin is ineffective.

agriffis:
StorageMixin is supposed to isolate the new `max_length` feature to
Django-1.8 and later, but it wasn't actually doing anything because the MRO
is backwards.


jschneier:
Thanks! I added a failing test case.

PR #66, opened 2015-07-24
New S3 Boto3 backend (closes #57)

### LINKED ISSUE #57: Boto3 Support
About a week ago, a new version of Boto was released. This new version fixes a number of shortcomings we've encountered with the current version of Boto, so it would be great to see this implemented into django-storages!

https://aws.amazon.com/blogs/aws/now-available-aws-sdk-for-python-3-boto3/

mbarrien:
This pull request implements https://github.com/jschneier/django-storages/issues/57 by adding a Boto 3 backend that tries to be a close-to-drop-in replacement for Boto 2. Due to the slight differences, I have kept it a separate backend, but with a lot of copy and paste code. Given that Boto 2 is heading towards maintenance mode according to https://github.com/boto/boto/commit/e3dd99695e8976fad88c1c55d69914199f1878db, I don't think it's worth trying to have 2 backends sharing code when the Boto 2 implementation looks on the way to being deprecated.

Because this is based on s3boto.py, you should be able to manually perform a diff of s3boto.py and the new s3boto3.py to understand the changes.

Changes:
* The AWS_HEADERS/storage.headers setting is replaced with AWS_S3_OBJECT_PARAMETERS/storage.object_parameters. The keys/values to this are intended to be the arguments to the http://boto3.readthedocs.org/en/latest/reference/services/s3.html#S3.Object.put boto3.Resource('s3').Object.put() method, which has all the arguments that would be used to generate the correct request headers. Note that this gives a little more access than just headers, since a user could also provide ACL and Content arguments in this dict.
* Unlike the s3boto implementation, s3boto3 backend does not currently support proxies (https://github.com/boto/botocore/issues/541), alternate calling formats, or alternate host/ports (https://github.com/boto/botocore/issues/601) because the underlying Boto3/Botocore library does not currently support it. It only kind of supports the endpoint URL
* If using s3v4, since botocore does not automatically redirect you to the correct region's endpoint nor sign properly unless it knows the region, there is an AWS_S3_REGION_NAME/storage.region_name setting to force the region.
* There's some behavior that was being done in the boto2 library that boto3 does not do, so equivalent code has been added to s3boto3 to do this. These include things like rewinding the file pointer, automatically checking for bucket and object existence, or directly writing the S3 contents to a file/file pointer. There's cases where Boto3 does not allow previous operations, like locally updating the last_modified attribute, which this new version performs by just reloading the object from the server.
* No need to parse timestamps, since boto3 already performs this conversion for you.

Known issues:
* For getting signed URLs via the .url() method there are multiple issues.
By default, most endpoints use the v1 URLs, and if you happen to add header parameters like response-content-disposition, Botocore provides the wrong signature. https://github.com/boto/botocore/pull/597
* For unsigned URLs (e.g. querystring_auth=False), Boto3 does not support this in its API (https://github.com/boto/boto3/issues/169). This implements compatibility by accessing some undocumented Boto3 internals to get the URL generated before it appends the signature. An alternate implementation I could be convinced of will generate the signed URL, but then strip off any signature parameters by parsing the querystring. Note that these parameter names will be different for s3v1/s3v4 URLs
* To support s3v4 URLs for endpoints that aren't v4 by default (most of them), there appears to be no way programmatically to force this yet. (https://github.com/boto/boto3/issues/168). The only way I found was to have a ~/.aws/config file as generated by running "aws configure set default.s3.signature_version s3v4". If deploying in a configuration where there is no user home directory, the location of this config can be set with AWS_CONFIG_FILE environment variable.
* Gzip compression and preload metadata are only barely tested.

This has been tested using s3v4 signatures with both signed and unsigned URLs, along with response content disposition and KMS server-side encryption key arguments.


mbarrien:
Closed in favor of #111

PR #70, opened 2015-08-13
django's File class expects an attribute 'name'

sixpearls:
Since django's ``File`` class expects a ``name`` atribute, I renamed the ``_name`` attribute to match.


jschneier:
Thanks!

PR #75, opened 2015-08-18
S3/Cloudfront Storage

hobarrera:
These changesets add a new class: S3CloudFrontStorage.

It's designed for uploading static files into S3, and have them served via CloudFront (which will need to be configured via the AWS control panel first).

This is ideal for both static files, and media files that will be served to many clients, as an alternative, eg: nginx.

---

Please let me know if there's something out of order with coding-style, or documentation. I've been using this quite a bit on a client project recently, and decided to strip it out, and submit it upstream (eg: here).

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -44,6 +44,21 @@ To allow ``django-admin.py`` collectstatic to automatically put your static file
 
     STATICFILES_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
 
+
+CloudFront
+~~~~~~~~~~
+
+If you're using S3 as a CDN (via CloudFront), you'll probably want this storage
+to serve those files using that::
+
+    AWS_S3_CUSTOM_DOMAIN = 'cdn.mydomain.com'
+
+Keep in mind you'll have to configure CloudFront to use the proper bucket as an
+origin manually for this to work.
+
+If you need to use multiple storages that are served via CloudFront, pass the
+`custom_domain` parameter to their constructors.
+
 Fields
 ------
 


taxido:
I'm all for this change.

hobarrera:
I've a somewhat hacky change on my local source that appends a querystring with a configurable value/function to the URLs.

I use this to force the CDN to re-fetch cached resources (I pass the app-version as a that appended value).

In other words, `{% static "somefile.css" %}` becomes `http://cdn.example.com/static/somefile.css?v=APP_VERSION`.

This is so after updating and re-running `collectstatic`, new requests bypass the CDN's cache.

Should I add this to the PR, or is it too hacky?

comandrei:
I think you can set AWS_S3_CUSTOM_DOMAIN to your cloudfront FQDN and obtain the same functionality.

Please see https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto.py#L479

hobarrera:
Huh, it would seem so. Please don't close this PR though; I'll adapt it so use that, make that parameter configurable, and document it properly. :D

hobarrera:
Looks like force-pushing closes an issue. I was unaware github did that.

I've rewritten this branch so as to merely document the existing capability. Thanks for pointing that out, @comandrei.

comandrei:
Thanks @hobarrera !

jschneier:
Sorry for the delay, merged in 7f2d476ebcd3972ae1277acd92216c52367ac2cf.

hobarrera:
Why was the author for this commit changes from myself to @jschneier before merging it?

jschneier:
It wasn't applying cleanly so I copy/pasted it out in lieu of rebasing. Mea culpa.

jschneier:
Okay, rebased it through with you. 65c9f72d51d6aa540e9089979f179db78dd9c13b

hobarrera:
Thanks. I was just curious, mostly.

PR #81, opened 2015-09-20
Dropbox fix and improvement

ZuluPro:
Made some fix for compatibility with others backend.

### DIFF
--- a/storages/backends/dropbox.py
+++ b/storages/backends/dropbox.py
@@ -27,12 +27,12 @@ class DropBoxStorageException(Exception):
 
 
 class DropBoxFile(File):
-    def __init__(self, name, storage):
+    def __init__(self, name, storage, mode='rb'):
         self.name = name
         self._storage = storage
 
     def read(self, num_bytes=None):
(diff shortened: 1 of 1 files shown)


jschneier:
This looks good (especially not returning from `read`, yeesh), what's the logic behind removing the `BytesIO` call?

On a side note I went spelunking through the dropbox api client code and there is an incredible amount of abstraction over what is essentially a thin wrapper on urllib3. It gets a bit much I think.

ZuluPro:
@jschneier:

>  what's the logic behind removing the BytesIO call?

Do not really remember why, but I tested manually it's works.

>  On a side note I went spelunking through the dropbox api client code and there is an incredible amount of abstraction over what is essentially a thin wrapper on urllib3. It gets a bit much I think.

I plan to add DropBox Driver to Libcloud, and in this lib all is made ± from raw. Would you prefer to use it in future ?

ZuluPro:
Any news ?

jschneier:
Thanks for staying on top of me on this.

PR #82, opened 2015-09-20
Moved tests into storages/tests

ZuluPro:
Like explain in many books tests should be included inside project.
It is very useful for import Redux's storage utils in other project.

I remove py.test for use Django native test framework and all settings in `tests/runtests.py` with a launcher for test.


ZuluPro:
Any news ?

jschneier:
hi. i'm not opposed to this in principle but i'm just wondering why you think it's necessary? i look at projects such as getsentry/sentry, django-taggit, django-rest-framework or django-localflavor and they all have the tests at the project level, next to the main directory.

ZuluPro:
@jschneier 
For example, in dropbox backend, I use some fixtures for my tests and would like to reuse it for an other application's tests.

Some example of lib which have tests inside code:
OpenStack and all its project make this.

ZuluPro:
Any news ?

ZuluPro:
@jschneier: Any news ?

ZuluPro:
@jschneier : Rebased from master !

ZuluPro:
@jschneier Any news ?

jdufresne:
FWIW, I disagree with this change.

Most projects I work on keep tests outside of the library. I find the organization of tests at the top level more conventional.

The library code is intended to be consumed by the other projects integrating django-storages with their code. Contrast with the library tests which are intended to be consumed by the library developers. As they serve a different audiences, it seems like a good idea to keep them separate.

> For example, in dropbox backend, I use some fixtures for my tests and would like to reuse it for an other application's tests.

Tests and tests fixtures are not a public API and could change without warning as the library evolves. Unlike the library itself, test fixtures have no backwards compatibility concerns. Relying on this will eventually cause future frustration.

jleclanche:
I also disagree with the change for very similar reasons.

PR #89, opened 2015-10-22
Make LibCloudStorage thread-safe

kaedroho:
Django creates a single, global instance of LibCloudStorage that is shared between threads. This instance uses the same ``Driver`` object for all operations.

The ``Driver`` object is not thread-safe though, which can lead to errors being raised if two threads performed an operation at the same time. Moving the initialisation of the ``Driver`` object into a property should solve this as each operation will be given its own ``Driver`` object.


kaedroho:
Closing as I'm still experiencing issues after deploying this fix.

PR #92, opened 2015-11-03
Update auto-sign private links azure_storage.py

robertavram:
Update for the azure_storage backend to accommodate for an auto-sign flag that enables signing of the links that are stored privately.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 import os.path
 import mimetypes
 import time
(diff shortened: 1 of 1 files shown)


jschneier:
Hi and thank you, this looks okay although I have a few minor nits which I will leave feedback on.

Also, I'm planning on adding support for only SAS tokens (not requiring account keys) shortly so this will conflict with that.

storages/backends/azure_storage.py
```diff
@@ -98,10 +101,25 @@ def _save(self, name, content):
 
     def url(self, name):
         if hasattr(self.connection, 'make_blob_url'):
+            if self.auto_sign:
+                access_policy = AccessPolicy()
+                access_policy.start = (datetime.utcnow() + timedelta(seconds=-120)).strftime('%Y-%m-%dT%H:%M:%SZ')
```
jschneier: I think we should just leave start to its default since that is when the service receives the blob which seems reasonable to me.

storages/backends/azure_storage.py
```diff
@@ -98,10 +101,25 @@ def _save(self, name, content):
 
     def url(self, name):
         if hasattr(self.connection, 'make_blob_url'):
+            if self.auto_sign:
+                access_policy = AccessPolicy()
```
jschneier: Let's just pass in the policies as kwargs:
```python
AccessPolicy(
    permission=self.policy_permission,
    expiry=datetime.utcnow() + timedelta(seconds=self.policy_expiration)
)
```

jschneier:
Ah sorry, after looking at this I'm going to update Azure to only support the 0.3x series. We can get this feature merged there if that interests you. The changes are fairly small.

sww314:
Fixed in #565

PR #93, opened 2015-11-05
Handle opening of non-existant files better in libcloud storage

kaedroho:
Currently, the `LibCloudStorage._read` method raises the following exception when the file doesn't exist:

```
'NoneType' object has no attribute 'size'
```

This is because the `obj = self._get_object(name)` line is returning `None`, but no check is made on `obj` before using it.

This commit changes this behaviour so it now checks if `_get_object` returned `None` and instead raises a `LibCloudFileNotFounderror` (a subclass of `IOError`).

The file is now downloaded when it is opened. This is so the exception is raised where the developer would expect it

### DIFF
--- a/storages/backends/apache_libcloud.py
+++ b/storages/backends/apache_libcloud.py
@@ -143,16 +143,28 @@ class LibCloudStorage(Storage):
 
     def _open(self, name, mode='rb'):
         remote_file = LibCloudFile(name, self, mode=mode)
+
+        # Trigger a download of the file
+        # this will raise an exception of the file doesn't exist
+        remote_file._get_file()
+
         return remote_file
 
     def _read(self, name):
         obj = self._get_object(name)
+        if not obj:
+            return None
+
         # TOFIX : we should be able to read chunk by chunk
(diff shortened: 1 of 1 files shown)


jschneier:
Hi, thanks for the report and sorry for the late reply.

This is interesting and this whole backend is fairly janky. Your solution is certainly safer from a backwards compatability perspective but I'm not convinced it's the best way forward.

What do you think about instead doing the following:

```python
def _get_object(self, name):
    """Get object by its name. Raises libcloud.storage.types.ObjectDoesNotExistError None if object not found"""
    clean_name = self._clean_name(name)
    return self.driver.get_object(self.bucket, clean_name)

def exists(self, name):
    try:
        self._get_object(name)
        return True
    except ObjectDoesNotExistError:
        return False
```

And etc as we change the various methods to account for this. I don't think it really makes sense for `.size` to return -1 if the file doesn't exist, this is not a C library.

kaedroho:
@jschneier I prefer that, it's much cleaner. Thanks!

sww314:
2 years old. Closing

PR #94, opened 2015-11-11
use encryption setting for S3BotoStorageFile multipart uploads

andersontep:
S3BotoStorageFile's write method was not initiating the multipart upload with encryption on, even when ```AWS_S3_ENCRYPTION``` was True in the settings


andersontep:
@jschneier do you mind taking a look when you have a moment? Thank you!

jschneier:
This looks good, thanks.

PR #96, opened 2015-11-17
Optionally override querystring_expire in S3BotoStorage.url

mattbriancon:
It's often inconvenient to instantiate a new instance of `S3BotoStorage` just to change the expiration of URLs.  Instead I'd like to (optionally) pass the new expiration with each call to `S3BotoStorage.url`.

```python
pic = MyModel.objects.get(...).picture
pic.storage.url(pic.name)  # default expiration
pic.storage.url(pic.name, expire=600)  # custom expiration
```

Happy to make adjustments or take another approach. Thanks!

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -473,17 +473,26 @@ class S3BotoStorage(Storage):
         # Parse the last_modified string to a local datetime object.
         return parse_ts(entry.last_modified)
 
-    def url(self, name, headers=None, response_headers=None):
+    def url(self, name, headers=None, response_headers=None, expire=None):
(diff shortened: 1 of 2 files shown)


mattbriancon:
@jschneier any thoughts here? Hoping to get this in quickly. Thanks!

jschneier:
this is fine. It's annoying that boto takes this one arg as a position arg and everything else as kwarg but that's the api i suppose. thanks for writing a test.

avinassh:
When this will be released in PyPi version? Last version was released 5 months ago (looking at the last timestamp on `setup.py`)

PR #100, opened 2015-12-02
Enable GZipping of SVG files automatically in S3Boto Storage

comandrei:
SVG images are mostly text and they compress really well, this should improve performance for SVG static files


jschneier:
Thanks!

PR #97, opened 2015-11-17
Update Travis

mattbriancon:

PR #101, opened 2015-12-04
enable automated testing on django1.9

nikolas:


browniebroke:
Wasn't it covered in https://github.com/jschneier/django-storages/pull/97 ?
It's also testing against supported versions of Python by Django, so it seems better to me.

nikolas:
Yeah, you're right! I missed that.

PR #106, opened 2015-12-21
Fix memory leak, close SpooledTemporaryFile

kmmbvnr:


jschneier:
This looks good. How did you discover the leak (I want to see if I can write a test). Also, the same code (basically) exists on line 174 + 175 and I'm not able to puzzle out why it's there, any thoughts on that?

kmmbvnr:
We just run out of memory when getting files over OurModel.objects.filter(...) Django queryset caches all model instances, thats means even with restriction `storage.max_memory_size` it could takes too much memory.

I didn't remove lines 174 + 175 b/c `_flush_write_buffer` called also at line 153 (but i have no ideas, why someone would like to close file on a write call)

jschneier:
Okay. I'll remove those lines after this, I don't think they make any sense. Thanks for the patch.

PR #120, opened 2016-02-11
Freeze modification time when gzipping files to S3

millarm:
to allow detection of changes in gzipped files by MD5 SUM in the same
way as for non gzipped files


jschneier:
Thanks. Good idea.

PR #122, opened 2016-02-18
Adds support for already gzipped content to S3 storage

cambonf:
This adds support to save to S3 data that is already gzipped with the proper Content-Encoding
In my use case, I need to be able to upload data that is already gzipped to S3.

Could it be merged? Do you see any possible side effect?


jschneier:
Looks great, merged in 12ceed1dc822ce960e02176b6c7dce9f6e08a0e0.

PR #128, opened 2016-03-10
Update amazon-S3.rst

taifu:
THIS IS DANGEROUS!!! Please, please, please document throughly that overwriting the file is now the default behaviour. Fix the doc or change the default!!!
I was bitten badly by this one... :-(

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -135,7 +135,7 @@ Files can be read in a little at a time, if necessary::
     >>> '-'.join(obj1.normal.chunks(chunk_size=2))
     'co-nt-en-t'
 
-Save another file with the same name::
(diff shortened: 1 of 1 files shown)


jschneier:
Has this changed? I agree it should be documented but I believe that was always this default.

taifu:
Django default has always been this (appending and "_1", "_2", now append a random number): 

http://django.readthedocs.org/en/latest/ref/files/storage.html#django.core.files.storage.Storage.get_available_name

You are right, I don't know if s3boto.py changed behaviour.

I think that should comply to other django default storage but as I don't want to break backward compatibility I made a pull request to change the (wrong) documentation.

jschneier:
I added some docs for this [here](50cbaa9c8a7c7f8825731ecf5d4ce1e79663a25f).

Issue #138, opened 2016-04-07
dropbox backend does not overwrite existing files


epetxepe:
This backend has no similar option to "file_overwrite = setting('AWS_S3_FILE_OVERWRITE', True)" as S3boto does.

I've overcome this issue by updating _save method with solution at http://stackoverflow.com/questions/31909627/dropbox-files-not-being-overwritten

```
def _save(self, name, content):
        #name = name.replace ('\\','/')     #137 
        file_metadata = self.client.metadata(name)
        the_rev = file_metadata['rev']
        self.client.put_file(name, content, overwrite=True, parent_rev=the_rev)
        return name
```

I think it would be interesting to add something similar for this backend

jschneier:
Sure! Would you be interesting in creating a pull request for this functionality?

Also, from that Stackoverflow link it looks like dropbox will ignore the write if the content is the same. I don't think it's that important for now but something to keep in mind.

riccardodivirgilio:
+1

jschneier:
Fixed in #873.

Issue #137, opened 2016-04-07
dropbox backend fails with "backslash not allowed" error since get_available_name method is not overrided


epetxepe:
Hi all, 

  I'm using django version 1.9, python 2.7. 

  I've set up dropbox backend and configured the Oauth token. I'm using django-storages for a FileField of my model.

  I'm getting:

```
ErrorResponse: [400] {u'path': u"Invalid path '/biocloud\\b2ce86b5-dffb-4752-b7c3-c9c27d9d40fd\\18.jpg': character at index 9: backslash not allowed"} 
```

  I've tried to convert the paths at django model level, but it gets converted back when django starts the saving process.

  I'm using S3 backend as an alternative, so I've done some debugging on how django actually executes both.

  S3 backend for django-storages overrides get_available_name, and does the backslash => slash conversion.

  dropbox for django-storages does not override get_available_name and the default django method, appends some alphanumeric if already exists or truncates path if it is > max_length. So no conversion is performed.  And return force_text(name.replace('\', '/')) at the end of django's storage.py save() is only applied to return a name to be saved in the model, but not to be send to the dropbox storage backend logic for the overrided "_save ()" function.

 Right now I've overcome this issue by adding "name = name.replace ('\','/')" at the beginning of dropbox backend "_save ()" function but I don't think is the best option and I think that get_available_name should be overrided in the dropbox backend.

Regards,

Gorka

jschneier:
Hi and thanks for the report! Would you be interested in creating a pull request to fix this issue?

jschneier:
I think this is fixed by e52a127523fdd5be50bb670ccad566c5d527f3d1.

PR #140, opened 2016-04-12
Fix of issues #137 and issue #138

epetxepe:
- #137 I've placed everything on the _save function given that "get_available_name" was not being called when file is written with open () method.
- #138 No file_overwrite variable is considered. Overwrite is forced, if the file has the same content, dropbox keeps the previous one so no overwrite is reflected.

### DIFF
--- a/storages/backends/dropbox.py
+++ b/storages/backends/dropbox.py
@@ -90,7 +90,13 @@ class DropBoxStorage(Storage):
         return remote_file
 
     def _save(self, name, content):
-        self.client.put_file(name, content)
+        name = name.replace ('\\','/')
+        try:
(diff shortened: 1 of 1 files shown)


zkhan93:
I am using v1.5.2, facing #137 what should I do? I think this pull request should be merged.

jschneier:
@zkhan93 I just pushed a fix to #137, can you try that out?

jschneier:
@zkhan93 that is, install from git as is covered in the README.

zkhan93:
@jschneier it is working fine thanks 👍

sww314:
Closing stale requests.

PR #84, opened 2015-09-24
fixing timezone comparison in s3

marcelchastain:
originally by @ryanbagwall on bitbucket
https://bitbucket.org/david/django-storages/pull-requests/102/fixing-timezone-comparison-in-s3/diff

Fixes #49

PR #141, opened 2016-04-14
Correct collectstatic time modified detection for S3

caleb-allen:
This will allow collectstatic to correctly upload the most recently modified file to S3.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,10 +1,3 @@
-==============
-Caleb's fork
-==============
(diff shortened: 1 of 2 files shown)


caleb-allen:
Looks like pr #84 achieves the same thing and handles it better.

jschneier:
Django 1.10+ `get_modified_time` handles this gracefully and in a way that we never quite managed to achieve so I'm going to let this one go.

caleb-allen:
Glad to see a graceful solution

PR #142, opened 2016-04-14
Adding Option to Amazon S3 Backend Docs

a12k:
Adds to the S3 document the documentation for the ability to change from the S3 django-storages default of 'public-read' to 'private', which is the standard Amazon S3 default.


jschneier:
Thanks, merged in eca46e509843f39137843441c33f1c51812a909c

PR #145, opened 2016-04-25
Be more precise about recursive includes.

fladi:
The tarball on PyPI has some artifacts in the `tests` folder right now (`.pyc`
files) which should not be shipped. Changing the directives to be more precise
as to what gets includes fixes this for future release tarballs.


jschneier:
Thanks. I swear, each project has at least 3 "fixing" `MANIFEST.in` commits.

PR #146, opened 2016-04-27
Google Cloud Storage backend using the gcloud-python library

eirsyl:
This is kind a work in progress and is just a proposal. Is a Google Cloud Storage backend using gcloud-python something this project needs? 

The gcloud-python package supports multiple types of authentication:
- Application running in Google App Engine
- JSON or PKCS12/P12 keyfile pointed to by GOOGLE_APPLICATION_CREDENTIALS environment variable
- Credentials provided by the Google Cloud SDK (via gcloud auth login)
- Application running in Google Compute Engine

The code is not perfect and needs some small changes before production. Any thoughts on this?

### DIFF
--- a/storages/backends/google.py
+++ b/storages/backends/google.py
@@ -1,0 +1,241 @@
+import mimetypes
+from tempfile import SpooledTemporaryFile
+
+from django.core.exceptions import ImproperlyConfigured
+from django.core.files.base import File
+from django.utils.encoding import force_bytes, force_text, smart_str
+from storages.compat import Storage
+from storages.utils import setting
+
+try:
+    from gcloud.storage.client import Client
+    from gcloud.storage.bucket import Bucket
+    from gcloud.storage.blob import Blob
+    from gcloud.exceptions import NotFound
+except ImportError:
+    raise ImproperlyConfigured("Could not load Google Storage bindings.\n"
+                               "See https://github.com/GoogleCloudPlatform/gcloud-python")
+
+
+class GoogleCloudFile(File):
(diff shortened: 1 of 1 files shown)


scjody:
@eirsyl I'm interested in this and may have some time to contribute towards getting this mergeable. What's on the TODO list other than improving test coverage?

There are a couple of other ways to use Google Cloud Storage as a Django storage backend, but neither of them are great:
* django-storages supports libcloud, which can access Google Cloud Storage but only in S3 compatibility mode. S3 compatibility mode is only designed for use during short term migrations, and requires per-developer keys that don't allow fine-grained access control.
* [https://pypi.python.org/pypi/django-gcloud-storage](django-gcloud-storage) says it's an alpha and explicitly warns against using it in production.

eirsyl:
The first draft is heavily inspired by the S3 implementation, we should probably spend some time understanding the gcloud library to make this implementation right. An like you said, tests and documentation is also a required to make this mergable.

scjody:
OK, I'll read up on the gcloud library and review the code with that in mind. I'm in the middle of another major project but I can work on this when that's blocked.

scjody:
Interestingly, there's also [gs.py](https://github.com/jschneier/django-storages/blob/master/storages/backends/gs.py). I can't find any documentation for it, and it seems to require S3 compatibility mode so I can't use it in any case.

scjody:
I had to rename the module and change the imports since Google renamed their library. I'm working on [my own PR](https://github.com/jschneier/django-storages/pull/236) since I don't see a way to collaborate on this one.

eirsyl:
Closing this in favor of #236

PR #149, opened 2016-05-18
Is not a fork anymore

camilonova:
This was needed when the name was django-storages-redux now is not the case.


jschneier:
Fair enough :)

PR #152, opened 2016-05-31
Added AWS_EXTRA_HEADERS option

darkowic:
We found it useful to add special headers for files from specific path (e.g. user avatars)
I added AWS_EXTRA_HEADERS and created get_headers method for S3BotoStorage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ setuptools*
 __pycache__
 .coverage
 .cache
+.idea
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -40,6 +40,16 @@ If you'd like to set headers sent with each file of the storage::
         'Cache-Control': 'max-age=86400',
     }
 
+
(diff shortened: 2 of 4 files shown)


darkowic:
Will resolve conflicts soon.
Also there is one problem - this only works with save and not with write method so I have to find solution for this.
@jschneier what do you think about this feature?

darkowic:
hey @jschneier ! I see that you are checking recently on this project.  I will have time to fix this feature and implement it for boto3. What do you think about it?

jleclanche:
@darkowic I got commit access -- I'll take you up on the offer for reimplementing this for the s3boto3 storage instead; ping me once you do and I can review it!

darkowic:
@jleclanche there is another idea for it
https://github.com/jschneier/django-storages/issues/231

What do you think? Btw I have to admit that I don't have much time recently to do this but maybe someone from my company can take care of it. We just need to decide how we want to implement it.

sww314:
Closing in favor or #231

PR #159, opened 2016-06-10
Fix bug in SFTPStorage._save().

sam-at-github:
SFTPStorage._save() is only ever called on closing a SFTPStorageFile. SFTPStorageFile passes this method bytes, but SFTPStorage._save() was treating the parameter as a file.


jschneier:
Hey, thanks for this. This was I believe fixed in #177 via a slightly different tweak. Let me know if you're still seeing issues and sorry for the late response.

PR #167, opened 2016-07-04
Raise NotImplementedError if attempting to open a file with mode 'a'

rapilabs:


jschneier:
Interesting, I don't think `a` is cross-platform compatible, at least the Python3.5 docs certainly seem to indicate that is not. Did you run into this and expect certain functionality? I don't think append-mode makes a lot of sense for remote files in general.

Issue #160, opened 2016-06-12
Problem with S3BotoStorageFile and a file greater than 5Mb


iceseyes:
My team and I have found a problem in file-like interface for S3botoStorageFile.

If I do something like that:

```
 storage = S3BotoStorage()
 with storage.open("test.csv", "w") as f:
     df.to_csv(f)
```

where `df` is a Pandas' Dataframe, and file is greater than 5Mb, when we run the program from an ec2 instance, program stops over a write instruction.

Debugging we have seen that when file size become 5Mb the test in s3boto.py:159

```
if self.buffer_size <= self._buffer_file_size:
     self._flush_write_buffer()
```

is `True` for every write request, so for every write I have a request to upload to s3.... and after a few of seconds program freezes.

I think that test should be something like that

```
if (self._write_counter * self.buffer_size) <= self._buffer_file_size:
     self._flush_write_buffer()
```

Thank you.

timur-orudzhov:
We've also encountered with that issue and pr, that provided by tommwatson, fixed that issue. In our case, we open local file that weight about 13 mb and write to file by chunks (~ 1mb), that on s3, as a result we have file with weight about 76 mb. How to reproduce:

```
from django.core.files.storage import default_storage

def read_in_chunks(file_object, chunk_size=1024 * 1000):
    """Lazy function (generator) to read a file piece by piece.
       Default chunk size: 1mb.
    """
    while True:
        data = file_object.read(chunk_size)
        if not data:
            break
        yield data


with open('/path/to/file', 'r+b') as local_file:
    with default_storage.open('path/to/s3_file', mode='wb') as s3_file:
        for chunk in read_in_chunks(local_file):
            s3_file.write(chunk)
```

Are there chances to merge fix i've mentioned earlier ?

PR #169, opened 2016-07-08
Truncate SpooledTemporaryFile after flushing the buffer

jiinus:
When uploading a file in chunks with S3BotoStorageFile I noticed that the underlying SpooledTemporaryFile is never truncated but instead successive write-calls just keep appending the file. 

With each flush the file is seeked to the beginning though which results in malformed data on S3 as the contents of the previous flushes are uploaded again and again on each write.

Also added the missing S3.py. License seems to permit it?

### DIFF
--- a/storages/S3.py
+++ b/storages/S3.py
@@ -1,0 +1,618 @@
+#!/usr/bin/env python
+
+#  This software code is made available "AS IS" without warranties of any
+#  kind.  You may copy, display, modify and redistribute the software
+#  code either by itself or as incorporated into your code; provided that
+#  you do not remove any proprietary notices.  Your use of this software
+#  code is at your own risk and you waive any claim against Amazon
+#  Digital Services, Inc. or its affiliates with respect to your use of
+#  this software code. (c) 2006-2007 Amazon Digital Services, Inc. or its
+#  affiliates.
+
+import base64
+import hmac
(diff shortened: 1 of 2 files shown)


jiinus:
This should also fix the problem mentioned in issue #160

jschneier:
Thanks to you and @iceseyes for bringing this issue up. I'm going to write a failing test with a 5MB file and get this one fixed asap, probably with your fix.

I intentionally removed the S3.py file because it was deprecated in the original repo and I wanted to reduce my own maintenance burden, especially of already deprecated pieces of the library. New fork yadda yadda.

m-barthelemy:
Hi there,

This is affecting us badly as well (from a 45MB file we end up with >15GB on S3 before Django give up).
Is there anything that could be done to help getting this fixed?

Thanks!

todofixthis:
Possibly related:
- https://github.com/jschneier/django-storages/commit/5e14712e3532ae676bd207f213cc8e4bfcaa7a3b
- https://github.com/jschneier/django-storages/pull/106

danpalmer:
Hi, ping on this? We have just encountered an issue where uploading ~10MB of data resulted in ~20GB of data being written to S3 - it _looks_ like this PR is the fix we need. Any possibility of a merge?

PR #172, opened 2016-07-13
Added info about AWS_S3_ENCRYPTION.

fala:
The docs were missing info about param responsible for enabling server-side encryption.


jschneier:
Thanks

PR #173, opened 2016-07-13
Add IAM role credentials support for Boto S3 backend

### LINKED ISSUE #34: Add support for IAM role credentials to Boto S3 backend
If access key and secret key credentials are not provided via the Django settings, or the environment, it would be helpful if the Boto S3 backend automatically tried to retrieve authentication credentials from instance metadata.

hanleyhansen:
This changeset adds IAM role credentials support to the Boto S3 backend
by duplicating a bit of what Boto already does under the covers (since
version 2.5.x or so).

The new credential retrieval process follows this order:

1. Django settings
2. Environment variables
3. IAM role credentials via instance metadata

Attempts to resolve #34.

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -14,7 +14,8 @@ try:
     from boto.s3.connection import S3Connection, SubdomainCallingFormat
     from boto.exception import S3ResponseError
     from boto.s3.key import Key as S3Key
-    from boto.utils import parse_ts, ISO8601
+    from boto.compat import six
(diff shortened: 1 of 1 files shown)


hanleyhansen:
@jschneier and @lmorchard excuse the forwardness but this is awesome work that we really need.

mbarrien:
This isn't needed. When None is passed to the underlying boto library for access key and secret, boto internally looks up iam credentials. Django-storages never looks it up itself, hence why you don't see the code here.

PR #179, opened 2016-08-02
S3boto3

jschneier:


storages/backends/s3boto3.py
```diff
@@ -0,0 +1,566 @@
...
+    data is only the data contained in the current buffer. So you
+    should not access the contained file object directly. You should
+    access the data via this class.
+
+    Warning: This file *must* be closed using the close() method in
+    order to properly write the file to S3. Be sure to close the file
+    in your application.
+    """
+    # TODO: Read/Write (rw) mode may be a bit undefined at the moment. Needs testing.
+    # TODO: When Django drops support for Python 2.5, rewrite to use the
```
galuszkak: @jschneier @mbarrien as far I know Django dropped Python 2.5 and 2.6. Probably this needs to be done...

ZuluPro:
Hello @jschneier 
I'm looking to how to make boto & boto3 storage more DRY,
And in fact we don't need boto3: https://aws.amazon.com/fr/blogs/aws/boto-python-3/
boto supports Python3 !

Are you agree to remove it ?

mbarrien:
@ZuluPro On the very page you link, they link to Boto's documentation at http://boto.readthedocs.io/en/latest/ which says this multiple times on the page:

"Note Boto3, the next version of Boto, is now stable and recommended for general use. It can be used side-by-side with Boto in the same project, so it is easy to start using Boto3 in your existing projects as well as new projects. Going forward, API updates and all new feature work will be focused on Boto3."

Given that, I don't think dropping Boto 3 and just going with original Boto is a viable option in the long term, even with Python 3 support.

PR #180, opened 2016-08-02
Allow boto3 auth via IAM roles

JshWright:
boto3 can automatically grab its own credentials when running on an ec2
instance that has been assigned an IAM role. This results in seamless
key rotation, etc.

In order to support that, the credentials must not be passed as kwargs
to the connection constructor.


mbarrien:
We've been running production for months using IAM roles without needing the change from this pull request. The storages.utils.settings already defaults to None for access_key/secret_key if the settings is missing, and it's okay to provide the aws_access_key_id/aws_secret_access_key explicitly as None to boto3.resource()

JshWright:
Interesting... We were seeing a credential issue. I'll take a closer look at our configs

JshWright:
Yep, looks like this was a partial setting name conflict (one name conflicted, the other didn't) on our end.

PR #181, opened 2016-08-03
Fix TZ awareness in S3Boto3Storage

JshWright:
Django expects a naive datetime from the [modified_time()] method.
Django 1.10 adds a [get_modified_time()] method which respects USE_TZ.

This commit fixes modified_time(), adds get_modified_time(), and adds
relevant tests.

[modified_time()]: https://docs.djangoproject.com/en/1.10/ref/files/storage/#django.core.files.storage.Storage.modified_time
[get_modified_time()]: https://docs.djangoproject.com/en/1.10/ref/files/storage/#django.core.files.storage.Storage.get_modified_time


JshWright:
Looks like the behavior of timezone.make_naive changed a bit from Django 1.7->1.8. I'll fix that and update this PR

jschneier:
Thank you. Guess I need to update...all of the backends for the new Django 1.10 API :p

PR #182, opened 2016-08-03
Test against Django 1.10

jayfk:


jayfk:
Don't know why tox is unable to recognize `py27-django110` :/. Seems to be working here: https://travis-ci.org/tomchristie/django-rest-framework/jobs/149483626

jschneier:
Hey I think you need to define `django110` in `tox.ini` [like DRF](https://github.com/tomchristie/django-rest-framework/blob/master/tox.ini#L20). Very happy to merge once this is working.

jschneier:
Fixed in #185, thanks for bringing to my attention!

PR #188, opened 2016-08-12
Implement listdir on Azure

mheinze:
Azure is a different since it uses containers instead of directories. This implementation will list the blobs that match the prefix.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -116,7 +116,7 @@ class AzureStorage(Storage):
         )
         results = []
         for path in blobs:
-            results.append(path.name)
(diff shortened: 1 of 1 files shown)


jschneier:
Eh, S3 is nominally the same (no directories) but there are MSDN articles about having paths (using `/`) and we claim S3 has directories. The combination of `prefix` and `delimiter` (which I think we simply force to be `/`) should be good enough for how people think of these things.

jschneier:
Although I think from some more reading I'm not sure if `delimiter` is very useful for us.

sww314:
Azure backend was rewritten in #565. Please reopen if the issue still exists.

PR #193, opened 2016-08-26
All storage classes are deconstructible

mnach:
Dear Josh,

I start using this library recently, and want to thank you for maintaining it! And also thangs to all authors and participants for their contribution!

I found that some of the storages doesn't decorated by `@deconstructible`. But this is required in case of passing storage as a parameter into FileField.

Can it be included in future versions?


jschneier:
Thanks!

PR #195, opened 2016-08-30
S3Boto3Storage: fix for file-like objects without name, multipart upload support

pacahon:
In fact, it's partially solves problem. If we pass subclass of `File` class and forgot override `__bool__` method - an error based on empty name still can occur.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,7 @@
+[pytest]
+django_find_project = false
+norecursedirs = .tox .git .idea docs storages
+python_files=test_*.py tests.py
+env =
+    DJANGO_SETTINGS_MODULE=tests.settings
+    PYTHONDONTWRITEBYTECODE=1
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -1,4 +1,5 @@
 Django>=1.7
+pytest-env
 pytest-cov==2.2.1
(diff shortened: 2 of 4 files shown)


linuxlewis:
👍 on seeing the SHA error. We've worked around it by ensuring we pass a `django.core.files.base.File` object with `name` defined.

pacahon:
@linuxlewis Hi, can you show me where it's already fixed?

linuxlewis:
@pacahon The "fix" is just a workaround in our code. We do something like this...

```python

from cStringIO import StringIO

from django.core.files.storage import default_storage
from django.core.files.base import File

default_storage.save('filename.txt', File(StringIO('content'), name='filename.txt'))
```

pacahon:
@linuxlewis Thanks, that was my first thought. But `Storage` class accepts any file-object, so I'm looking for solution. Of course, my PR in its current state is just another workaround.

pacahon:
Should check if already was documented that `upload_fileobj` waits for `bytes` or `seekable file object` in binary mode (the same as old `put` method)

jleclanche:
PR needs rebasing but looks good to merge. @jschneier

jschneier:
Updated and will be applied in #368. Thanks for the report and patch. Apologies for the delay.

PR #196, opened 2016-08-31
S3Boto3Storage: Fix _get_or_create_bucket

linuxlewis:
The current implementation throws an `ImproperlyConfigured` error instead of creating the bucket. This occurs even when `AWS_AUTO_CREATE_BUCKET` has been set to true.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -347,6 +347,8 @@ class S3Boto3Storage(Storage):
                                                "region than we are connecting to. Set "
                                                "the region to connect to by setting "
                                                "AWS_S3_REGION_NAME to the correct region." % name)
+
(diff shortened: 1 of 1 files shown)


jschneier:
Hi and thanks for the PR! My one point of concern is that we are now calling `.head_bucket` unconditionally which would add more calls (that people pay for!). Do you have any thoughts on that?

linuxlewis:
@jschneier I updated the PR to address your concerns. Are head requests actually charged for? I couldn't find them on the pricing page: https://aws.amazon.com/s3/pricing/

jschneier:
I assume that it's contained with this:

`GET and all other Requests	$0.004 per 10,000 requests`

jschneier:
Thanks, merged

PR #197, opened 2016-09-02
fix for files with size more than buffer size

### LINKED ISSUE #160: Problem with S3BotoStorageFile and a file greater than 5Mb

vinayinvicible:
fixes #160

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -71,6 +71,8 @@ class S3BotoStorageFile(File):
         if buffer_size is not None:
             self.buffer_size = buffer_size
         self._write_counter = 0
+        # file position of the latest part file uploaded
+        self._last_part_pos = 0
 
     @property
     def size(self):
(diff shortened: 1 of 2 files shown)


vinayinvicible:
@jschneier can you take a look at this

vinayinvicible:
@jschneier

jschneier:
Thanks, will look at this today.

vinayinvicible:
@jschneier 
any updates on this.

jsh2134:
can this same change be ported to s3boto3.py ?

jschneier:
Consolidated and fixed in #504.

PR #200, opened 2016-09-12
Use a different instance of boto3 Session per thread.

jdufresne:
The default mod_wsgi configuration runs in 15 threads. Per the
documentation, boto3 is _not_ thread-safe by default. The library
recommends using a different instance of Session per thread. This avoids
race conditions in the library.

https://boto3.readthedocs.io/en/latest/guide/resources.html#multithreading

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 django-storages change log
 ==========================
 
+1.5.2 (XXXX-XX-XX)
+******************
+
+* Fix ``S3Boto3Storage`` to avoid race conditions in a multithreaded WSGI environment
+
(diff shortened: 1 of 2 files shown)


jdufresne:
Also filed the upstream PR https://github.com/boto/boto3/pull/806 however, not sure how that will be accepted.

leonsmith:
Sorry to comment on an old closed PR but this looks like it has not been resolved yet and we have bumped into this issue on a couple of projects.

@jdufresne do you think the best place to fix this upstream then?

jschneier:
Merged in the resurrected version of this @leonsmith.

PR #206, opened 2016-09-20
Add location to the create_bucket parameters

nobe4:
We're developing a EU project using this package, to interface nicely with AWS S3 storage. When trying to create a bucket automatically, this error is thrown:

```
S3ResponseError: S3ResponseError: 400 Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Error>
    <Code>IllegalLocationConstraintException</Code>
    <Message>The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.</Message>
    <RequestId>XXX</RequestId>
    <HostId>XXX</HostId>
</Error>
```

Turns out the `create_bucket` can take a `location` argument, but it's not in use currently.

From [the documentation](http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region): 

> If you use a region other than the US East (N. Virginia) endpoint to create a bucket, you must set the LocationConstraint bucket parameter to the same region.

The bucket location is already parsed from the django settings, inside the `location` property of the `S3BotoStorage` object. This change simply add the location argument when creating the bucket.

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -218,6 +218,7 @@ class S3BotoStorage(Storage):
     querystring_expire = setting('AWS_QUERYSTRING_EXPIRE', 3600)
     reduced_redundancy = setting('AWS_REDUCED_REDUNDANCY', False)
     location = setting('AWS_LOCATION', '')
+    origin = setting('AWS_ORIGIN', '')
     encryption = setting('AWS_S3_ENCRYPTION', False)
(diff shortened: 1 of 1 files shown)


jschneier:
@nobe4 this looks good. Can you update it to import `Location` from `boto.s3.connection` and set the default to be `Location.DEFAULT`? I see that that is actually the empty string but I don't think it's that clear.

jschneier:
Thanks!

PR #223, opened 2016-10-30
Use assertEqual in place of deprecated assertEquals

jdufresne:
Fixes warnings of the form "DeprecationWarning: Please use assertEqual instead." when running tests with warnings enabled.


jschneier:
thanks

PR #226, opened 2016-11-14
Fix double localization of mtime

### LINKED ISSUE #225: modified_time in s3boto3 raises ValueError: astimezone() cannot be applied to a naive datetime
In the [recent additions](https://github.com/jschneier/django-storages/commit/302aa743c621f7d801142ac54c825aabed8c1125) to the `s3boto3` TZ handling, running with `USE_TZ = False` under Django<1.10 (which calls `modified_time`), the following traceback occurs:

```
...
File "./django/contrib/staticfiles/management/commands/collectstatic.py", line 305, in copy_file
    if not self.delete_file(path, prefixed_path, source_storage):
  File "./django/contrib/staticfiles/management/commands/collectstatic.py", line 222, in delete_file
    self.storage.modified_time(prefixed_path)
  File "./storages/backends/s3boto3.py", line 537, in modified_time
    return localtime(self.get_modified_time(name)).replace(tzinfo=None)
  File "./site-packages/django/utils/timezone.py", line 309, in localtime
    value = value.astimezone(timezone)
ValueError: astimezone() cannot be applied to a naive datetime
```

This is a result of `get_modified_time` returning a naive datetime (as a result of `USE_TZ = False`), which `modified_time` then attempts to again convert to a localized time via `localtime`.

tomchuk:
Fixes #225

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -11,7 +11,7 @@ from django.utils.deconstruct import deconstructible
 from django.utils.encoding import force_text, smart_str, filepath_to_uri, force_bytes
 from django.utils.six.moves.urllib import parse as urlparse
 from django.utils.six import BytesIO
-from django.utils.timezone import localtime
+from django.utils.timezone import localtime, is_aware
 
 try:
     from boto3 import resource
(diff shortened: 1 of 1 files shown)


jschneier:
Fixed in #235.

PR #217, opened 2016-10-10
Apply smart_text on file names in python2.7 [s3boto3]

AGASS007:
Not a 100% sure it's the best approach, but there it is: my attempt to address #216.

I feel that `_encode_name` and `_decode_name` are starting to lose their naming/meaning values since they now end up doing almost the same thing.

Also, I didn't know what to precisely test. The only thing I know for sure is that the test I added fail without the switch from `smart_str` to `smart_text`.

PR #236, opened 2016-12-14
Add Google Cloud Storage backend using the gcloud-python library

scjody:
Based on https://github.com/jschneier/django-storages/pull/146

This uses the native `gcloud-python` library, which is recommended by Google for production setups and allows finer-grained authentication than current solutions (which only work when Google Cloud Storage is used in S3 compatibility mode).

* [x] Review code (see: https://github.com/jschneier/django-storages/pull/146#issuecomment-266873319)
* [x] Add tests
* [x] Add documentation

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -1,0 +1,190 @@
+Google Cloud Storage
+====================
+
+Usage
+*****
+
+This backend provides support for Google Cloud Storage using the
+library provided by Google.
+
+It's possible to access Google Cloud Storage in S3 compatibility mode
+using other libraries in django-storages, but this is the only library
+offering native support.
+
+By default this library will use the credentials associated with the
+current instance for authentication. To override this, see the
+settings below.
+
+
+Settings
+--------
+
+To use gcloud set::
+
+    DEFAULT_FILE_STORAGE = 'storages.backends.gcloud.GoogleCloudStorage'
+
+``GS_BUCKET_NAME``
+
+Your Google Storage bucket name, as a string.
+
+``GS_PROJECT_ID`` (optional)
+
+Your Google Cloud project ID. If unset, falls back to the default
+inferred from the environment.
+
+``GS_CREDENTIALS`` (optional)
+
+The OAuth 2 credentials to use for the connection. If unset, falls
+back to the default inferred from the environment.
+
+``GS_AUTO_CREATE_BUCKET`` (optional, default is ``False``)
+
+If True, attempt to create the bucket if it does not exist.
+
+``GS_DEFAULT_ACL`` (optional)
+
+If set to ``private`` changes uploaded file's Access Control List from the default permission ``public-read`` to give owner full control and remove read access from everyone else. 
+
+``GS_BUCKET_ACL`` (optional)
+
+ACL used when creating a new bucket; defaults to ``GS_DEFAULT_ACL``.
+
+``GS_FILE_CHARSET`` (optional)
+
+Allows overriding the character set used in filenames.
+
+``GS_FILE_OVERWRITE`` (optional: default is ``True``)
+
+By default files with the same name will overwrite each other. Set this to ``False`` to have extra characters appended.
+
+``GS_MAX_MEMORY_SIZE`` (optional)
+
+The maximum amount of memory a returned file can take up before being
+rolled over into a temporary file on disk. Default is 0: Do not roll over.
+
+Fields
+------
+
+Once you're done, default_storage will be Google Cloud Storage::
+
+    >>> from django.core.files.storage import default_storage
+    >>> print default_storage.__class__
+    <class 'storages.backends.gcloud.GoogleCloudStorage'>
+
+This way, if you define a new FileField, it will use the Google Cloud Storage::
+
+    >>> from django.db import models
+    >>> class Resume(models.Model):
+    ...     pdf = models.FileField(upload_to='pdfs')
+    ...     photos = models.ImageField(upload_to='photos')
+    ...
+    >>> resume = Resume()
+    >>> print resume.pdf.storage
+    <storages.backends.gcloud.GoogleCloudStorage object at ...>
+
+Storage
+-------
+
+Standard file access options are available, and work as expected::
+
+    >>> default_storage.exists('storage_test')
+    False
+    >>> file = default_storage.open('storage_test', 'w')
+    >>> file.write('storage contents')
+    >>> file.close()
+
+    >>> default_storage.exists('storage_test')
+    True
+    >>> file = default_storage.open('storage_test', 'r')
+    >>> file.read()
+    'storage contents'
+    >>> file.close()
+
+    >>> default_storage.delete('storage_test')
+    >>> default_storage.exists('storage_test')
+    False
+
+Model
+-----
+
+An object without a file has limited functionality::
+
+    >>> obj1 = MyStorage()
+    >>> obj1.normal
+    <FieldFile: None>
+    >>> obj1.normal.size
+    Traceback (most recent call last):
+    ...
+    ValueError: The 'normal' attribute has no file associated with it.
+
+Saving a file enables full functionality::
+
+    >>> obj1.normal.save('django_test.txt', ContentFile('content'))
+    >>> obj1.normal
+    <FieldFile: tests/django_test.txt>
+    >>> obj1.normal.size
+    7
+    >>> obj1.normal.read()
+    'content'
+
+Files can be read in a little at a time, if necessary::
+
+    >>> obj1.normal.open()
+    >>> obj1.normal.read(3)
+    'con'
+    >>> obj1.normal.read()
+    'tent'
+    >>> '-'.join(obj1.normal.chunks(chunk_size=2))
+    'co-nt-en-t'
+
+Save another file with the same name::
+
+    >>> obj2 = MyStorage()
+    >>> obj2.normal.save('django_test.txt', ContentFile('more content'))
+    >>> obj2.normal
+    <FieldFile: tests/django_test_.txt>
+    >>> obj2.normal.size
+    12
+
+Push the objects into the cache to make sure they pickle properly::
+
+    >>> cache.set('obj1', obj1)
+    >>> cache.set('obj2', obj2)
+    >>> cache.get('obj2').normal
+    <FieldFile: tests/django_test_.txt>
+
+Deleting an object deletes the file it uses, if there are no other objects still using that file::
+
+    >>> obj2.delete()
+    >>> obj2.normal.save('django_test.txt', ContentFile('more content'))
+    >>> obj2.normal
+    <FieldFile: tests/django_test_.txt>
+
+Default values allow an object to access a single file::
+
+    >>> obj3 = MyStorage.objects.create()
+    >>> obj3.default
+    <FieldFile: tests/default.txt>
+    >>> obj3.default.read()
+    'default content'
+
+But it shouldn't be deleted, even if there are no more objects using it::
+
+    >>> obj3.delete()
+    >>> obj3 = MyStorage()
+    >>> obj3.default.read()
+    'default content'
+
+Verify the fix for #5655, making sure the directory is only determined once::
+
+    >>> obj4 = MyStorage()
+    >>> obj4.random.save('random_file', ContentFile('random content'))
+    >>> obj4.random
+    <FieldFile: .../random_file>
+
+Clean up the temporary files::
+
+    >>> obj1.normal.delete()
+    >>> obj2.normal.delete()
+    >>> obj3.default.delete()
+    >>> obj4.random.delete()
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -41,13 +41,17 @@ back to the default inferred from the environment.
 
 If True, attempt to create the bucket if it does not exist.
 
-``GS_DEFAULT_ACL`` (optional)
+``GS_AUTO_CREATE_ACL`` (optional, default is ``projectPrivate``)
 
-If set to ``private`` changes uploaded file's Access Control List from the default permission ``public-read`` to give owner full control and remove read access from everyone else. 
+ACL used when creating a new bucket, from the
+`list of predefined ACLs <https://cloud.google.com/storage/docs/access-control/lists#predefined-acl>`_.
+(A "JSON API" ACL is preferred but an "XML API/gsutil" ACL will be
+translated.)
 
-``GS_BUCKET_ACL`` (optional)
-
-ACL used when creating a new bucket; defaults to ``GS_DEFAULT_ACL``.
+Note that the ACL you select must still give the service account
+running the gcloud backend to have OWNER permission on the bucket. If
+you're using the default service account, this means you're restricted
+to the ``projectPrivate`` ACL.
 
 ``GS_FILE_CHARSET`` (optional)
 
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -130,13 +130,6 @@ class GoogleCloudStorage(Storage):
                                        "setting GS_AUTO_CREATE_BUCKET to "
                                        "``True``." % name)
 
-    def _normalize_name(self, name):
-        """
-        No normalizing supported. This can be implemented later.
-        TODO: Implement normalizing, like the s3boto backend.
-        """
-        return name
-
     def _clean_name(self, name):
         """
         Cleans the name so that Windows style paths work
@@ -150,25 +143,24 @@ class GoogleCloudStorage(Storage):
         return force_text(name, encoding=self.file_name_charset)
 
     def _open(self, name, mode='rb'):
-        name = self._normalize_name(self._clean_name(name))
+        name = self._clean_name(name)
         file_object = self.file_class(name, mode, self)
         if not file_object.blob:
             raise IOError('File does not exist: %s' % name)
         return file_object
 
     def _save(self, name, content):
-        cleaned_name = self._clean_name(name)
-        name = self._normalize_name(cleaned_name)
+        name = self._clean_name(name)
         size = getattr(content, 'size')
 
-        content.name = cleaned_name
+        content.name = name
         encoded_name = self._encode_name(name)
         file = self.file_class(encoded_name, 'rw', self)
         file.blob.upload_from_file(content, size=size)
-        return cleaned_name
+        return name
 
     def delete(self, name):
-        name = self._normalize_name(self._clean_name(name))
+        name = self._clean_name(name)
         self.bucket.delete_blob(self._encode_name(name))
 
     def exists(self, name):
@@ -179,11 +171,11 @@ class GoogleCloudStorage(Storage):
             except ImproperlyConfigured:
                 return False
 
-        name = self._normalize_name(self._clean_name(name))
+        name = self._clean_name(name)
         return bool(self.bucket.get_blob(self._encode_name(name)))
 
     def listdir(self, name):
-        name = self._normalize_name(self._clean_name(name))
+        name = self._clean_name(name)
         # for the bucket.list and logic below name needs to end in /
         # But for the root path "" we leave it as an empty string
         if name and not name.endswith('/'):
@@ -206,18 +198,18 @@ class GoogleCloudStorage(Storage):
         return list(dirs), files
 
     def size(self, name):
-        name = self._encode_name(self._normalize_name(self._clean_name(name)))
+        name = self._encode_name(self._clean_name(name))
         blob = self.bucket.get_blob(self._encode_name(name))
         return blob.size if blob else 0
 
     def modified_time(self, name):
-        name = self._normalize_name(self._clean_name(name))
+        name = self._clean_name(name)
         blob = self.bucket.get_blob(self._encode_name(name))
         return blob.updated if blob else None
 
     def url(self, name):
         # Preserve the trailing slash after normalizing the path.
-        name = self._normalize_name(self._clean_name(name))
+        name = self._clean_name(name)
         blob = self.bucket.get_blob(self._encode_name(name))
         return blob.public_url if blob else None
 
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -8,10 +8,10 @@ from storages.compat import Storage
 from storages.utils import setting
 
 try:
-    from gcloud.storage.client import Client
-    from gcloud.storage.bucket import Bucket
-    from gcloud.storage.blob import Blob
-    from gcloud.exceptions import NotFound
+    from google.cloud.storage.client import Client
+    from google.cloud.storage.bucket import Bucket
+    from google.cloud.storage.blob import Blob
+    from google.cloud.exceptions import NotFound
 except ImportError:
     raise ImproperlyConfigured("Could not load Google Storage bindings.\n"
                                "See https://github.com/GoogleCloudPlatform/gcloud-python")
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -53,10 +53,11 @@ class GoogleCloudFile(File):
     def read(self, num_bytes=None):
         if 'r' not in self._mode:
             raise AttributeError("File was not opened in read mode.")
-        if num_bytes is not None:
-            return super(GoogleCloudFile, self).read(num_bytes)
-        else:
-            return super(GoogleCloudFile, self).read()
+
+        if num_bytes is None:
+            num_bytes = -1
+
+        return super(GoogleCloudFile, self).read(num_bytes)
 
     def write(self, content):
         if 'w' not in self._mode:
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -5,7 +5,7 @@ from django.core.files.base import File
 from django.utils.deconstruct import deconstructible
 from django.utils.encoding import force_bytes, force_text, smart_str
 from storages.compat import Storage
-from storages.utils import clean_name, setting
+from storages.utils import clean_name, safe_join, setting
 
 try:
     from google.cloud.storage.client import Client
@@ -134,6 +134,14 @@ class GoogleCloudStorage(Storage):
         """
         return clean_name(name)
 
+    def _normalize_name(self, name):
+        """
+        Normalizes the name so that paths like /path/to/ignored/../something.txt
+        and ./file.txt work.  Note that clean_name adds ./ to some paths so
+        they need to be fixed here.
+        """
+        return safe_join('', name)
+
     def _encode_name(self, name):
         return smart_str(name, encoding=self.file_name_charset)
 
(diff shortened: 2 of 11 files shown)


storages/backends/gcloud.py
```diff
@@ -0,0 +1,241 @@
...
+                self.file.seek(0)
+                content_type, _ = mimetypes.guess_type(self.name)
+                content_type = getattr(self.file, 'content_type', content_type)
+                size = getattr(self.file, 'size')
+                self.blob.upload_from_file(self.file, content_type=content_type, size=size)
+            self._file.close()
+            self._file = None
+
+
+class GoogleCloudStorage(Storage):
```
scjody: @eirsyl The existing backends all use the `@deconstructible` decorator here but I'm not sure why. Commit comments suggest it has something to do with migrations but I'm able to migrate to a `FileField` using this backend for storage with this code as is. Therefore I'm inclined to leave the decorator out and if someone needs it, they can add it. What do you think?

scjody: Never mind, I figured out why the decorator is needed and added it.

scjody:
I reviewed the code and made some minor changes. It looks good - very close to the S3 implementation, but S3 and Google Cloud Storage are very close to begin with.

Onto tests!

scjody:
@eirsyl @jschneier Does this look reasonable?

Next steps on our end are that we'll be using this in our staging environment within the next 2 weeks.

eirsyl:
Great work! Looks good to me 👍

scjody:
@jschneier We've been using this successfully in our staging environment for several days. Would you be able to merge it soon please, or let me know what's still missing?

scjody:
@jschneier Can you please take a look at this? The README says to bug you until it's merged, so that's what I'm doing 😸.   It also said to add myself to AUTHORS so I added myself and @eirsyl. @eirsyl please let me know if you want a different name or don't want to be credited.

scjody:
@jschneier Will you have time to look at this sometime soon please?

@ZuluPro Is this something you could look at instead? (I can't tell if you have write access but it looks like you might.)

jschneier:
Yes, currently on vacation but have set aside Monday afternoon to do maintenance on this library, thanks for your patience.

sherbondy:
I think there might be a renewed interest in getting this PR merged given the events that unfolded today :/

scjody:
@jschneier Just a reminder: can you please review this soon? I've updated AUTHORS again to fix the merge conflict.

p0r:
This is great development, and I hope it gets merged soon. I'd like to present one wish though - please take #281 into account and make use of tempfile.SpooledTemporaryFile optional / configurable.

stinky2nine:
@jschneier Friendly ping.

mhindery:
Would also love this to be merged, as the current solution for GCS is far from ideal ...

jschneier:
Hi.

Can someone articulate what the issue is with the current `gs.py` binding? I'm happy to merge something like this since it seems that there is demand for it. I also greatly appreciate that the code is similar to the other backends that exist although I can't help but feel we can probably find a better abstraction since there is quite a bit shared (and that the code quality of `S3BotoStorage` is showing its age a bit).

When we do merge this I'd like to deprecate `gs.py`.

stinky2nine:
@jschneier `gs.py` is using Google Cloud Storage's [XML Interoperable API](https://cloud.google.com/storage/docs/interoperability) which uses keyed-hash message authentication code (a.k.a. developer keys) that is linked to your Google account. Interoperable API is really meant for migration to Google Cloud Storage. The biggest problem with the developer keys is security and privacy. Developer keys should not be shared with anyone as they can be used to gain access to other Google Cloud Storage buckets linked to your Google account.

I believe that this pull request depends on [`gcloud.storage.client.Client`](https://googlecloudplatform.github.io/google-cloud-python/0.18.1/storage-client.html) which would require using [OAuth 2.0](https://cloud.google.com/storage/docs/authentication) for Google Cloud Storage API authN and authZ.

mhindery:
As extra note on the explanation above:

- If you work with multiple developers on a project, they each have to enable the interoperability mode and create keys which then have to be picked up by you project config/settings. This means quite some unnecessary work and config. Alternatively you would have to distribute the credentials for 1 account to everybody. This is not wanted.
- When deploying on Google Cloud instances or (kubernetes) clusters, there are by default application credentials present for authentication against GCS (service accounts). It is these services accounts that are used to manage roles and authentication on the Google Cloud platform everywhere. On developer machines, they are also present system-wide for all applications to pick up by default. Each program thus runs under the dev account when running locally, and uses the instance/cluster's account when deployed somewhere, and this all happens automatically. They are the recommended way of authenticating and are used throughout the platform (https://developers.google.com/identity/protocols/application-default-credentials), available for multiple languages.

scjody:
@jschneier It's mostly about authentication as others have articulated above. (I've personally tested the new functionality using an instance's service account and with a developer's user account, but I expect any authentication type supported by the `google-cloud-python` library will also work.)

I also believe that since the new backend uses the official Google library directly and does not require a compatibility mode intended only for short term data migrations, it is more likely to be reliable in the long term.

scjody:
@mhindery Are you sure you're in the same virtualenv in both cases? Can you try debugging the `try` part of the code above using `ipdb`? Or removing the `try`/`except` to see what the exact error is?

scjody:
@jschneier I just rebased this to deal with your recent merges. This is ready to go again; please let me know if you have any concerns or want any changes.

There are some strange things in the existing backends, especially s3boto, but addressing that is beyond the scope of this PR (and in many cases would result in backwards-incompatible changes). I did move some code from `s3boto` to `utils` to avoid duplication between the backends.

storages/backends/gcloud.py
```diff
@@ -0,0 +1,226 @@
...
+        name = self._normalize_name(self._clean_name(name))
+        file_object = self.file_class(name, mode, self)
+        if not file_object.blob:
+            raise IOError('File does not exist: %s' % name)
+        return file_object
+
+    def _save(self, name, content):
+        cleaned_name = self._clean_name(name)
+        name = self._normalize_name(cleaned_name)
+        size = getattr(content, 'size')
```
jschneier: Why do this with `getattr`? If it doesn't exist this is going to fail with `AttributeError` anyway so it can just be inlined right i.e. `size=content.size`. Does the `.upload_from_file` not automatically infer size as well?

storages/backends/gcloud.py
```diff
@@ -0,0 +1,226 @@
...
+            self._file.close()
+            self._file = None
+
+
+@deconstructible
+class GoogleCloudStorage(Storage):
+    client_class = Client
+    file_class = GoogleCloudFile
+
+    not_found_exception = NotFound
```
jschneier: I think we can remove this and probably some of the other class based properties. In this case I think it is there to make it easy to create the `gs.py` file unless you can think of a solid reason to keep it.

storages/backends/gcloud.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def size(self, name):
+        name = self._normalize_name(self._clean_name(name))
+        blob = self.bucket.get_blob(self._encode_name(name))
+        return blob.size if blob else 0
+
+    def modified_time(self, name):
+        name = self._normalize_name(self._clean_name(name))
+        blob = self.bucket.get_blob(self._encode_name(name))
+        return blob.updated if blob else None
```
jschneier: Need to add `get_modified_time` as well. Is this a naive time? This has to be naive while `get_modified_time` depends on the setting `USE_TZ`. Looking at `S3BotoStorage` might be useful.

storages/backends/gcloud.py
```diff
@@ -0,0 +1,226 @@
...
+                bucket = self.client.create_bucket(name)
+                bucket.acl.all().grant(self.bucket_acl)
+                bucket.acl.save()
+                return bucket
+            raise ImproperlyConfigured("Bucket %s does not exist. Buckets "
+                                       "can be automatically created by "
+                                       "setting GS_AUTO_CREATE_BUCKET to "
+                                       "``True``." % name)
+
+    def _clean_name(self, name):
```
jschneier: Should we still have this method?

tests/test_gcloud.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def test_get_available_name(self):
+        self.storage.file_overwrite = True
+        self.assertEqual(self.storage.get_available_name(self.filename), self.filename)
+
+        self.storage._bucket = mock.MagicMock()
+        self.storage._bucket.get_blob.return_value = None
+        self.storage.file_overwrite = False
+        self.assertEqual(self.storage.get_available_name(self.filename), self.filename)
+        self.storage._bucket.get_blob.assert_called_with(self.filename)
```
jschneier: Let's add a unicode test name.  Would like to not have Python 2.7 issues aka [this bug](https://github.com/jschneier/django-storages/pull/217).

storages/backends/gcloud.py
```diff
@@ -0,0 +1,226 @@
...
+        return safe_join('', name)
+
+    def _encode_name(self, name):
+        return smart_str(name, encoding=self.file_name_charset)
+
+    def _open(self, name, mode='rb'):
+        name = self._normalize_name(self._clean_name(name))
+        file_object = self.file_class(name, mode, self)
+        if not file_object.blob:
+            raise IOError('File does not exist: %s' % name)
```
jschneier: How about we stick with `NotFound` throughout.

storages/backends/gcloud.py
```diff
@@ -0,0 +1,226 @@
...
+                # File
+                files.append(parts[0])
+            elif len(parts) > 1 and parts[0]:
+                # Directory
+                dirs.add(parts[0])
+        return list(dirs), files
+
+    def size(self, name):
+        name = self._normalize_name(self._clean_name(name))
+        blob = self.bucket.get_blob(self._encode_name(name))
```
jschneier: Ugh really do not like that `get_blob` suppresses errors, it should not do that. I'd prefer to fail loudly when we do an action on something that doesn't exist. Size of 0 doesn't make too much sense to me. Maybe we wrap it up in our own `_get_blob` and just `raise NotFound` when it's...not found? Thoughts?

jschneier:
@scjody okay, left a few thoughts, should have done it as a review, sorry about that. going to let it stew for a bit and then maybe add some other thoughts. let me know what you think about them.

also, when we take this in you are okay if i give you access and you can be a maintainer of this particular piece of code?

jschneier:
Ah, can you also add a deprecation warning for `gs.py`! Forgot about that. And include the authentication/security reasons from above in this thread.

dummerbd:
It looks like @scjody already has a good handle on this, but if you need any help I can lend a hand. I just spent most of today writing up my own backend using the `google.cloud` library before I came across this PR, so I might be able to help out.

mhindery:
@scjody I have looked into the import error. I am certainly in the right environment (see attached traceback; print the installed packages, check my gcloud credentials, print django storage class -> error)
[traceback.txt](https://github.com/jschneier/django-storages/files/913318/traceback.txt)

I then tried the import with `ipdb` which gives 
```python
In [1]: from django.core.files.storage import default_storage

In [2]: print default_storage.__class__
> /Users/mhindery/repositories/otainsight/parity/storages/backends/gcloud.py(13)<module>()
     11     import ipdb
     12     ipdb.set_trace(context=5)
---> 13     from google.cloud.storage.client import Client
     14     from google.cloud.storage.blob import Blob
     15     from google.cloud.exceptions import NotFound

ipdb> n
ImportError: 'No module named compat'
> /Users/mhindery/repositories/otainsight/parity/storages/backends/gcloud.py(13)<module>()
     11     import ipdb
```

Some googling did not lead me to a solution on this, neither did I find something on it in the issues of the google-cloud-python repo.

storages/backends/gcloud.py
scjody: I'm not sure why it was done this way anymore, but your suggestion makes sense so I'll implement it.

`upload_from_file()` can infer the size if it's an actual file, but not if it's a `StringIO` or similar.

storages/backends/gcloud.py
scjody: OK, I'll remove this and anything like it.

jschneier:
@mhindery there used to be a `compat.py` file in this project but it has since been removed. You might need a rebased version of this branch.

storages/backends/gcloud.py
scjody: This is currently aware, so I'll make it naive and add the other method.

storages/backends/gcloud.py
scjody: It doesn't need to be a method, no. I've changed it to just call `clean_name` directly.

tests/test_gcloud.py
scjody: Good idea, will do.

storages/backends/gcloud.py
scjody: s3boto raises `IOError` in this case, so I want to stick with that to avoid breaking things for someone who switches from s3boto to gcloud.

storages/backends/gcloud.py
scjody: Yes, that makes sense. I made that change here and in `modified_time` and `url`.

storages/backends/gcloud.py
jschneier: Sounds good :+1:

storages/backends/gcloud.py
```diff
@@ -203,24 +191,33 @@ def listdir(self, name):
                 dirs.add(parts[0])
         return list(dirs), files
 
+    def _get_blob(self, name):
+        # Wrap google.cloud.storage's blob to raise if the file doesn't exist
+        blob = self.bucket.get_blob(name)
+
+        if blob is None:
+            raise NotFound('File does not exist')
```
jschneier: Can you include the `name` in the error message.

storages/backends/gcloud.py
```diff
@@ -0,0 +1,223 @@
...
+    from google.cloud.storage.client import Client
+    from google.cloud.storage.blob import Blob
+    from google.cloud.exceptions import NotFound
+except ImportError:
+    raise ImproperlyConfigured("Could not load Google Cloud Storage bindings.\n"
+                               "See https://github.com/GoogleCloudPlatform/gcloud-python")
+
+
+class GoogleCloudFile(File):
+    def __init__(self, name, mode, storage, buffer_size=None):
```
jschneier: `buffer_size` is unused?

storages/backends/gcloud.py
```diff
@@ -0,0 +1,223 @@
...
+                self.blob.download_to_file(self._file)
+                self._file.seek(0)
+        return self._file
+
+    def _set_file(self, value):
+        self._file = value
+
+    file = property(_get_file, _set_file)
+
+    def read(self, *args, **kwargs):
```
jschneier: `read` should not take in arbitrary `args` and `kwargs`

storages/backends/gcloud.py
```diff
@@ -0,0 +1,223 @@
...
+        if 'r' not in self._mode:
+            raise AttributeError("File was not opened in read mode.")
+        return super(GoogleCloudFile, self).read(*args, **kwargs)
+
+    def write(self, content, *args, **kwargs):
```
jschneier: nor should `write`

storages/backends/gcloud.py
scjody: 👍

storages/backends/gcloud.py
scjody: Yes, well spotted!

storages/backends/gcloud.py
scjody: Fixed.

storages/backends/gcloud.py
scjody: Fixed.

scjody:
@jschneier I addressed all your comments and added the deprecation warning. I think this is ready to go now.

> also, when we take this in you are okay if i give you access and you can be a maintainer of this particular piece of code?

Yes, no problem at all.

mhindery:
FWIW I have got it working on our staging environment (for some reason there was a storages/backends/google.py file leftover in my fork I used to try this out).

I am also wondering if it is possible to not set the ACL when uploading files / creating buckets. Instead, just user/inherit the default permissions we have configured in the web UI. It seems now it has to set it to something, which is by default `public-read`. We would like to configure this in the Gcloud UI, not in the django settings.

scjody:
> I am also wondering if it is possible to not set the ACL when uploading files / creating buckets. Instead, just user/inherit the default permissions we have configured in the web UI.

@mhindery That should work with the PR as is. I've been integration testing the PR with our software on a GCE server that authenticates to GCS using instance credentials and it works. Can you provide more detail on why you need to configure the ACL in the Django settings?

storages/backends/gs.py
```diff
@@ -1,3 +1,13 @@
+# DEPRECATION NOTICE: This backend is deprecated in favour of the
+# "gcloud" backend.  This backend uses Google Cloud Storage's XML
+# Interoperable API which uses keyed-hash message authentication code
+# (a.k.a. developer keys) that are linked to your Google account.  The
+# interoperable API is really meant for migration to Google Cloud
+# Storage. The biggest problem with the developer keys is security and
+# privacy. Developer keys should not be shared with anyone as they can
+# be used to gain access to other Google Cloud Storage buckets linked
+# to your Google account.
```
jschneier: The text is good but the user won't know. Usually I handle this the same as Django:

```python
import warnings
warnings.warn(
       "...text here", PendingDeprecationWarning)
```

Side question: I think that the xml compatibility layer will live on in the libcloud backend, does that seem correct?

storages/backends/gcloud.py
```diff
@@ -0,0 +1,233 @@
...
+    def get_modified_time(self, name):
+        name = self._normalize_name(clean_name(name))
+        blob = self._get_blob(self._encode_name(name))
+        updated = blob.updated
+        return updated if setting('USE_TZ') else timezone.make_naive(updated)
+
+    def url(self, name):
+        # Preserve the trailing slash after normalizing the path.
+        name = self._normalize_name(clean_name(name))
+        blob = self._get_blob(self._encode_name(name))
```
jschneier: Let's push the `_encode_name(_normalize_name(clean_name(name)))` into `_get_blob`. It's always the same.

storages/backends/gcloud.py
```diff
@@ -0,0 +1,233 @@
...
+        file_object = GoogleCloudFile(name, mode, self)
+        if not file_object.blob:
+            raise IOError(u'File does not exist: %s' % name)
+        return file_object
+
+    def _save(self, name, content):
+        cleaned_name = clean_name(name)
+        name = self._normalize_name(cleaned_name)
+
+        content.name = cleaned_name
```
jschneier: Curious as to why `content.name` is not wrapped in `_encode_name`?

storages/backends/gcloud.py
jschneier: Sort of wondering if we can just do something like `_get_name(name)` and have that method alone do all 3 things, they basically always come together except in a couple places.

storages/backends/gcloud.py
```diff
@@ -0,0 +1,233 @@
...
+            self._file = None
+
+
+@deconstructible
+class GoogleCloudStorage(Storage):
+    project_id = setting('GS_PROJECT_ID', None)
+    credentials = setting('GS_CREDENTIALS', None)
+    bucket_name = setting('GS_BUCKET_NAME', None)
+    auto_create_bucket = setting('GS_AUTO_CREATE_BUCKET', False)
+    default_acl = setting('GS_DEFAULT_ACL', 'public-read')
```
stinky2nine: @scjody I am wondering if the default here should be set to `None` and if `None`, ensure to use the default [object ACLs](https://cloud.google.com/storage/docs/access-control/lists#defaultobjects) set on the bucket.

docs/backends/gcloud.rst
```diff
@@ -0,0 +1,190 @@
...
+The OAuth 2 credentials to use for the connection. If unset, falls
+back to the default inferred from the environment.
+
+``GS_AUTO_CREATE_BUCKET`` (optional, default is ``False``)
+
+If True, attempt to create the bucket if it does not exist.
+
+``GS_DEFAULT_ACL`` (optional)
+
+If set to ``private`` changes uploaded file's Access Control List from the default permission ``public-read`` to give owner full control and remove read access from everyone else.
```
stinky2nine: Perhaps list the list of the options here or reference to https://cloud.google.com/storage/docs/access-control/lists#predefined-acl.

In addition, the predefined ACL aliases for JSON API are different than XML API.

JSON API | XML API
--- | ---
private | private
bucketOwnerRead | bucket-owner-read
bucketOwnerFullControl | bucket-owner-full-control
projectPrivate | project-private
authenticatedRead | authenticated-read
publicRead | public-read
publicReadWrite | public-read-write

storages/backends/gcloud.py
```diff
@@ -0,0 +1,233 @@
...
+
+    file = property(_get_file, _set_file)
+
+    def read(self, num_bytes=None):
+        if 'r' not in self._mode:
+            raise AttributeError("File was not opened in read mode.")
+        if num_bytes is not None:
+            return super(GoogleCloudFile, self).read(num_bytes)
+        else:
+            return super(GoogleCloudFile, self).read()
```
scjody: Unfortunately we do. [Django defines the default argument to be `None`](https://docs.djangoproject.com/en/dev/ref/files/file/#django.core.files.File.read) but `SpooledTemporaryFile` (and Python `file` objects in general) can't handle the case where the first argument (`size`) is `None`:

```
  File "/home/ubuntu/gcs_integration/westore/tests.py", line 31, in test_write_to_storage
    self.assertEqual(f.read(), 'storage contents')
  File "/home/ubuntu/gcs_integration/storages/backends/gcloud.py", line 56, in read
    return super(GoogleCloudFile, self).read(num_bytes)
  File "/usr/lib/python2.7/tempfile.py", line 585, in read
    return self._file.read(*args)
TypeError: an integer is required
```

Instead the argument has to be omitted or set to a negative number.

The reason you haven't run into this in the `s3boto` backend is because [it uses `*args and **kwargs`](https://github.com/jschneier/django-storages/blob/e52a127523fdd5be50bb670ccad566c5d527f3d1/storages/backends/s3boto.py#L134-L137), which means that if the user omits the argument there's no `None` to pass through to the underlying `read()`. I believe if the user _does_ specify `num_bytes=None` s3boto will break, but I haven't tested this.

I agree that `num_bytes` is an unfortunate name 😿.

storages/backends/gcloud.py
scjody: This is for compatibility with `s3boto`. My basic approach is that someone should be able to take code that works with `s3boto`, drop in `gcloud`, and have it work (including quirks like this).

We do run `_encode_name` before creating the remote object, but the unencoded name is passed to `upload_from_file` (in `content`), similarly to how `s3boto` passes the unencoded name to `set_contents_from_file`.

storages/backends/gcloud.py
scjody: Yes we could, but there are strange cases like the one you highlighted above. The way things are written now, it's easy to verify that `gcloud` works in the same way as `s3boto`.

I can make that change in this PR if you really want me to, but I think the right way to do it would be to merge this PR first then clean up this backend, `s3boto`, and any other backend that does this sort of thing in a separate PR.

storages/backends/gs.py
scjody: OK, I'll make it a warning.

I believe you're correct about libcloud, but I haven't tested it myself.

storages/backends/gcloud.py
scjody: This is only used in the case where we create the bucket (`auto_create_bucket = True`). In that case we need a default because there are no default object ACLs we can use.

scjody: I'll simplify the code and docs though. There's no need for two settings here.

docs/backends/gcloud.rst
scjody: From my understanding of the docs, you should be able to use either.

Unfortunately while testing this I ran into what's either a bug in the `google-cloud` python library or a huge misunderstanding on my part. I'm going to submit a support request to Google and see where that goes, and I'll update the PR when I have more information.

storages/backends/gs.py
```diff
@@ -1,3 +1,14 @@
+import warnings
+warnings.warn("DEPRECATION NOTICE: This backend is deprecated in favour of the "
```
scjody: @jschneier Like this, or would you like me to line wrap the warning output?

scjody:
@jschneier I made the warning change you requested and commented on your other requests. Please don't merge this until I deal with https://github.com/jschneier/django-storages/pull/236#discussion_r111692934 though.

docs/backends/gcloud.rst
stinky2nine: It seems like `google-cloud` is using JSON API and does **transform** predefined ACL aliases from XML to JSON format [here](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/storage/google/cloud/storage/acl.py#L493). I would suggest only reference to JSON format here, e.g. `s/public-read/publicRead/g'.

scjody:
@stinky2nine I addressed your comments by:
* Simplifying the ACL options to one (`GS_AUTO_CREATE_ACL`).
* Improving the documentation, including a link to the list on Google's site and a note on what I learned from my support request.

Thanks for the input!

@jschneier This is ready to go again. Please let me know if you want any more changes. I commented above re: the `*_name` methods, but I'm happy to make the change you suggested if you think that's the best way forward.

storages/backends/gcloud.py
jschneier: Interesting, that appears to be fixed on Python 3.5 at least.

jschneier: I guess the only way to do it would be something like:
```python
if num_bytes is None:
  num_bytes = -1
...
```
Also I think we can just change the name from `num_bytes` to `size` and the default to `-1`, this is a new backend and I'll note it in the changelog if it is surprising.

storages/backends/gcloud.py
jschneier: Yeah, let's not touch it. We have all these backends I'm going to fix all this in all backends with common code in one go later.

storages/backends/gs.py
jschneier: The syntax is fine. `warn` takes a second argument of the type to issue which in this case should be `DeprecationWarning` or `PendingDeprecationWarning`.

storages/backends/gcloud.py
jschneier: To be clear I'm not usually very cavalier about breaking backwards compat but I think we'll have to go and do it once in a biggish way to wrestle all of these things to decent shape. We can leave this one here if you want for s3boto interaction for now until we hit that point but I'd be pretty surprised if anyone is kwarg'ing the arg to `read`.

jschneier:
@scjody I just gave you collaboration access, fix up my last couple of comments at your digression and then go ahead and merge.

stinky2nine:
@scjody Looks good to me. I work at Google so feel free to ask any questions and thanks for putting together this pull request.

storages/backends/gcloud.py
scjody: I'd be surprised too but [Django specifies `num_bytes` as the kwarg](https://docs.djangoproject.com/en/dev/ref/files/file/#django.core.files.File.read) so it would be reasonable for someone to do that, and really confusing to debug if they did.

Reading the Django docs a bit more closely I noticed that the text gives the name of the argument as `size`, so there's an inconsistency there. I think the best way forward is to submit a PR to fix the Django docs, wait for a version with the fix to be released, _then_ change this argument (here and in other backends).

In any case I've made the `if` change you suggested because your suggestion makes it a bit clearer.

storages/backends/gs.py
scjody: I intended to do that but forgot. Thanks for noticing it!

storages/backends/gcloud.py
jschneier: Good call: https://github.com/django/django/pull/8379

dummerbd:
Anyone got an ETA on when this version will be released?

Sgiath:
@scjody This is great I am able to use it properly but I have one problem: how I can make all objects uploaded by Django to the Google Storage publicly visible? Are there some settings for it? Maybe I can set it up on Google Cloud Storage itself? @stinky2nine 
Thanks

PR #238, opened 2016-12-18
Fix S3Boto3Storage to avoid race conditions in a multi-threaded WSGI environment

jdufresne:
Use a different boto3 session for each instance of the S3 connection.

---

Resurrecting from PR #200 as upstream does not appear to be moving on a change. This fixes the race as far as django-storages is concerned.


jschneier:
Unfortunate about upstream. Thanks for the fix.

Issue #28, opened 2015-02-15
AWS S3 Frankfurt region not working


FPurchess:
**COPY OF BITBUCKET ISSUE #214 - https://bitbucket.org/david/django-storages/issue/214/aws-s3-frankfurt-region-not-working**

"Andreas Schilling created an issue 2015-01-04

Using Frankfurt region (Germany) with django-storages produces HTTP 400 error. S3 in the the new region supports only Signature Version 4. In all other regions, Amazon S3 supports both Signature Version 4 and Signature Version 2.

I assume django-storages only supports Signature Version 2. Is there any chance to support Version 4?"

Thanks @jschneier for the fork! Is there a chance for django-storages-redux to support the eu-central-1 region?

FPurchess:
s3 boto user can find a temporary workaround here: https://github.com/boto/boto/issues/2741

frisellcpl:
I can confirm that the above workaround works for Frankfurt.

Big thanks @FPurchess !

rotherfuchs:
I can confirm this, too! Thank you!!

BodhiSukha:
You can just add this to the settings:

```
os.environ['S3_USE_SIGV4'] = 'True'
AWS_S3_HOST = 's3.eu-central-1.amazonaws.com'
```

(Note: I'm using the bitbucket repo)

jschneier:
@BodhiSukha thanks for that, going to close now. If you notice any regression in moving from the BitBucket repo to this one please open an issue.

satyrius:
@BodhiSukha `AWS_S3_HOST = 's3.eu-central-1.amazonaws.com'` does not work for me. Any idea why? Hacking class as @FPurchess works fine.
```
boto==2.36.0
django-storages==1.1.8
```

BodhiSukha:
@satyrius sorry just saw you comment. 

I'm using it in 2 different setups  
```
boto==2.34.0 
django-storages==1.1.8
```
and 
```
boto==2.36.0 
django-storages==1.1.8
```
works fine on both of them. Unfortunately I'm not sure about your problem.  Only other thing different in my settings is the calling format. In python 2.7.9 environment I use this extra setting
```
AWS_S3_CALLING_FORMAT = boto.s3.connection.OrdinaryCallingFormat()
```

miracle2k:
When I try using this workaround, I get ``HostRequiredError: BotoClientError: When using SigV4, you must specify a 'host' parameter.``.

[...]

PR #241, opened 2016-12-30
Add 'AWS4-HMAC-SHA256 signatures' section for Amazon S3

ad-m:
No extra comments required. 

I am not native speak, so please check English especially.

Thank @sdeleeuw for works in #28 and @mbarrien for prepare alternative backend.


jschneier:
Thanks for this contribution. I'll need to improve the English a bit but it's very valuable information.

ad-m:
Do you need any additional information?

jleclanche:
Why not set default AWS_SIGNATURE_VERSION to s3v4? there is no harm doing so.

ad-m:
@jleclanche , not all S3  non-Amazon servers such as Ceph Object Gateway support s3v4.

jleclanche:
Okay sure, that's why we have the setting -- but the default should imho still look at aws s3 as the first class citizen.

The default behaviour being broken in some AWS regions is worse than the default behaviour being broken for some non-aws servers.

jleclanche:
On top of this, looking at the [changes in v4](https://docs.aws.amazon.com/general/latest/gr/sigv4_changes.html), it sounds a lot more secure than v2 ... not including the key in the request is a big deal. Much saner default.

jschneier:
We now have some language and I have opened #335 to default this.

PR #242, opened 2016-12-30
Improve syntax of docs

ad-m:
Hello,

I made small improvements syntax of docs eg.t:
- use definition lists to settings list - it looks better,
- use ``.. note:: `` for notes,
- fix case of word "optional" in whole docs,

Using definitions lists require some changes in text or reorder paragraph, so I tried unify them.


jschneier:
Awesome, thanks.

PR #251, opened 2017-01-22
add apache-libcloud installation note

oonid:
I try to configure django-storages to use google cloud storage refer to this file,
but I think it would be better to add apache-libcloud installation on this guideline,
so noob people like me won't need to have the exception first because know that must install apache-libcloud first.


ZuluPro:
Thanks for this contribution!

PR #252, opened 2017-01-23
Fix broken pypi version badge

nikolas:


jschneier:
Amusing how many of these badge services exist, someone got there already, sorry.

PR #260, opened 2017-01-29
Fix libcloud

fuwaneko:


ZuluPro:
@fuwaneko
In fact, this second argument is the `chunk_size`.
I think a better improvement would be to define this size in `__init__`/configuration,
I suggest a size of 8192 by default.

fuwaneko:
@ZuluPro I concur with setting it to fixed configuration value. Should I update this PR?

Issue #263, opened 2017-02-02
 Adds support for already gzipped content to S3 storage to the boto3 backend


ldng:
The PR #122 fixes an encoding bug in the boto backend but the fix has not been reported to the boto3 backend. Is a PR needed or is the fix sufficiently obvious ?

jschneier:
Merged in the corresponding PR, thanks!

PR #264, opened 2017-02-02
Adds support for already gzipped content to S3 storage to the boto3 backend #263

ldng:
…ackend #263


jschneier:
Hello, the code changes looks okay but the test fails. The way boto3 adds the data is not the same as the way boto does, fix that and then I can merge.

ldng:
Hi,

sorry took me so long but I got busy and forgot about the PR.
I fixed the test but I'm not familiar with codecov and do not know how to handle codecov/patch failure.

jschneier:
Don't worry about that part. Test coverage is something that I need to work on in general. Looks good, thanks!

PR #266, opened 2017-02-06
Added support for boto3 Cache-Control and Expires headers

TaiSHiNet:
AWS_HEADERS was completely ignored in boto3 backend. Cache-Control and Expires are kwargs for put_object (and derivates). Other headers should go into Metadata dict (not covered by this PR)


TaiSHiNet:
You're right, it wasn't documented and I missed it when looking at the code. 
Either way, AWS_HEADERS is ignored/misleading with boto3 backend

Bests

mbarrien:
When I wrote that, it was intentional; AWS_HEADERS is very Boto2 specific, and keeping a mapping between Boto2 and Boto3 fields is impossible to maintain long term (since a header may not map to a single field, or may map to something that isn't a parameter). It's especially not worth it when Boto2 is considered a dead end by the Boto developers. This Boto3 backend is intended to be a clean break from Boto2 and not intended to be a complete drop in replacement, at least that's what jschneier agreed with when we merged Boto3 support in.

jschneier:
Thanks @mbarrien. This is a documentation failure more than anything else. 8b506f3c1f4a02ccbce41fcb04f226e7b58a517a

PR #269, opened 2017-02-19
Added project support to GCS through LibCloud

slightfoot:
Please see the docs here for reference to this paremter.
https://libcloud.readthedocs.io/en/latest/compute/drivers/gce.html

Perhaps the docs should be updated to reflect this.


slightfoot:
@jschneier thanks for the merge. libcloud docs mentioned above refer to this parameter as required, but you probably know better.

Just looking at my commit it might be better to do use the `provider` as the `extra_kwargs` default value. This would allow for optional parameters to be passed across from provider config to drivers.

jschneier:
Yeah I agree. What we currently have can only lead to madness.

PR #271, opened 2017-02-21
Document dependencies on boto and boto3

Flimm:
It took me a while to realise that `boto` is still required even when using `boto3`. I've updated the documentation to make this clearer.

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -17,11 +17,11 @@ in version 1.2
 Settings
 --------
 
-To use boto3 set::
+To use boto3, ensure that the Pypi packages ``boto`` and ``boto3`` are installed and set::
 
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
 
-To use the boto version of the backend set::
+To use the boto version of the backend, ensure that the Pypi package ``boto`` is installed and set::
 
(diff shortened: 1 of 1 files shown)


docs/backends/amazon-S3.rst
```diff
@@ -17,11 +17,11 @@ in version 1.2
 Settings
 --------
 
-To use boto3 set::
+To use boto3, ensure that the Pypi packages ``boto`` and ``boto3`` are installed and set::
```
mbarrien: S3Boto3Storage does not boto installed, only boto3.

Flimm: Nope, I just tried it, it needs `boto` installed, as well as `boto3`. Try it yourself.

mbarrien: As the original author of the boto3 version of the plugin (more like the one who ported it over from Boto), we have used this in production without boto2 for over a year.

jschneier:
Agreed with @mbarrien - just tested this with a fresh project. Boto3 does require botocore (and a few other libraries) but those are all dependencies of boto3 already.

Flimm:
I fully admit that boto3 may not in fact depend on boto.

Could we please document the expected dependencies on boto and/or boto3 somewhere? It might have tried to troubleshoot this harder if I realised that there must be a mistake on my end, rather than trying to guess what the intended dependencies are.

I would edit this pull request in GitHub's interface, but I can't seem to find the buttons to do it.

PR #273, opened 2017-03-01
Dropbox api upgrade

jschneier:


ZuluPro:
@jschneier : Sounds good, but not yet tested

PR #274, opened 2017-03-02
Use extras_require in setup for backend dependencies

mbarrien:
To help ease making it clear what dependencies need to be installed for a backend, this PR adds the backend dependencies to setup.py's extras_require. Instead of asking a user to install the requirements separately, a user can pip install (or add to requirements.txt) a string like:

```
pip install django-storages[boto3]
pip install django-storages[dropbox,sftp]   # Multiple backends
```

This partially addresses the issues raised in #271 (although documentation would need to be updated to note this). Users can still just install django-storages directly without the extras, and install their own dependencies, providing backwards compatibility.

Future backends will be responsible for updating the setup.py file for their own dependencies. Deprecated backends (e.g. image, mogilefs) are intentionally not included in this PR.


jschneier:
This is a great idea.

jleclanche:
@jschneier Looks good to land :) Anything blocking it?

setup.py
```diff
@@ -15,6 +15,16 @@ def get_requirements_tests():
     name='django-storages',
     version=storages.__version__,
     packages=['storages', 'storages.backends'],
+    extras_require={
+        'azure': ['azure'],
+        'boto': ['boto>=2.32.0'],
+        'boto3': ['boto3>=1.2.3'],
+        'database': ['pyodbc'],
+        'dropbox': ['dropbox>=3.24'],
```
jdufresne: Frmo `requirements-tests.txt`, the new min version is 7.2.1.

setup.py
```diff
@@ -15,6 +15,16 @@ def get_requirements_tests():
     name='django-storages',
     version=storages.__version__,
     packages=['storages', 'storages.backends'],
+    extras_require={
+        'azure': ['azure'],
+        'boto': ['boto>=2.32.0'],
+        'boto3': ['boto3>=1.2.3'],
+        'database': ['pyodbc'],
```
jdufresne: This backend was deprecated and removed.

setup.py
```diff
@@ -15,6 +15,16 @@ def get_requirements_tests():
...
+        'dropbox': ['dropbox>=3.24'],
+        'google': ['boto>=2.32.0'],
```
jdufresne: Should this be `google-cloud-storage>=0.22.0` instead?

PR #277, opened 2017-03-09
Azure: Use BlockBlobService for automatic handling of files bigger th…

cantorek:
…an 64MB

This will ensure that we don't have to worry about blob fragmentation and lib will do it for us automagically.

Tested with azure-storage-0.34.0

```
appdirs==1.4.3
azure==1.0.3
azure-common==1.1.4
azure-mgmt==0.20.2
azure-mgmt-common==0.20.0
azure-mgmt-compute==0.20.1
azure-mgmt-network==0.20.1
azure-mgmt-nspkg==1.0.0
azure-mgmt-resource==0.20.1
azure-mgmt-storage==0.20.0
azure-nspkg==1.0.0
azure-servicebus==0.20.1
azure-servicemanagement-legacy==0.20.2
azure-storage==0.34.0
cffi==1.9.1
cryptography==1.7.2
Django==1.10.6
django-storages==1.5.2
idna==2.5
packaging==16.8
pyasn1==0.2.3
pycparser==2.17
pyparsing==2.2.0
python-dateutil==2.6.0
requests==2.13.0
six==1.10.0
```

Will not work with azure-storage-0.20.3.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -11,18 +11,13 @@ from django.utils.deconstruct import deconstructible
 
 try:
     import azure  # noqa
+    from azure.storage.blob import BlockBlobService
+    from azure.storage.blob import ContentSettings
+    from azure.common import AzureMissingResourceHttpError
 except ImportError:
     raise ImproperlyConfigured(
         "Could not load Azure bindings. "
         "See https://github.com/WindowsAzure/azure-sdk-for-python")
-
-try:
-    # azure-storage 0.20.0
-    from azure.storage.blob.blobservice import BlobService
-    from azure.common import AzureMissingResourceHttpError
-except ImportError:
-    from azure.storage import BlobService
-    from azure import WindowsAzureMissingResourceError as AzureMissingResourceHttpError
 
 from storages.utils import setting
 
@@ -45,7 +40,7 @@ class AzureStorage(Storage):
     @property
     def connection(self):
         if self._connection is None:
-            self._connection = BlobService(
+            self._connection = BlockBlobService(
                 self.account_name, self.account_key)
         return self._connection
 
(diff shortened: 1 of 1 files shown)


jesushernandez:
+1 for this but we should make it backwards compatible as well (with 0.20.X)? Could you do that?

jschneier:
@jesushernandez does this functionality exist in the 0.20.x series? I emailed with the Microsoft people today and they told me that 0.20.x is no longer supported at all and shortly even installed it via `pip install azure` won't work.

jschneier:
It does exist.

jschneier:
Okay, this has not been touched in a very long time I see. The fallback isn't for the 0.3x series, it's for something extremely legacy.

storages/backends/azure_storage.py
```diff
@@ -93,9 +88,9 @@ def _save(self, name, content):
         else:
             content_data = content.read()
 
-        self.connection.put_blob(self.azure_container, name,
-                                 content_data, "BlockBlob",
-                                 x_ms_blob_content_type=content_type)
+        self.connection.create_blob_from_bytes(self.azure_container, name,
+                                 content_data,
+                                 content_settings=ContentSettings(content_type=content_type))
```
jschneier: This still reads everything into memory in the `if/else` above. I think if we just pass the file to `create_blob_from_file` it might _just work_.

storages/backends/azure_storage.py
```diff
@@ -11,19 +11,14 @@
 
 try:
     import azure  # noqa
+    from azure.storage.blob import BlockBlobService
+    from azure.storage.blob import ContentSettings
```
jschneier: Let's do these 2 imports on one line

sww314:
Azure backend was rewritten in #565.

Issue #282, opened 2017-03-15
django-storages support for AWS_SESSION_TOKEN for the boto backend


bxm156:
In AWS Lambda, you can assign IAM role to a function, which should describe the permissions allowed for the function.This is placed into the env, via 3 params:
* AWS_ACCESS_KEY_ID
* AWS_SECRET_ACCESS_KEY
* AWS_SESSION_TOKEN

Lambda is using STS, and all 3 must be supplied to the S3Connection object.

Normally, boto would handle this itself, but since we manually create the S3Connection, we need to handle this ourselves. This is achieved by passing security_token=<AWS_SESSION_TOKEN> into  the S3Connection object.

# Examples

## Without the session token
```
zappa invoke prod "import boto; from boto.s3.connection import S3Connection; import os; c = S3Connection(os.getenv('AWS_ACCESS_KEY_ID'), os.getenv('AWS_SECRET_ACCESS_KEY')); rs = c.get_all_buckets(); print rs[7].get_key('index.html');" --raw

=>

Calling invoke for environment prod..
START RequestId: fc2d8da1-0965-11e7-b4fc-45f80254b4aa Version: $LATEST
S3ResponseError: 403 Forbidden
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidAccessKeyId</Code><Message>The AWS Access Key Id you provided does not exist in our records.</Message><AWSAccessKeyId>**REDACTED**</AWSAccessKeyId><RequestId>9B0A00B299B5BD02</RequestId><HostId>x+cb94gePg+XCM119UdKNktXAI26fqrru7Ih7rpVqvQVJ+K4Z/tTSSsYDsGXUQ03XP+E8BHhgZo=</HostId></Error>: S3ResponseError
Traceback (most recent call last):
  File "/var/task/handler.py", line 507, in lambda_handler
    return LambdaHandler.lambda_handler(event, context)
  File "/var/task/handler.py", line 241, in lambda_handler
    return handler.handler(event, context)
  File "/var/task/handler.py", line 370, in handler
    exec(raw_command)
  File "<string>", line 1, in <module>
  File "/tmp/pip-build-LWM5xq/boto/boto/s3/connection.py", line 444, in get_all_buckets
S3ResponseError: S3ResponseError: 403 Forbidden
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #283, opened 2017-03-16
Add support for AWS_SESSION_TOKEN and AWS_SECURITY_TOKEN

bxm156:
Add support for reading the AWS_SESSION_TOKEN and AWS_SECURITY_TOKEN from the environment.

Verified this works as expected in AWS Lambda.

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -205,6 +205,7 @@ class S3BotoStorage(Storage):
     # used for looking up the access and secret key from env vars
     access_key_names = ['AWS_S3_ACCESS_KEY_ID', 'AWS_ACCESS_KEY_ID']
     secret_key_names = ['AWS_S3_SECRET_ACCESS_KEY', 'AWS_SECRET_ACCESS_KEY']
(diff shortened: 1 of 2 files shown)


bxm156:
A few questions:

1) Should this be behind a setting ?
2) Should we care if the there is a AWS_SESSION_TOKEN but no AWS_ACCESS/SECRET keys in the env?

AusIV:
I just encountered the same issue, and was glad to see someone had already done the legwork. I've also verified that this solves my problem on Lambda.

narfman0:
I've verified this in my project as well, thanks Bryan! Should fix #282 too.

jschneier:
Looks good!

jschneier:
@bxm156 re: your questions. The docs are overdue for an overhaul. Authentication should be given its own section. I think relying on just the environment variables is likely better in this case but I need to read the aws docs again just to make sure.

jschneier:
Oof. This only sets `.security_token` if it exists. I need to switch to live tests because you always try to reference the parameter even if it isn't getting set.

bxm156:
Ah, sorry about that, I was planning on writing some tests for this but never got around to it :(

PR #287, opened 2017-03-31
Fix _full_path function in dropbox.py

pberi:
The function _full_path() was unnecessarily creating absolute paths from the system home, and this path was indeed getting reflected inside the dropbox folder. 

For eg. when DROPBOX_ROOT_PATH = '/django-cms/', the function created a full path of C:\django-cms\<upload file path> inside the root dropbox folder. 

Also the error of 'backslashes not allowed' was getting encountered on the Windows platform.
This commit fixes both the above issues.

### DIFF
--- a/storages/backends/dropbox.py
+++ b/storages/backends/dropbox.py
@@ -18,12 +18,11 @@ from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import File
 from django.core.files.storage import Storage
 from django.utils.deconstruct import deconstructible
-from django.utils._os import safe_join
-
 from storages.utils import setting
 
 from dropbox.client import DropboxClient
 from dropbox.rest import ErrorResponse
+from os.path import join
 
 DATE_FORMAT = '%a, %d %b %Y %X +0000'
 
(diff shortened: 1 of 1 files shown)


jschneier:
Fixing the Windows file separator looks good. `.safe_join` is also there to protect against possible directory traversals. We probably need something that that we can reuse across backends.

pberi:
Thanks a lot for that feedback. I'll try to find a better reusable
solution.

On Apr 8, 2017 10:28 PM, "Josh Schneier" <notifications@github.com> wrote:

Fixing the Windows file separator looks good. .safe_join is also there to
protect against possible directory traversals. We probably need something
that that we can reuse across backends.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jschneier/django-storages/pull/287#issuecomment-292730842>,
or mute the thread

jschneier:
That would be great if you can but I'm happy to merge a fix first and then work out the best abstraction point later. Just don't want to merge something with a security possibility!

jschneier:
I just pushed e52a127523fdd5be50bb670ccad566c5d527f3d1 which should fix the Windows path issues. Still need to look at the other issue.

jschneier:
Finally figured this one out. You can't start with a `/`.

Issue #49, opened 2015-05-13
Files not overwritten when "collectstatic" run multiple times


mvanderlinde:
I believe this is the same issue that was reported in the original django-storages here: https://bitbucket.org/david/django-storages/issue/167

When I run "collectstatic" more than once within a few hours, no files are recognized as modified.

jaylett:
It almost certainly is. There's a [PR against the upstream](https://bitbucket.org/david/django-storages/pull-request/102/fixing-timezone-comparison-in-s3/diff) that looks credible to me.

CarlosBorroto:
+1, please look into the existing PR, we also got hit by this issue.

Thanks!

jaylett:
Note that Django's API for this is changing from 1.10 (with a deprecation path). The previous fix is still a good idea, but there's a new set of methods to implement that provide full timezone support (and so consistency with the rest of Django).

jschneier:
Fixed by #290

Issue #134, opened 2016-04-01
collectstatic is not overriding changed files


matheusjardimb:
collectstatic is not overriding changed files and it's related to changed date of files and timezone.
I found this issue https://bitbucket.org/david/django-storages/issues/167/collectstatic-is-not-overwriting-changed and someone recently complained about it too.

Files only get updated when I "touch" them in order to modify it's changed date to some day in the future.

I seems S3 is not sending proper changed date or the algorithm is not considering timezone

dcloud:
This looks like a duplicate of #49 (which references pr #84, fwiw).

jschneier:
Dupe #49, going to probably implement Django 1.10s logic for backward compat or else simply allow the new 1.10 api to take care of this,

PR #290, opened 2017-04-04
Fix collectstatic tz handling

jschneier:


jschneier:
Fixes #49, Refs #141, #134, #84 

Think this one is finally fixed for real, could use some help testing. Here are some people who previously expressed interest, should be quick to test:

@marcelchastain, @adam-thomas, @matheusjardimb, @dcloud, @mvanderlinde @CarlosBorroto

Going to merge in the next day or 2 so I can get a release out for Django 1.11 with `get_modified_time` in there.

PR #299, opened 2017-04-25
Fix azure_storage ignoring the mode string

mac-chaffee:
Currently, if you open an azure file with a mode string of "rt" or "r", you'll get a binary file anyway. This is because azure always returns the contents of the file as a "bytes" object, which when passed to the [ContentFile constructor](https://github.com/django/django/blob/master/django/core/files/base.py#L139), creates a file using BytesIO.

This becomes a problem for me because I need to parse CSV files from azure, and Python3's csv reader [will not work with files opened in binary mode](http://stackoverflow.com/questions/8515053/csv-error-iterator-should-return-strings-not-bytes)

So this change just selectively uses `encode` and `decode` depending on what mode the user wants. There's probably a more efficient/elegant way to do this, so I'm definitely open to changes!


eborche2:
I know this project is focused on S3, but would it be possible to get this merged in for Azure users? This is the solution I'm manually implementing at this time. 
Thank You

sww314:
Azure backend was rewritten in #565. Please reopen if the issue still exists.

PR #309, opened 2017-05-05
New setting GS_USE_STANDARD_REQUEST_URI

faxioman:
Using this setting you can directly retrieve a Google Cloud Storage object url using the standard request uri endpoint.
No API call is performed to obtain the public url for the object.

https://cloud.google.com/storage/docs/xml-api/reference-uris

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,9 @@ django-storages change log
 ******************
 
 * **Breaking:** Remove backends deprecated in v1.5.1 (`#280`_)
+* **Deprecation:** The undocumented ``gs.GSBotoStorage`` backend. See the new ``gcloud.GoogleCloudStorage``
+  or ``apache_libcloud.LibCloudStorage`` backends instead. (`#236`_)
+* Add a new backend, ``gcloud.GoogleCloudStorage`` based on the ``google-cloud`` bindings. (`#236`_ thanks @scjody)
 * Pass in the location constraint when auto creating a bucket (`#257`_, `#258`_ thanks @mattayes)
 * Add support for reading ``AWS_SESSION_TOKEN`` and ``AWS_SECURITY_TOKEN`` from the environment
   to ``S3Boto3Storage`` and ``S3BotoStorage``. (`#283`_ thanks @bxm156)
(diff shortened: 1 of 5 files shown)


jschneier:
Is one strictly better than the other? Wondering why we should add a setting.

faxioman:
Because it works only if you enable public access to your bucket (acl allUser - read)

faxioman:
mmm ... the blob "public_url" method also works only if public access is enabled. It's better to use my settings for retrieve public file ... but without my setting a user should be able to avoid enabling the pubblic access and retrieve a signed url:

```
bucket.get_blob(name).generate_signed_url(
            expiration=datetime.datetime.now() + datetime.timedelta(hours=1))
```
If you not agree with this implementation (eg, you think it's better to force the enabling of public url acl in gcloud) yes, probably the setting is unuseful and we should always use the standard request uri for url method.

faxioman:
Probably the safest choice is to use my method to retrieve the public url without the setting, and in the future enable the generation of signed url by means of a dedicated setting. What do you think?

svengt:
I think it would also be great if you could define a custom domain just like the S3 boto backend.
https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto.py#L457

faxioman:
@svengt ok, I'll proceed with a new PR

stinky2nine:
Wouldn't a better way to use `selfLink` that is available in the `blob` (GCS Object)? See https://cloud.google.com/storage/docs/json_api/v1/objects.

sww314:
Google Storages is deprecated.

faxioman:
@sww314 so django-storages will soon drop google cloud storage support?😱

sww314:
@faxioman Sorry I read this wrong. I thought it was for the old XML based gs.py. That storage backend is going away.

jschneier:
Okay so this discussion I have been now trying to centralize to #491 but since this is close to what I was considering doing for public urls I'd like your opinion if you have time @faxioman (sorry for the delay goes without saying).

Here's what they have in the Google Cloud storage codebase:

```python
@property
def public_url(self):
    """The public URL for this blob.

    Use :meth:`make_public` to enable anonymous access via the returned
    URL.

    :rtype: `string`
    :returns: The public URL for this blob.
    """
    return '{storage_base_url}/{bucket_name}/{quoted_name}'.format(
        storage_base_url=_API_ACCESS_ENDPOINT,
        bucket_name=self.bucket.name,
        quoted_name=quote(self.name.encode('utf-8')))
```

Question: What is the difference between the subdomain and path placement for the bucket.
Question: What sort of ACLs do you need to have enabled for the various functionalities? As you mentioned in one of your comments the proper way to do all this is similar to the boto backends which I have outlined in a comment on #491.

sww314:
Fixed in #570.

PR #320, opened 2017-05-17
[gcloud] Add support for MIME type

### LINKED ISSUE #303: [gcloud] set the correct Content-Type for file
gcloud backend uploads the file with default Content-Type so it is set to `application/octet-stream` on Google Cloud Storage.
Use this API https://googlecloudplatform.github.io/google-cloud-python/stable/storage-blobs.html#google.cloud.storage.blob.Blob.content_type to set the correct type (at least for `.css` and `.js` files).

Sgiath:
Resolves #303 Issue

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -1,3 +1,4 @@
+import mimetypes
 from tempfile import SpooledTemporaryFile
 
 from django.core.exceptions import ImproperlyConfigured
@@ -21,6 +22,7 @@ except ImportError:
 class GoogleCloudFile(File):
     def __init__(self, name, mode, storage):
         self.name = name
+        self.mime_type = mimetypes.guess_type(name)[0]
         self._mode = mode
         self._storage = storage
         self.blob = storage.bucket.get_blob(name)
@@ -70,7 +72,7 @@ class GoogleCloudFile(File):
         if self._file is not None:
             if self._is_dirty:
                 self.file.seek(0)
-                self.blob.upload_from_file(self.file)
+                self.blob.upload_from_file(self.file, content_type=self.mime_type)
             self._file.close()
             self._file = None
 
@@ -154,7 +156,8 @@ class GoogleCloudStorage(Storage):
         content.name = cleaned_name
         encoded_name = self._encode_name(name)
         file = GoogleCloudFile(encoded_name, 'rw', self)
-        file.blob.upload_from_file(content, size=content.size)
+        file.blob.upload_from_file(content, size=content.size,
+                                   content_type=file.mime_type)
         return cleaned_name
 
     def delete(self, name):
--- a/tests/test_gcloud.py
+++ b/tests/test_gcloud.py
@@ -6,6 +6,7 @@ except ImportError:  # Python 3.2 and below
     import mock
 
 import datetime
+import mimetypes
 
 from django.core.files.base import ContentFile
 from django.test import TestCase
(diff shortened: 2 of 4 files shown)


Sgiath:
I also checked that this is working on my site.

svengt:
Why not use the `mimetypes.guess_type` method, just like the s3boto backend? https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto.py#L364
https://docs.python.org/3.6/library/mimetypes.html

Sgiath:
@svengt I did not know about this function. Thanks for it! I fixed it now.

Sgiath:
@jschneier or @scjody can you please Merge it or tell me what should be fixed? Thanks

jleclanche:
Looks fine with me but I don't know gcloud well enough to decide on it.

It needs rebasing, by the way. Please squash your commits together and rebase on master.

storages/backends/gcloud.py
```diff
@@ -20,6 +21,7 @@
 class GoogleCloudFile(File):
     def __init__(self, name, mode, storage):
         self.name = name
+        self.mime_type = mimetypes.guess_type(name)[0]
```
scjody: Since `guess_type` isn't 100% reliable, I think we need to allow the user to override the type here if needed. Can you add that?

storages/backends/gcloud.py
```diff
@@ -153,7 +155,7 @@ def _save(self, name, content):
         content.name = cleaned_name
         encoded_name = self._encode_name(name)
         file = GoogleCloudFile(encoded_name, 'rw', self)
-        file.blob.upload_from_file(content, size=content.size)
+        file.blob.upload_from_file(content, size=content.size, content_type=file.mime_type)
```
scjody: It's better to wrap lines at 79 columns (although the [Django Coding Style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/) allows up to 119 if needed for readability). Here you can put `content_type=...` on the next line without hurting readability, so please do.

tests/test_gcloud.py
```diff
@@ -91,7 +92,7 @@ def test_open_write(self, MockBlob):
         # File data is not actually written until close(), so do that.
         f.close()
 
-        MockBlob().upload_from_file.assert_called_with(tmpfile)
+        MockBlob().upload_from_file.assert_called_with(tmpfile, content_type=mimetypes.guess_type(self.filename)[0])
```
scjody: Here it would be preferable to put `content_type` on the next line. This will still make the line longer than 79 characters but that's OK.

storages/backends/gcloud.py
Sgiath: When the user should override it? Globally at the Django settings (set the desired MIME type for specific file extension)? Something like:
```
GS_MIME_TYPE = {
    '.css': 'text/css',
    '.mp4': 'video/mp4',
}
```

scjody: Now that I've looked more closely, I don't see a good place to add this. Django's storage API doesn't have any support for MIME types as far as I can tell, so we're well within our rights to set whatever we want when saving a file to the storage provider.

jschneier:
I'm okay with this change in principle across the storages. Since this is something that has been being fixed piecemeal in all of the cloud backends I wonder if there is a nicer general solution.

Sgiath:
OK squashed and rebased. You can merge it.

PR #322, opened 2017-05-19
Fix S3 safe_join() to allow colons

### LINKED ISSUE #248: S3BotoStorage fails when keys contain ":"
Trying to do this:
`storage.delete("2017-01-13T12:55:32.497821+00:00/somekey")`
Which fails like so:
```
  File "...", line 76, in ...
    storage.delete("2017-01-13T12:55:32.497821+00:00/somekey")
  File ".../local/lib/python2.7/site-packages/storages/backends/s3boto.py", line 438, in delete
  File "/.../local/lib/python2.7/site-packages/storages/backends/s3boto.py", line 364, in _normalize_name
    name)
django.core.exceptions.SuspiciousOperation: Attempted access to "2017-01-13T12:55:32.497821+00:00/somekey" denied.
```

I have tracked this down to [storages.backends.s3boto.safe_join](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto.py#L54)'s call to `urlparse.urljoin`. Having ":" confuses the join to skip the `/` at the beginning of `final_path` which then fails during the check that `final_path` is a sub path of `base`.

It's probably a small fix, but I unfortunately don't have enough knowledge of the urlparse lib to know how to do it. 

PS.
And yes, AWS supports having ":" part of the keys. Running `aws s3 rm s3://<bucket>/2017-01-13T12:55:32.497821+00:00/somekey` works fine.

jdufresne:
Combine the identical s3boto3 and s3boto implementations of safe_join()
and its tests to reduce code duplication.

Fixes #248

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -22,46 +22,13 @@ except ImportError:
     raise ImproperlyConfigured("Could not load Boto3's S3 bindings.\n"
                                "See https://github.com/boto/boto3")
 
-from storages.utils import setting
+from storages.utils import setting, safe_join
 
 boto3_version_info = tuple([int(i) for i in boto3_version.split('.')])
 
 if boto3_version_info[:2] < (1, 2):
     raise ImproperlyConfigured("The installed Boto3 library must be 1.2.0 or "
                                "higher.\nSee https://github.com/boto/boto3")
-
-
-def safe_join(base, *paths):
-    """
-    A version of django.utils._os.safe_join for S3 paths.
-
-    Joins one or more path components to the base path component
-    intelligently. Returns a normalized version of the final path.
-
-    The final path must be located inside of the base path component
-    (otherwise a ValueError is raised).
-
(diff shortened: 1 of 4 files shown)


jleclanche:
Why is this using urljoin at all instead of `os.path.join`? Clearly the input isn't URL path components, otherwise behaviour would be broken (as it is).

jdufresne:
Good question. I think using `posixpath.join()` does make more sense and actually simplifies the implementation. I have updated the PR with this change.

jleclanche:
Looks good. Can you think of any other scenarios not covered with the current tests? Changing this method feels a little stressful all in all.

jdufresne:
Can't think of any at the moment. I added a new test for `utils.safe_join("base", "/etc/passwd")` as well as fix a typo in test `test_trailing_slash_multi()`. If you can think of another test to write, I'll happily add it. Let me know.

jleclanche:
Happy with this.

Issue #118, opened 2016-02-01
If "~/.ssh/known_hosts" does not exist, SFTP backend fails


jonazn:
In the SFTPStorage backend, the _connect() method attempts to "automatically add host keys from the current user," by loading the "~/.ssh/known_hosts" file. However, it doesn't check if that file actually exists before trying to use it. I spent a while trying to figure out why the SFTP backend wasn't working on my own server before pinpointing this issue. This happens on line 97 of sftpstorage.py:

`self._ssh.load_host_keys(os.path.expanduser(os.path.join("~", ".ssh", "known_hosts")))`

Perhaps it would be useful to wrap that line in a try/except?

jschneier:
Yeah we should definitely catch that error.

Hmm, the default is to automatically accept unknown hosts which seems...eager? My thinking is to deprecate that default and then if we do catch an error and you haven't told us to auto add I'll just raise a full on exception. What do you think of that plan?

jonazn:
That sounds like a reasonable solution, but there are also times you'd maybe want that behavior? For example, the reason I'm using SFTP is so that I can have an inexpensive DigitalOcean server to host all my media files, but I'm using AWS for everything else (S3 data transfer is just too expensive). My Django app will upload user files to the DigitalOcean server. But EC2 instances that spin up with Elastic Beanstalk don't have a known_hosts file by default, nor is there an easy, built-in way to set one. I'm hacking together a solution by simply running the command "touch known_hosts" in ~/.ssh every time a new EC2 instance starts up to have an empty file.

Perhaps a solution could be an additional setting that, if set, lets you accept unknown hosts? It could default to False.

jschneier:
Oh yeah, of course. I'm not even sure how I feel about the fact that it automatically loads the keys in `~/.ssh/knows_hosts`.

Not sure how big of a deal it is though.

freshquiz:
This issue just bit me, but in a roundabout way.

In the `SFTPStorage.exists()` method there is try/catch `IOError` around this `self.sftp.stat(remote_path)` and that call goes down to hit this issue.

So that try/catch is catching the wrong exception and returning False, which causes infinite recursion when trying to create a directory that doesn't exist.

sallyruthstruik:
Pull request with fix: https://github.com/jschneier/django-storages/pull/325

jschneier:
Fix merged

PR #325, opened 2017-06-02
Fix issue #118

sallyruthstruik:
Fixed error described in https://github.com/jschneier/django-storages/issues/118

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ setuptools*
 __pycache__
 .coverage
 .cache
+.idea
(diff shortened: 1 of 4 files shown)


jschneier:
Thanks for your contribution! This looks good although I sort of itch to want to remove this auto-adding at all.

jschneier:
Note that there is a minor functionality change where a non-existent file could have been passed in and we'd fall back. I count that as a bug anyway so no big deal.

sallyruthstruik:
I found critical bug: when $HOME/.ssh/known hosts exists, whatever None is passed in load_known_hosts. Fixed in this PR: https://github.com/jschneier/django-storages/pull/330

PR #326, opened 2017-06-02
Generate Signed URL

sheepeatingtaz:
The deprecated gs backend had the ability to generate a signed URL, which is useful for sensitive data and required for non-public buckets.

This simple addition of setting an expiry time in the settings tells the system to generate a signed URL instead of the normal public URL.


sww314:
Overlaps with #399

PR #330, opened 2017-06-07
Fix minor bug from PR #325

sallyruthstruik:
There are little bug in https://github.com/jschneier/django-storages/pull/325 PR. 
When $HOME/.ssh/known hosts exists, whatever None is passed in load_known_hosts. 

The PR was already merged, so I've decide to open new one with bug test and fix


jschneier:
Yup, thanks.

PR #334, opened 2017-06-16
Avoid incorrect behaviour of S3BotoStorage.exists

lukeburden:
Calling `S3BotoStorage._save` writes to `_entries` if `preload_metadata`
is True. Future access to `_entries` is limited as the property `entries` only
loads metadata for the bucket if `_entries` is empty, which due to `_save`
it may not be. This in turn causes `exists` to incorrectly return False for
existent files.

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -233,6 +233,7 @@ class S3BotoStorage(Storage):
         self._entries = {}
         self._bucket = None
         self._connection = None
+        self._loaded_meta = False
 
         self.security_token = None
         if not self.access_key and not self.secret_key:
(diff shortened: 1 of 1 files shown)


storages/backends/s3boto.py
```diff
@@ -270,9 +270,14 @@ def entries(self):
         """
         Get the locally cached files for the bucket.
         """
-        if self.preload_metadata and not self._entries:
-            self._entries = dict((self._decode_name(entry.key), entry)
-                                 for entry in self.bucket.list(prefix=self.location))
+        if self.preload_metadata and not getattr(self, '_loaded_meta', False):
```
jleclanche: Set `_loaded_meta` to False in `__init__` instead of using getattr.

lukeburden:
Hello @jleclanche - your request has been fulfilled.

PR #336, opened 2017-06-21
Fix safe_join to allow joining a base path with an empty string

jdufresne:
Regression introduced in 895a068fcf3f23a6b71294c5b4b67a822df6e642

---

This regression was my fault. If the PR looks good, I'm hoping this can go out in a bug fix release as 1.6.0 has broken some of my projects. Thanks!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 django-storages change log
 ==========================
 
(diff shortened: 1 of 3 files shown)


jschneier:
Okay can release point release tonight.

jschneier:
I uploaded 1.6.1

PR #338, opened 2017-06-22
Integrate isort and use it to consistently sort imports across files

jdufresne:
Reduces need to think and order imports, simply let the tool do it. Provides more consistent code style across Python modules by always ordering imports a single way.

Use same configuration as Django project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache: pip
 matrix:
   include:
     - env: TOX_ENV=flake8
+    - env: TOX_ENV=isort
     - python: 2.7
       env: TOX_ENV=py27-django18
     - python: 3.3
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,3 +6,12 @@ exclude =
     .tox,
     docs
 max-line-length = 119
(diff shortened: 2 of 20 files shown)


jleclanche:
👍 but there's no need to create a separate tox env for it, add it to flake8.

jdufresne:
I was following configurations of other projects as examples which typically include isort as a separate tox environment. For example:

* [Django](https://github.com/django/django/blob/master/tox.ini#L52-L57)
* [Django Debug Toolbar](https://github.com/jazzband/django-debug-toolbar/blob/master/tox.ini#L39-L42)
* [isort](https://github.com/timothycrosley/isort/blob/develop/tox.ini#L11-L12)

jleclanche:
I remain unconvinced. It's all codestyle checking, there's no real reason to split it up (in all the projects I manage, we either use flake8-import-order where it *has* to be part of the flake8 run, or we keep isort and qa/lint/flake8 together)

jdufresne:
Thanks for the suggestion. I've integrated this in PR #345.

PR #340, opened 2017-06-22
Fix GSConnection getting security_token kwarg

svleeuwen:
`GSConnection` doesn't expect the security_token kwarg like `S3Connection`, causing a `TypeError`.
This change fixes this issue.


jschneier:
Hi I opened #342 with an alternate fix, can you take that for a test drive?

jschneier:
Nevermind I was able to reproduce it trivially, will merge in my fix and release.

jschneier:
I released 1.6.2 with this fixed, thanks for the report.

svleeuwen:
That should work, thanks for the fix.

PR #341, opened 2017-06-22
Fix safe_join to handle trailing slash with intermediate paths

jdufresne:
Regression introduced in 895a068fcf3f23a6b71294c5b4b67a822df6e642

Partially reverts 39a2a7a8c08eadd2952cab485ea76434ae86b12f

---

@jschneier Apologies for the 2nd round of regression fixes. I believe I've covered all edge cases with this fix. I have added _many_ more unit tests and more thoroughly tested against my own projects. All feedback welcome. If the diff looks good, I'd appreciate if a new release could get cut as my previous changes to `safe_join()` is broken. Thanks!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 django-storages change log
 ==========================
 
+1.6.2 (UNRELEASED)
+******************
+
+* Fix regression in ``safe_join()`` to handle a trailing slash in an
+  intermediate path.
+
 1.6.1 (2017-06-22)
 ******************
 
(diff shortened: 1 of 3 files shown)


jschneier:
This project has been critically under-tested since I obtained, part of the reason I am hesitant to release so it's not that big of a surprise. Thanks for the vigilance.

jleclanche:
@jschneier don't be hesitant to make fast releases! IMHO it'll only get more testing if the pypi tip is kept close to master.

jschneier:
I released 1.6.2 with this fix.

jschneier:
@jleclanche yeah, a good point.

Issue #301, opened 2017-05-03
Can't upload large files (>200 MB) to Dropbox


maxmalysh:
This is what I receive when trying to upload a large file:
  
    ErrorResponse: [413] Error parsing response body or headers: 
    Body - b'<!DOCTYPE html>\n<html>\n<head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"> 
    Headers - HTTPHeaderDict({'Date': 'Wed, 03 May 2017 03:07:21 GMT', 'Server': 'nginx', 'Content-Type': 'text/html', 'X-Robots-Tag': 'noindex, nofollow, noimageindex', 'Content-Length': '25546', 'Connection': 'close', 'X-Dropbox-Request-Id': 'f5a26cc8ee0d984112ef3131868893d2', 'ETag': '"58ed15a2-63ca"'})

The reason is that `django-storages` uses `files_upload` API method to upload files, and documentation recommends against using it to upload files larger than 150 MB. 

There is a topic on Dropbox forum describing exactly the same problem in the Perl API wrapper: https://www.dropboxforum.com/t5/API-support/413-response-with-a-353M-video-and-HTML-garbage-response/td-p/203635

It looks likes something like `ChunkedUploader` should be used instead.

PR #302, opened 2017-05-03
Dropbox api upgrade

patroqueeet:
adressing #301

tests not proven as #227#issuecomment-298834277

but backup worked for me

PR #346, opened 2017-06-24
Update for dropbox api v2

scps950707:
I fix three problems I met,
1. From [dropbox-python-sdk](https://github.com/dropbox/dropbox-sdk-python/releases/tag/v7.1.0):"File upload endpoints no longer accept file-like objects. Only byte strings are allowed", so I modify the argument of `files_upload`
2. From the discussion [here](https://www.dropboxforum.com/t5/API-support/Python-API-V2-Metadata-of-root-folder/m-p/228433#M12382), `files_get_metadata` API v2 is not allowed to check root folder's metadata, so I replace it with `files_list_folder` to modify `listdir` function
3. According to official doc, `files_upload` isn't used to upload single file larger than 150MB, so I use `files_upload_session_start`,`files_upload_session_append` and `files_upload_session_finish` to implement large file upload, I tested files range from 600~800MB and works well, resolves issue #301

### DIFF
--- a/storages/backends/dropbox.py
+++ b/storages/backends/dropbox.py
@@ -21,7 +21,7 @@ from django.utils._os import safe_join
 from django.utils.deconstruct import deconstructible
 from dropbox import Dropbox
 from dropbox.exceptions import ApiError
-from dropbox.files import FileMetadata, FolderMetadata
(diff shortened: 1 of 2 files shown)


scps950707:
Sorry, I didn't find out these works done in #302

PR #347, opened 2017-06-24
Fix Dropbox tests to work with version 8.0.0

jdufresne:
The client module was removed from the library.


jschneier:
Yeah that was oversight on my part when I did the v2 fixes. Thanks.

PR #348, opened 2017-06-26
Simplify S3Boto3Storage.exists() by using boto3's head_object()

jdufresne:
Avoids unnecessary API requests to fetch and create the bucket when simply checking if a key exists.

https://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.head_object


jleclanche:
👍

PR #352, opened 2017-06-28
Rewrite S3Boto3Storage.listdir() for efficiency

### LINKED ISSUE #95: Inefficient implementation of `listdir` in the S3 backend

jdufresne:
Use boto3 paginator to fetch results. Code based on django-s3-storage.

Fixes #95

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -520,25 +520,22 @@ class S3Boto3Storage(Storage):
             return False
 
     def listdir(self, name):
-        name = self._normalize_name(self._clean_name(name))
-        # for the bucket.objects.filter and logic below name needs to end in /
-        # But for the root path "" we leave it as an empty string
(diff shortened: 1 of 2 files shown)


jschneier:
@jdufresne is this still relevant?

jdufresne:
It should be more efficient than the previous version, so I think so. Did you have specific concerns or reservations?

PR #354, opened 2017-07-02
Allow specification of GS_LOCATION for placing assets in subdirectory.

prdonahue:


prdonahue:
Apologies—didn't mean to submit PR here, just as starting point to my own fork. Will re-open once I've implemented location support for Google Cloud Storage (basing it off how S3 uses).

PR #358, opened 2017-07-07
Create a session/connection per thread for s3boto3

### LINKED ISSUE #268: S3Boto3Storage still showing multithreading issues in 1.5.2
I'm using `S3Boto3Storage` in 1.5.2 with uWSGI & 8 threads, and my workers are exiting like so:

    app_1    | DAMN ! worker 2 (pid: 15) died, killed by signal 11 :( trying respawn ... 
    app_1    | Respawned uWSGI worker 2 (new pid: 20)

Unfortunately I'm not getting a stacktrace. I saw the bug fix that went into 1.5.2 in the release notes (#238) and, on a whim, switched to 1 thread per worker, and the segfaults went away. I also tried `S3BotoStorage` with multiple threads, and it works fine.

Happy to help debug this if I can, but I'm not entirely sure where to start.

tomkins:
Documentation for boto3 recommends a session per thread - https://boto3.readthedocs.io/en/latest/guide/resources.html#multithreading-multiprocessing

As the storage class is (usually) only instantiated once per process, we need to set a thread local for each thread/connection used.

Fixes #268

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -1,6 +1,7 @@
 import mimetypes
 import os
 import posixpath
+import threading
 from gzip import GzipFile
 from tempfile import SpooledTemporaryFile
 
@@ -236,7 +237,7 @@ class S3Boto3Storage(Storage):
 
         self._entries = {}
         self._bucket = None
-        self._connection = None
+        self._connections = threading.local()
 
         self.security_token = None
         if not self.access_key and not self.secret_key:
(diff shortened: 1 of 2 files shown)


jschneier:
Thanks! This looks good and fixes a longstanding annoyance. Thoughts on the name `connections`? Considering that the pattern uses it as a singleton I'm not sure if that name is appropriate.

tomkins:
Seemed like the right thing at the time, as it felt like a thread local stores multiple things across various threads. But it should probably be more representative of what it's storing in the thread.

Shall I switch it back to `_connection`?

jschneier:
I think you're right, we will merge it as is.

PR #365, opened 2017-07-24
fix(s3boto3): spool buffer file to end of all uploaded parts after each part upload.

### LINKED ISSUE #364: Problem with S3Boto3StorageFile and a file greater than 5Mb
Similar to issue #160 but for S3Boto3. Fix would be also very similar

tommwatson:
Fixes #364, similar issue to #160 for s3boto3. Inspired by vinayinvicible's fix for #160.


jschneier:
Consolidated and fixed in #504.

PR #366, opened 2017-07-25
Update S3Boto3Storage class by updating size method to read the file …

mabuaisha:
Update size method of ``S3Boto3Storage`` to read size using ``size`` and not ``content_lenght``


mabuaisha:
@jschneier I add a fix for getting size of file when using boto3 and I need that fix so that django-storage and boto3 can working

jschneier:
Merged in 4501f992ae5ac12d094a153c01ef37e5c3c04a1f. Version bump will come soon, please use master until then. Thanks for the report and patch.

mabuaisha:
@jschneier Thanks for merging this to master

mabuaisha:
@jschneier It seems the commit fix did not fix the issue because the following method 
```
    def size(self, name):
        if self.entries:
            entry = self.entries.get(name)
            if entry:
                return entry.size
            return 0
        return self.bucket.Object(self._encode_name(name)).content_length
```
``entry`` instance could be a type of the following ``s3.Object`` which need to read from ``content_length`` or  it could be an instance type of s3.ObjectSummary`` which need to read from ``size``

So to solve the above issue I think we need to check the instance type before reading from size 

```
if isinstance(entry, s3.Object):
    return entry.content_length

elif isinstance(entry, s3.ObjectSummary)
      return entry.size
```

Please let me know what do you think ?

jschneier:
I'm wondering if this is a bug in boto3 that might depend on your version. What version of `boto3` are you using?

jschneier:
Nevermind, it appears to be intentional. We can use a `hasattr` check.

jschneier:
Pushed 523ab59696061cb1f940cb4fe75d58e68426fe45.

mabuaisha:
@jschneier I'm using ``boto3==1.4.4``

jschneier:
@mabuaisha that was not relevant -- sorry about that. i pushed a fix that duck types the method so it should work now. it is included in 1.6.4 which i released 2 days ago as well.

mabuaisha:
@jschneier Thanks for your released to handle that :)

PR #370, opened 2017-07-27
Always lookup the security token

jschneier:


storages/backends/s3boto.py
```diff
@@ -238,10 +238,9 @@ def __init__(self, acl=None, bucket=None, **settings):
         self._connection = None
         self._loaded_meta = False
 
-        self.security_token = None
+        self.security_token = self._get_security_token()
```
aarcro: This will be less surprising and broken. So that setting KEY_ID and SECRET in settings.py doesn't prevent the ability to use the TOKEN. 

BUT, I would set `security_token` at line 180 following the same pattern as `access_key` and `secret_key`

jschneier:
Agreed

PR #373, opened 2017-07-31
Ensure that we're reading the file from the start

mr-bo-jangles:
This is needed for Django 1.11 as we can no longer assume that the file we're looking is starting from the beginning, this can cause issues with empty files being saved instead of GZ variants.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -387,6 +387,7 @@ class S3Boto3Storage(Storage):
 
     def _compress_content(self, content):
         """Gzip a given string content."""
+        content.seek(0)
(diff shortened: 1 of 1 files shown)


jschneier:
I've just merged a failing test case for this issue, can you rebase to confirm that this then fixes that and then I would be very happy to merge.

jschneier:
Verified locally, thanks.

humphrey:
So good!  I've just spent 3 days trying to solve this zero length file issue (ok, 2 days of that was trying to use WebPack as an alternative to django-pipeline)... and I just found that it was this line of code, and hour after the commit!  Yay!  It works now!

PR #375, opened 2017-08-03
GoogleCloudStorage: Don't list all objects to get directories

alexcwatt:
I ran into performance issues with directories with thousands of objects because the current implementation of `listdir` gets all objects matching the folder prefix, and then determines the folders by checking for a `/` in the filename. This implementation is a lot faster.

I just got tox set up locally and realize that the tests aren't passing for this; basically the HTTPIterator object that is actually used has a `prefixes` property, but the original tests just construct a list of blobs instead of an object that gives the blobs for `list(iterator)` and the directories for `iterator.prefixes`. Any thoughts on the best way to do this? I'd really love to update the tests and get this patch merged.


stinky2nine:
@alexcwatt Are you still working on this?

alexcwatt:
@stinky2nine I wasn't sure the best way to rewrite the test. If I have time soon, I'll try to do that. However, I know this implementation works; it is running in production on a project where the older implementation was causing massive performance issues.

stinky2nine:
It shouldn't be that hard using mocks library as you can mock google.cloud.storage operations.

alexcwatt:
@jschneier Is there anything else I should do so we can get this merged? I've been using this in production for a project for almost a year now, but only recently revisited this PR to update the tests. Thanks for this awesome collection of Django storage adapters!

PR #376, opened 2017-08-06
Send copy of ContentFile.file to upload_fileobj.

jpmelos:
This will stop upload_fileobj from modifying the internal file. This is important because the default storage mixin for hashed static files that comes in Django expects files to be open after they are saved, and
upload_fileobj closes them.


jschneier:
Sorry, is this another fix for #371? Seems like it's dealing with the same issue.

jpmelos:
It could be, but I'm not sure. I'll verify and come back to you. Thanks for taking a look.

jschneier:
Merged #754 to fix this instead, thanks for your contribution.

Issue #378, opened 2017-08-10
Dropbox storage throws an exception when saving a file


maxmalysh:
Dropbox storage backend throws an exception when saving a file. I guess it stopped working after the API upgrade (see e46f3067fbc0ba9f8d1c351026a59aa9e0a390eb and #302).

How to reproduce:

    $ touch foo.txt
    $ echo "hello, world" >> foo.txt
    $ python3 manage.py shell
    >>> from storages.backends.dropbox import DropBoxStorage
    >>> dropbox_key = 'OAUTH_ACCESS_TOKEN'
    >>> dropbox_storage = DropBoxStorage(dropbox_key)
    >>> 
    >>> f = open('foo.txt', 'rb')
    >>> dropbox_storage.save('foo.txt', f)

```
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/code.py", line 91, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/[...]/venv/lib/python3.5/site-packages/django/core/files/storage.py", line 54, in save
    return self._save(name, content)
  File "/[...]/venv/lib/python3.5/site-packages/storages/backends/dropbox.py", line 111, in _save
    self.client.files_upload(content, self._full_path(name))
  File "/[...]/venv/lib/python3.5/site-packages/dropbox/base.py", line 1267, in files_upload
    f,
  File "/[...]/venv/lib/python3.5/site-packages/dropbox/dropbox.py", line 234, in request
    timeout=timeout)
  File "/[...]/venv/lib/python3.5/site-packages/dropbox/dropbox.py", line 325, in request_json_string_with_retry
    timeout=timeout)
  File "/[...]/venv/lib/python3.5/site-packages/dropbox/dropbox.py", line 369, in request_json_string
    type(request_binary))
TypeError: expected request_binary as binary type, got <class 'django.core.files.base.File'>
```

I've already found the bug: `DropBoxStorage._save` receives a `File` or any file-like object (passed as the `content` argument), and then passes it straight to the `self.client.files_upload` method. However, `DropboxBase.files_upload` accepts only `bytes`.

I'll provide a pull request fixing both this and #301, if you don't mind.

jschneier:
Please do, thanks.

Lrcezimbra:
Has this fix already been released? Do you have a date?

Thanks.

[...]

PR #379, opened 2017-08-10
Dropbox: large file support and a regression fix

maxmalysh:
This PR:

- fixes a regression in `DropBoxStorage._save` introduced with e46f306 (see https://github.com/jschneier/django-storages/issues/378);
- adds large file support for Dropbox (see https://github.com/jschneier/django-storages/issues/301).

I've tested uploading a 1.5 GB file and it works fine.

### DIFF
--- a/storages/backends/dropbox.py
+++ b/storages/backends/dropbox.py
@@ -108,5 +108,7 @@ class DropBoxStorage(Storage):
         return remote_file
 
     def _save(self, name, content):
-        self.client.files_upload(content, self._full_path(name))
+        content.open()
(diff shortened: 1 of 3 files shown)


storages/backends/dropbox.py
```diff
@@ -21,6 +21,7 @@
 from django.utils.deconstruct import deconstructible
 from dropbox import Dropbox
 from dropbox.exceptions import ApiError
+from dropbox.files import UploadSessionCursor, CommitInfo
```
maxmalysh: The only failing build is lint. Change import order to make it happy:

    from dropbox.files import CommitInfo, UploadSessionCursor

Issue #306, opened 2017-05-04
[gcloud] fix path generating for object


Sgiath:
The backend generates paths like this: `<link rel="stylesheet" type="text/css" href="https://storage.googleapis.com/<bucket-name>/admin%2Fcss%2Fdashboard.css">`
instead of this: `<link rel="stylesheet" type="text/css" href="https://storage.googleapis.com/<bucket-name>/admin/css/dashboard.css">`

It is working but looks odd. Consider using this API https://googlecloudplatform.github.io/google-cloud-python/stable/storage-blobs.html#google.cloud.storage.blob.Blob.public_url

jschneier:
Are you using the new `gcloud` backend (still in master, not yet released)? It is using `.public_url`.

Sgiath:
Yes I am installing the `django-channels` from git and using the new `gcloud` backend.

svengt:
I also ran into this problem.

whoknowwhat:
So do I. Should I use `gs` instead of `gcloud`?

whoknowwhat:
Still have same problem with 1.6.5

```
google-cloud-storage==1.3.1
django-storages==1.6.5
Django==1.11
```

jschneier:
Sorry for the delay, this was fixed upstream.

PR #388, opened 2017-09-01
Fixing path encoding on gcloud #306

gregmccoy:
Simply running the url through `unquote_plus()` so that it no longer contains the `%2F`.


de-don:
Why not merged it? :(

jschneier:
Sorry for the delay, this was fixed upstream.

PR #392, opened 2017-09-08
Fix inconsistent capitalization of "Dropbox"

davidbstein:
The "b" is capitalized in some but not all places. "Dropbox" is one word, not two.
 https://www.dropbox.com/branding#in-writing

Might not be worth merging if you're worried about backwards compatibility (or I can add some class aliases).


sww314:
@davidbstein Prefer not to break backward compatibility.

PR #405, opened 2017-10-11
added GS_AUTO_CREATE_STORAGE_CLASS, GS_AUTO_CREATE_LOCATION settings 

f1aky:
I could not fix the tests for gcloud, can someone else will do it


sww314:
The storage class should not be responsible for creating the buckets. It makes the most common case retrieving a file more complicated and adds additional network trips to Google. Instead, create your bucket outside the storage class. You need dynamic bucket creation a more explict approach would work better. It also causes permissions problems as the SerivceAccount needs more permissions than it should (#412).

jschneier:
I am going to add this concept as a recipe to the docs. This functionality has negative performance aspects for the far more common use cases.

PR #407, opened 2017-10-14
Detect Content-Type when content_type property is null

### LINKED ISSUE #406: S3BotoStorage doesn't properly fallback to other Content-Type heuristics
Content Type is determined by the following:

1. a property called content-type
2. the filename
3. a default is used

If the `content-type` property doesn't exist, the logic falls back to 2 and 3. But the logic doesn't properly fallback if the content_type exists, but it set to None. 

I found that in django my InMemoryFileUpload had a content_type of None.  And so django-storages would try to make a s3 upload via boto3 with the following params {"Content-Type:": None}

This would cause boto to crash with 

```Invalid type for parameter ContentType, value: None, type: <class 'NoneType'>, valid types: <class 'str'>```

bxm156:
Fallback on other content-type detection methods when the content_type property is None
Fixes #406

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -378,8 +378,8 @@ class S3BotoStorage(Storage):
         name = self._normalize_name(cleaned_name)
         headers = self.headers.copy()
         _type, encoding = mimetypes.guess_type(name)
-        content_type = getattr(content, 'content_type',
-                               _type or self.key_class.DefaultContentType)
+        content_type = getattr(content, 'content_type', None)
+        content_type = content_type or _type or self.key_class.DefaultContentType
 
         # setting the content_type in the key object is not enough.
         headers.update({'Content-Type': content_type})
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -419,8 +419,8 @@ class S3Boto3Storage(Storage):
         name = self._normalize_name(cleaned_name)
         parameters = self.object_parameters.copy()
         _type, encoding = mimetypes.guess_type(name)
-        content_type = getattr(content, 'content_type',
-                               _type or self.default_content_type)
+        content_type = getattr(content, 'content_type', None)
+        content_type = content_type or _type or self.default_content_type
 
         # setting the content_type in the key object is not enough.
         parameters.update({'ContentType': content_type})
--- a/tests/test_s3boto.py
+++ b/tests/test_s3boto.py
@@ -66,6 +66,26 @@ class S3BotoStorageTests(S3BotoTestCase):
             rewind=True
         )
 
+    def test_content_type(self):
+        """
+        Test saving a file with a None content type.
+        """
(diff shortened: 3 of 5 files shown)


bxm156:
The failure is unrelated to my changes as far as I can tell. Perhaps a cache issue with pip on Travis CI?

bxm156:
@jschneier Updated with a fix for failing tests on master.

paramiko now requires the enum library, which wasn't added to Python 3 until Python 3.4. I updated tox to include the dependency for python 3.3 tests.

alexhayes:
Works for me also - thanks @bxm156

AnderUstarroz:
@jschneier this commit fixes an issue that is giving headaches to many developers, any chance to get it merged into master?

Well done @bxm156

bxm156:
Perhaps @jleclanche could help get this merged?

jleclanche:
Sure. I can't make releases though.

bhrutledge:
This resolves an [issue plaguing Django CMS users](https://github.com/divio/django-filer/issues/1014). @jschneier Any thoughts on a release?

jschneier:
Thanks for the ping, sure. @jleclanche thanks for the assist on the maintenance. I'm close to getting to the point where I can spend 1-2 full days/month on open source.

bxm156:
Thanks @jschneier! 

Side Question:
What work do you think is needed in order to make this project viable for something like Jazzband?

jschneier:
Absolutely.

Would like to get it a place where this is: https://github.com/etianen/django-s3-storage

In particular the live testing of pull requests with real credentials. It’s too easy to break things using mocks since this library is all about using the actual underlying library and what is placed in the provider. Ideally I’d like to make some base class (maybe BaseCloudStorageBackend?) which can clean up some of the duplicated logic across backends/make it easier to add new ones (someone mentioned Digital Ocean for example a few months ago I think). Whether or not ftp/sftp gets pulled out at that point is pretty arbitrary. Also would deprecate s3boto.py (not boto3 obviously) and then remove that + gs.py. So quite a lot to do altogether.

jleclanche:
FWIW I don't think any of those things are required to get into jazzband. Ci and an easy to run test suite so that pull requests can be merged without too deep an understanding of the codebase is what's important.

Regarding testing with real credentials, it's certainly nice to have the possibility, but in CI I recommend "real" mock implementations. Either stuff like [moto](https://github.com/spulec/moto) or [minio](https://github.com/minio/minio) for S3, for example.

jschneier:
I recall breaking the Dropbox backend (1 of many I believe) because we were mocking out the client library. S3 probably has sufficient adoption that full mock libraries exist for it but I don’t think that goes for the others. And then when we had that s3v4 vs s3 issue (something along those lines) that was mostly evident with different regions (aka not just Virginia)…which is truthfully further than I was going to go regardless.

bhrutledge:
@jschneier Thank you for the prompt response! It was a pleasant surprise.

PR #414, opened 2017-10-28
Include license file in the generated wheel package

jdufresne:
The wheel package format supports including the license file. This is done using the `[metadata]` section in the `setup.cfg` file. For additional information on this feature, see:

https://wheel.readthedocs.io/en/stable/index.html#including-the-license-in-the-generated-wheel-file

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,3 +15,6 @@ known_first_party = storages
 line_length = 79
 multi_line_output = 5
 not_skip = __init__.py
+
+[metadata]
+license_file = LICENSE


jdufresne:
Rebased so travis is green.

jdufresne:
Ping? Without this, django-storages isn't in compliance with its own license. From `LICENSE`:

> Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

jschneier:
Hmm, should probably be compatible with our own license.

PR #417, opened 2017-10-31
Use extras_require in setup for backend dependencies -- take II

browniebroke:
Taking over work from #274, which appear abandoned and addressing the comments from the code review there. I've tried [to update that pull request](https://github.com/mbarrien/django-storages/pull/2), but getting it merged requires action from the original contributor @mbarrien.


browniebroke:
Any chance we could get this merged @jleclanche ? 

It's fixing the build as well as adding a nice new way to install the package once the next release is out. I depend on `boto3` only through `django-storages`, would be nicer to install it as extra...

jleclanche:
py33 is not supported anymore, but ok

PR #422, opened 2017-11-08
New option GS_USE_STATIC_URL_AS_BASE_URL to use Django settings' STATIC_URL as the base URL

### LINKED ISSUE #385: Use CDN URL to show gcloud images
Hello!

We are using storages.backends.gcloud.GoogleCloudStorage to handle our images. We are uploading to a Google Cloud segment, and so far so good: The images are uploaded fast and are shown with no issues. The problem is we want the images to be retrieved using our CDN URL, not the https://storage.googleapis.com/ URL that is used by default.

I have dived into the docs and code, but haven't seen a solution. Just the issue #309, but I'm not sure if it is about the same situation we have. Is there any option or workaround we can use?

Thanks for your time!

stinky2nine:
Fix #385.

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -66,6 +66,13 @@ By default files with the same name will overwrite each other. Set this to ``Fal
 The maximum amount of memory a returned file can take up before being
 rolled over into a temporary file on disk. Default is 0: Do not roll over.
 
+``GS_USE_STATIC_URL_AS_BASE_URL`` (optional: default is ``False``)
+
+If you would like to use Django settings' ``STATIC_URL`` as the base URL, then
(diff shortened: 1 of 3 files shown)


stinky2nine:
Decided not to proceed with this pull request.

zisk0:
Hello!

We are thinking about patching django-storages with this PR, as it implement a feature I am looking for. Can you tell me a little why you didn't proceed? Just in case I am missing something :)

Thank you!

stinky2nine:
Check pull #423 and #424, using the `public_url` value instead (https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html?highlight=public_url#google.cloud.storage.blob.Blob.public_url).

PR #423, opened 2017-11-08
Use Django's cache framework to cache gcloud file URLs

### LINKED ISSUE #413: Why listing image urls with DRF using GCS is so slow?
I'm using google-cloud-storage==1.3.2 + django-storages==1.6.5. Currently, I'm able to upload images to my bucket using the settings provided in the documentation, get listing my images in a view is extremely slow.

I have something like:
`class Picture(models.Model):
    created = models.DateTimeField(auto_now_add=True)
    image = models.ImageField(upload_to='picture')
`
I'm using DRF with a serializer like this one:
`class PictureSerializer(serializers.ModelSerializer):
    class Meta:
        model = Picture
        fields = ('id', 'image')
`
Each time I try to get the list of my Pictures, which should only provide something like:
`
 {
        "id": 1,
        "image": "https://storage.googleapis.com/bucket/pic1.jpg"
    },
    {
        "id": 2,
                "image": "        "image": "https://storage.googleapis.com/bucket/pic1.jpg""
    },
`
it takes a lot of time. My assumption is that it needs to resolve the **https** route to my bucket, and this should be done using the gcloud-api, but it would be nice to have a fixed(fast) URL for it since it's always the same: "https://storage.googleapis.com/bucket/picture_id.jpg"

What should be the approach for solving this issue?
On my frontend, I have enabled lazy loading for the pictures, but this URL solving time-consuming issue is not helping.

stinky2nine:
Fix #413

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,6 +32,7 @@ By order of apparition, thanks:
     * Stanislav Kaledin (Bug fixes in SFTPStorage)
     * Filip Vavera (Google Cloud MIME types support)
     * Max Malysh (Dropbox large file support)
+    * Alex Yu (Google Cloud Storage enhancements)
 
 Extra thanks to Marty for adding this in Django,
 you can buy his very interesting book (Pro Django).
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -1,6 +1,7 @@
 import mimetypes
 from tempfile import SpooledTemporaryFile
 
+from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import File
 from django.core.files.storage import Storage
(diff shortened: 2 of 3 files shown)


stinky2nine:
The failing tests have nothing to do with this pull request.

stinky2nine:
Ping, anyone?

stinky2nine:
Why is this a bad implementation? Regardless, the first call is already slow. The existing implementation retrieves the object every single time. Sure, https://googlecloudplatform.github.io/google-cloud-python/latest/storage/buckets.html#google.cloud.storage.bucket.Bucket.path, can be used to generate the URL but it doesn't validate whether the file object exists which is a change of the original behavior.

If you are OK to change the original behavior, I can easily create a new pull to use google.cloud.storage.bucket.Bucket.path.

PR #424, opened 2017-11-09
bucket.exists should return False if file is not found on the Google Cloud Storage

stinky2nine:
sorl-thumbnail for example validates whether the thumbnail for any given image exists on the server before regeneration occurs. `exists` should always return True or False and not throw any exceptions.

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -171,9 +171,11 @@ class GoogleCloudStorage(Storage):
                 return True
             except ImproperlyConfigured:
                 return False
-
         name = self._normalize_name(clean_name(name))
-        return bool(self.bucket.get_blob(self._encode_name(name)))
+        try:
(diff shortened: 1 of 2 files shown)


stinky2nine:
Ping, anyone?

sww314:
@stinky2nine Would it be cleaner is the code called the exists API from Google?
https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.exists

stinky2nine:
@sww314 It has been so long and I just took another quick look. I am not if sure if it will be any cleaner.

```python
if self.bucket.exists():
  return bool(self.bucket.get_blob(self._encode_name(name)))
else:
  return False
```

I have no strong opinions about this.

jschneier:
I don't see why this is needed at all. `bucket.get_blob` can't raise `NotFound`.

PR #427, opened 2017-11-15
Use tox environment variable TOXENV as supported by tox

jdufresne:
tox reads to environment variable `TOXENV` (not `TOX_ENV`) to decide what environment to test. Modify to fall in line with upstream. As a result, can use this feature instead of passing the environment through the `-e`
CLI argument.

More idiomatic with other project's use of tox and Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,29 +4,29 @@ cache: pip
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


jdufresne:
This was fixed in cecb101a8eacf3dda708e251471890941fc994a8

PR #445, opened 2018-01-06
Prefer readthedocs.io instead of readthedocs.org for doc links

jdufresne:
Read the Docs moved to hosting at readthedocs.io instead of readthedocs.org. Fix all links in the project.

For additional details, see:

https://blog.readthedocs.com/securing-subdomains/


jdufresne:
Rebased so travis is green.

Issue #447, opened 2018-01-11
Add support for the 'Infrequent Access' S3 Storage Class


tveastman:
Currently, django-storages supports the STANDARD storage class and the REDUCED_REDUNDANCY storage class.

Amazon appear to be phasing out REDUCED_REDUNDANCY, and in fact, it's more expensive now than STANDARD.

There's a new class, "infrequent access" (`STANDARD_IA`) for which we should add support.

This'll be easy in `s3boto3.py`, but it might be trickier with `s3boto.py`

jschneier:
Closing in favor of #549. Thanks for the suggestion.

PR #448, opened 2018-01-11
Add support for the 'STANDARD_IA' storage class to s3boto3.py

tveastman:
Adding support for `STANDARD_IA` as per #447.

The simplest thing that could possibly work. Please note that it's not yet actually tested to confirm that it *does* work.


jschneier:
Thanks for the PR. I'm going to close this in favor of deprecating the reduced redundancy setting and adding a configuration value for the storage class since there are now 5 of them.

PR #453, opened 2018-01-20
Add option to detect content encoding to fix gzip uploads.  Resolves #451

### LINKED ISSUE #451: Uploading .gz file to S3 results in file being stored uncompressed
I have compressed log files I'm trying to store in S3 using django-storages s3 backend, but when I inspect the files I discover that S3 stored them in their uncompressed form.  I have done some digging and discovered that django-storages is properly identifying my files as gzipped, but setting that as a ContentEncoding argument so that S3 interprets the data as gzip for HTTP transfer encoding and uncompresses it at the HTTP layer at put time.

Content encoding detection:
https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L421-L433
In the _save() method encoding is detected with mimetypes.guess_type() which in my case results in 'gzip'.

If I were to set AWS_IS_GZIPPED and add my gzip file type to GZIP_CONTENT_TYPES then it is going to attempt to gzip my data a second time for transfer when self._compress_content() is called.  This feels undesirable with 100MB compressed log files

The bucket object `obj` that is created is used to call upload_fileobj with the ContentEncoding parameter which eventually calls Client.put_object with ContentEncoding set to gzip
http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_object

I can create a pull request, that would fix this for my use case, but I would like to understand what is expected and what option is least likely to break other people's use cases.
Options for fixing this:

1. Detect the .gz file extension and do not set the ContentEncoding value
2. Disable ContentEncoding detection entirely if AWS_IS_GZIPPED == False - Is there a use case for locally compressed assets like stylesheets that need to be decompressed on S3?
3. Other options I haven't thought of?

If you have any thoughts on what kind of approach you would like to see taken I can take them and get a pull request submitted for review.

skruger:
The behavior selected in the resolution of #263 has broken django storages when trying to upload normal gzip files without double compressing them or when `AWS_IS_GZIPPED=False`.

`AWS_DETECT_CONTENT_ENCODING` has been added to allow the behavior that was introduced last year to be used when desired.  If `AWS_DETECT_CONTENT_ENCODING=True` and `AWS_IS_GZIPPED=False` gzip encoded files will be stored in their uncompressed form on S3.  The detect content encoding default in the storage classes is `False` which reverses the behavior introduced in pull request #264.  This default of `False` would appear to be the least confusing to someone in a new project trying to use django-storages to upload a .gz file.

This pull request modifies the existing unit test and adds one new unit test to verify the behavior of both AWS_DETECT_CONTENT_ENCODING states in both the boto and boto3 based storage classes.  The new configuration option has also been added to the documentation.

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -102,6 +102,9 @@ Available are numerous settings. It should be especially noted the following:
 ``AWS_IS_GZIPPED`` (optional: default is ``False``)
     Whether or not to enable gzipping of content types specified by ``GZIP_CONTENT_TYPES``
 
+``AWS_DETECT_CONTENT_ENCODING`` (optional: default is ``False``)
+    Detect content encoding on files and set ContentEncoding headers accordingly when sending to S3.  Files gzipped before being sent to S3 will be stored uncompressed.
+
 ``GZIP_CONTENT_TYPES`` (optional: default is ``text/css``, ``text/javascript``, ``application/javascript``, ``application/x-javascript``, ``image/svg+xml``)
     When ``AWS_IS_GZIPPED`` is set to ``True`` the content types which will be gzipped
 
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -199,6 +199,7 @@ class S3BotoStorage(Storage):
     secure_urls = setting('AWS_S3_SECURE_URLS', True)
     file_name_charset = setting('AWS_S3_FILE_NAME_CHARSET', 'utf-8')
     gzip = setting('AWS_IS_GZIPPED', False)
+    detect_content_encoding = setting('AWS_DETECT_CONTENT_ENCODING', False)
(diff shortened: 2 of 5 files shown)


skruger:
The Django 1.8 on Python 3.3 unit test appears to be failing because it installed paramiko 2.4.0 which is officially incompatible with Python 3.3.  According to the categories metadata on pypi.python.org paramiko 2.3.1 does support python 3.3 so forcing the lower version for python 3.3 installations might work to resolve this unit test failure.

skruger:
All tests are passing except for the error resolved by pull request #415

sww314:
@skruger  can you rebase this fix?

skruger:
@sww314 rebase complete. There were some rough conflicts.

skruger:
@sww314 I saw that there were some new merge conflicts so I rebased again.  Still ready for merge or further review.

PR #457, opened 2018-01-30
Add KMS Encryption support to S3Boto3Storage

linuxlewis:
This PR makes a backwards compatible change to add support for the other s3 encryption type using kms encryption key ids.


jleclanche:
Can you add tests for this please?

jschneier:
Given the comment from @mbarrien sounds like we can deprecate some of the parameter settings and tell people to use `AWS_S3_OBJECT_PARAMETERS`.

jschneier:
Rather than special casing everything I am reducing it all to just using `AWS_S3_OBJECT_PARAMETERS`. Furthermore I am adding a dedicated override point so that you can craft the value based on the uploaded object.

PR #458, opened 2018-02-06
Unset boto access creds to let boto do its fallbacks

myoung34:


myoung34:
This lets boto do its thing and use sessions properly.

myoung34:
Tests fail for pieces not related to my change.

jschneier:
Can't you just not set any of the credentials in this case? And I don't think region is handled from a fallback.

alanjds:
Region fallback envvar is `AWS_DEFAULT_REGION`. But the stack of boto3 fallbacks is looong: https://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials

>The order in which Boto3 searches for credentials is:
>
>1. Passing credentials as parameters in the boto.client() method
>1. Passing credentials as parameters when creating a Session object
>1. Environment variables
>1. Shared credential file (~/.aws/credentials)
>1. AWS config file (~/.aws/config)
>1. Assume Role provider
>1. Boto2 config file (/etc/boto.cfg and ~/.boto)
>1. Instance metadata service on an Amazon EC2 instance that has an IAM role configured.

jschneier:
And the defaults are ``None``. This would break this for a lot of people. I agree that we can probably make the documentation clearer about what needs to be set in your settings file vs what ``boto3`` will pull from automatically but I don't understand why this PR is necessary. Can you explain?

jschneier:
I also agree that we don't need to look things up from the env for the user since boto3 will do that automatically. But removing the values that come from `AWS_S3*` will definitely break things. Any explanation about what I'm missing in that logic is most welcome.

alanjds:
@jschneier The main issue is: If you provide `AWS_S3*` settings as None, that is the default right now, Boto will not fallback to 3+ on this list:

>The order in which Boto3 searches for credentials is:
>
>1. Passing credentials as parameters in the boto.client() method
>1. Passing credentials as parameters when creating a Session object
>1. Environment variables
>1. Shared credential file (~/.aws/credentials)
>1. AWS config file (~/.aws/config)
>1. Assume Role provider
>1. Boto2 config file (/etc/boto.cfg and ~/.boto)
>1. Instance metadata service on an Amazon EC2 instance that has an IAM role configured.

My suggestion is to, OR do not use the AWS_S3* settings and tell people to put everything on the environment or files (`3. Environment variables` and others) OR to pass to `boto.client()` only the `AWS_S3*` settings explicitly set.

This PR implements the former. More logic is needed to implement the latter, but would be good too.

Issue #231, opened 2016-11-23
Add customization callback for setting metadata attributes


lucasmg-dev:
I need to be able to set AWS_HEADER for a separate file.

For example:

```
photos = models.ImageField(upload_to = 'photos') # max-age = 60
photos2 = models.ImageField(upload_to = 'photos') # max-age = 3600
```

darkowic:
I had an idea for it. Even do some PR for it and we use it in our company. But it is not great. And do not cover all cases and all possible configurations. Check it it https://github.com/jschneier/django-storages/pull/152

darkowic:
@jschneier what do you think to enhance `AWS_HEADER` setting by

1) additional keys with path to file
```
AWS_HEADERS = {
'*': { # or 'DEFAULT' etc
    'Expires': 'Thu, 15 Apr 2010 20:00:00 GMT',
    'Cache-Control': 'max-age=86400',
},
'admindata/pages': {
"Cache-Control": "max-age=86400" 
}
}
```

2)  implementation that detect setting based on model - like in http://easy-thumbnails.readthedocs.io/en/2.1/usage/#thumbnail-aliases

```
AWS_HEADERS = {
    <target>: {
       "Cache-Control": "max-age=86400" 
    },
    ...
}
```

> Use the target '' for project-wide aliases. Otherwise, the target should be a string which defines the scope of the contained aliases:
'sprocket.Widget.image' would make the aliases available to only the ‘image’ field of a ‘Widget’ model in an app named ‘sprocket’.
'sprocket.Widget' would apply to any field in the ‘Widget’ model.
'sprocket' would target any field in any model in the app.

jschneier:
@darkowic what do you think of instead doing this as a callback? There are a number of projects that I use that allow you to specify a path to a function and we'll merge the results of that function into the parameters. I'm not sure how flexible the regex based solution is.

darkowic:
Sorry for late reply @jschneier  I like the callback idea. But where do you want to place the callback? We need to check for header's setting in file's write and storage save method. What I remember that in the code we have two different logic for setting headers for saving and write. How do you imagine it?

So, what we could do in the function? only update headers or work on the whole bucket key?

jschneier:
This was fixed for AWS

PR #459, opened 2018-02-09
New settings GS_CACHE_CONTROL to define caching policy via the Cache-Control HTTP header

cerny-jan:
If you use gcloud as a storage in your django project you usually want to leverage browser caching and this will allow you to do that.


sww314:
The default for this should be None (or no cache). 
The way it coded now it allows public caching of files which may not be what people intended.

cerny-jan:
@sww314 fair point, I've updated it

erikkn:
Is it possible to create different cache-control headers for different type of files?

jschneier:
Not sure why #505 didn't end up pushing here (I was trying to force push the rebase here). @erikkn that sounds like #231, I would take a PR for that.

PR #461, opened 2018-02-23
add AZURE_2018_SDK switch to make it work against current Azure SDK

depahelix:
I tested this in Django to write to Azure Blob storage.  I added it with a configuration setting so if you don't add AZURE_2018_SDK=True to your settings.py file, everything works the exact same as before.


jleclanche:
Please don't increase version numbers in PRs.

sww314:
Closed via #565. If not fixed in this update please reopen.

PR #468, opened 2018-03-12
Fixed an error 404 File does not exist.

### LINKED ISSUE #467: NotFound server error for GoogleCloudStorage instead of browser 404 response
In a case if an static file is missing on the storage, we will get  NotFound server error in url method.
I think in this case we should not raise NotFound exception and return something that will trigger browser 404 response.

An example what we have for now:
screen shot 2018-03-12 at 9 01 40 pm

mikekeda:
I Think in the case if we don't have requested file instead of raising NotFound exception in url method we should return something that will trigger browser 404 response (just file name).

fix #467


jschneier:
There is ongoing discussion in #491 where for public files we are not going to hit Google anymore. Please feel free to chime in over there.

PR #469, opened 2018-03-14
Boto/Boto3 metadata customization

roxel:
I added `update_headers` and `update_parameters` methods in Boto/Boto3 storages. These methods allow to implement metadata customization for specific files in inheriting classes without rewriting save method. The way this customization is performed can depend on project needs, but in a most basic scenario you can just use an if-else statement checking saved file name and contents.

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -405,20 +405,13 @@ class S3BotoStorage(Storage):
     def _save(self, name, content):
         cleaned_name = self._clean_name(name)
         name = self._normalize_name(cleaned_name)
-        headers = self.headers.copy()
         _type, encoding = mimetypes.guess_type(name)
         content_type = getattr(content, 'content_type', None)
         content_type = content_type or _type or self.key_class.DefaultContentType
 
-        # setting the content_type in the key object is not enough.
-        headers.update({'Content-Type': content_type})
-
         if self.gzip and content_type in self.gzip_content_types:
             content = self._compress_content(content)
-            headers.update({'Content-Encoding': 'gzip'})
-        elif encoding:
-            # If the content already has a particular encoding, set it
-            headers.update({'Content-Encoding': encoding})
+            encoding = 'gzip'
 
         content.name = cleaned_name
         encoded_name = self._encode_name(name)
@@ -430,7 +423,7 @@ class S3BotoStorage(Storage):
             key.last_modified = datetime.utcnow().strftime(ISO8601)
 
         key.set_metadata('Content-Type', content_type)
-        headers = self.update_headers(headers, name, content)
+        headers = self.get_headers(name, content, content_type, encoding)
         self._save_content(key, content, headers=headers)
         return cleaned_name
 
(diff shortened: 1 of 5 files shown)


storages/backends/s3boto.py
```diff
@@ -494,3 +495,9 @@ def get_available_name(self, name, max_length=None):
             name = self._clean_name(name)
             return name
         return super(S3BotoStorage, self).get_available_name(name, max_length)
+
+    def update_headers(self, headers, name=None, content=None):
+        """
+        Method that can be used to adjust file parameters.
+        """
+        return headers
```
mikekeda: Probably we can move headers update functionality from _save() into this method

```
    def update_headers(self, name=None, content=None, content_type=None):
        """
        Method that can be used to adjust file parameters.
        """
        headers = self.headers.copy()

        # setting the content_type in the key object is not enough.
        headers.update({'Content-Type': content_type})
 
        if self.gzip and content_type in self.gzip_content_types:
            headers.update({'Content-Encoding': 'gzip'})
        elif encoding:
            # If the content already has a particular encoding, set it
            headers.update({'Content-Encoding': encoding})

        return headers
```

storages/backends/s3boto3.py
```diff
@@ -580,3 +581,9 @@ def get_available_name(self, name, max_length=None):
             name = self._clean_name(name)
             return name
         return super(S3Boto3Storage, self).get_available_name(name, max_length)
+
+    def update_parameters(self, parameters, obj=None, content=None):
+        """
+        Method that can be used to adjust file parameters.
+        """
+        return parameters
```
mikekeda: Probably we can move parameters update functionality from _save() into this method

```
    def update_parameters(self, obj=None, content=None, content_type=None):
         """
         Method that can be used to adjust file parameters.
         """
        parameters = self.object_parameters.copy()

         # setting the content_type in the key object is not enough.
         parameters.update({'ContentType': content_type})
 
         if self.gzip and content_type in self.gzip_content_types:
             parameters.update({'ContentEncoding': 'gzip'})
         elif encoding:
             # If the content already has a particular encoding, set it
             parameters.update({'ContentEncoding': encoding})

         return parameters
```

storages/backends/s3boto.py
```diff
@@ -494,3 +495,9 @@ def get_available_name(self, name, max_length=None):
             name = self._clean_name(name)
             return name
         return super(S3BotoStorage, self).get_available_name(name, max_length)
+
+    def update_headers(self, headers, name=None, content=None):
+        """
+        Method that can be used to adjust file parameters.
```
mikekeda: should it be "Method that can be used to adjust file **headers**."?

roxel: yeah, right, I'll change that

storages/backends/s3boto.py
roxel: yes. this approach would make `save` cleaner. I'm conviced.

roxel:
@mikekeda what do you think about current implementation?

roxel:
@jschneier Is it something we can merge to master and release soon? Or do you think some updates/fixes are needed here?

jschneier:
The general idea was right and I implemented something similar in #828.

PR #470, opened 2018-03-14
Update changelog with changes on master.

bxm156:
Add a new section for tracking changes on master. Should be helpful to start tracking this info since there hasn't been a new release in a while.


sww314:
Already in another commit

PR #473, opened 2018-03-20
If `file_overwrite` is set, construct an in-memory key without hitting S3.

gomlgs:


jschneier:
That backend is deprecated and in feature freeze.

PR #460, opened 2018-02-19
Add GS_DEFAULT_ACL to Google Cloud Backend

sww314:
Add ability to set the ACL for each Blob/File as it is added. This allows the bucket to be private and the files to be publically readable.

PR #477, opened 2018-04-10
gcloud: Set predefined_acl when saving file

nkhumphreys:
The predefined_acl should be applied to files, not just the bucket when
its created (if auto create is on).

For example, if the acl is set to `publicRead` in the settings, this
should be applied to files even if the bucket already existed.


sww314:
@nkhumphreys Please only make one change per pull request. 
#460 provides same functionality. Setting the default ACL for each file. It makes a different ACL for files vs buckets (so you can have public files in in private bucket).

PR #402, opened 2017-10-05
Dropbox 8.x support

danni:
Fixes support for recent versions of the Dropbox SDK.

The API was radically different to the previous SDK, but the mocks used in the unit tests made it so the unit tests appeared to continue passing. This updates the backend to use the return types in the new SDK.

Fixes #396 and #397.

PR #478, opened 2018-04-12
Fix size in dropbox storage

zuck:
Probably a typo


jschneier:
Closing in favor of #402.

PR #484, opened 2018-04-15
Implement CustomDomainSignedS3Boto3Storage

### LINKED ISSUE #456: Public and private objects on S3 via Cloudfront?
The django-storages docs just say I should set the `AWS_S3_CUSTOM_DOMAIN` setting and configure Cloudfront. The Cloudfront docs say I can restrict access and require signed URLs.

I already use signed S3 URLs with django-storages, and these URLs have `Expires`, `Signature` and `AWSAccessKeyId` querystring params. The Cloudfront docs say signed cloudfront URLs need `Expires`, `Signature` and `Key-Pair-Id` querystring params.

How can I get `django-storages` to generate a signed Cloudfront URL? Is it even possible?

mattjegan:
Provides support for signed requests to custom urls (e.g. cloudfront). The tests provided are the same as the S3Boto3Storage tests.

Fixes #456


cburger:
@mattjegan Thank you for creating this PR. I was looking for a solution to get signed Cloudfront URLs. Just one question. Why does the custom domain have to end with a slash? I have tested the standard storage class and the domain does not need a slash. Since it is a "domain" and not a URL, I would say it should not have a slash?

mattjegan:
Hey @cburger, thanks for pointing that out. When I get the chance to update the PR I will however I can't put a time limit on it at the moment.

PR #486, opened 2018-04-18
Alllow set verify parameter for boto3

Smosker:
Boto3 has parameter `verify` but currently `django-storages` doesn't allow to set it.
http://boto3.readthedocs.io/en/latest/reference/core/session.html#boto3.session.Session.client
I implement such feature because i need to set path to my internal certificate in order to work with internal s3 endpoint.


jschneier:
Hello can you rebase, then I will merge.

Smosker:
done

jschneier:
I meant [rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) but no worries I have merged this commit in #580. Thanks for opening the issue and providing the fix.

PR #497, opened 2018-05-08
New logo for Django-Storages

reallinfo:


jschneier:
Hmm I added the logo files https://github.com/jschneier/django-storages/commit/7064f73676f552d893a8fc2f428e50e6a463b412 here that I grabbed out of the Drive. Will also add the svg to the same spot. Then I need to figure out how to get it into RST or just bite the bullet and rewrite the README in markdown. Thanks so much again.

PR #498, opened 2018-05-14
S3 storage backend is not `@override_settings` friendly

luanfonceca:
I tried to implement a test for a scenario that only happens on s3 and when i added the [`@override_settings`](https://docs.djangoproject.com/en/2.0/topics/testing/tools/#django.test.override_settings) decorator it didn't work. I looked over the `storages/backends/s3boto3.py` file and figured out that all the configuration settings imports are being made on python Importing time and not on Running time, if it was at Running time the Django decorator would work properly.

That should happen on other backends too.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -174,7 +174,11 @@ class S3Boto3Storage(Storage):
     # If config provided in init, signature_version and addressing_style settings/args are ignored.
     config = None
 
+    
     def __init__(self, acl=None, bucket=None, **settings):
+        self.access_key_names = ['AWS_S3_ACCESS_KEY_ID', 'AWS_ACCESS_KEY_ID']
+        self.secret_key_names = ['AWS_S3_SECRET_ACCESS_KEY', 'AWS_SECRET_ACCESS_KEY']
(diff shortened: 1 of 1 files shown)


jschneier:
Yeah. This is probably the single biggest issue with the original design of this library. Fixing this would be a major breaking change (basically would want to move to some sort of `DJ_STORAGES` var in settings) and I'd want to couple with fixing the myriad of other technical issues in this codebase.

jdufresne:
I've hit this problem myself. However, this approach would break the following snippet:

```py
class PrivateS3Boto3Storage(S3Boto3Storage):
    default_acl = 'private'
```

Could probably use something like `hasattr` to solve.

A test would be nice. Here is a potential one:

```py
    def test_override_settings(self):
        with override_settings(AWS_LOCATION='foo1'):
            storage = s3boto3.S3Boto3Storage()
            self.assertEqual(storage.location, 'foo1')
        with override_settings(AWS_LOCATION='foo2'):
            storage = s3boto3.S3Boto3Storage()
            self.assertEqual(storage.location, 'foo2')
```

jdufresne:
I've created a more complete version of this change that fixes all necessary backends and adds tests at #524

sww314:
This needs to be addressed for all storage backends. Being worked in #524.

PR #450, opened 2018-01-15
S3 large multipart uploads

tveastman:
This pull request fixes the multipart-upload bug described in #449, the download `.read()` issue described in #383, and adds a safeguard to prevent opening the file in `rw` mode, which at the moment would lead to corruption of the object in S3.

Issue #383, opened 2017-08-21
Amazon S3: can't get file chunk without loading whole file into memory


user0007:
Example:

    DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"

    from django.core.files.storage import default_storage
    default_storage.open("mykey").read(1024)

causes MemoryError to me, because it try to load ~5GB file.

It's because in a `S3Boto3StorageFile._get_file` method the whole file is loaded into memory every time we request it:

      ...
      self._is_dirty = False
      self._file.write(self.obj.get()['Body'].read())
      self._file.seek(0)

But boto3 supports reading file by chunk:

    s3.Object('mybucket', 'mykey').get()['Body'].read(1024)

the only way to workaround is:

    default_storage.open("mykey").obj.get()["Body"].read(1024)

but it's not generic anymore.

jeffcjohnson:
I'm having the same trouble.  I'm trying to download a 2.6 GB file with the following code but the open() call tries to read the entire file into memory before I get to call chunks().

```python
def download_storage_file(file_field: FileField, dest_path: pathlib.Path):
    with file_field.open(mode='rb') as reader:
        with open(dest_path, 'wb+') as writer:
            for chunk in reader.chunks():
                writer.write(chunk)
```

yunmanger1:
@jeffcjohnson nice catch.

I will leave this traceback here for those googling for solution
 
~~~
 File "/venv/lib/python2.7/site-packages/storages/backends/s3boto3.py", line 123, in _get_file
    self._file.write(self.obj.get()['Body'].read())
  File "/venv/lib/python2.7/site-packages/botocore/response.py", line 74, in read
    chunk = self._raw_stream.read(amt)
  File "/venv/lib/python2.7/site-packages/botocore/vendored/requests/packages/urllib3/response.py", line 239, in read
    data = self._fp.read()
  File "/usr/lib/python2.7/httplib.py", line 581, in read
    s = self._safe_read(self.length)
  File "/usr/lib/python2.7/httplib.py", line 693, in _safe_read
    return ''.join(s)
MemoryError
~~~

dummerbd:
For those who've come across this issue and don't want to wait for the PR to be merged, checkout `django-s3-storage`, they're already doing this the right way: https://github.com/etianen/django-s3-storage/blob/master/django_s3_storage/storage.py#L216

scheparev-moberries:
[...]

PR #504, opened 2018-05-30
Fix data-corruption issue with s3boto and s3boto3 multipart uploads

### LINKED ISSUE #160: Problem with S3BotoStorageFile and a file greater than 5Mb

### LINKED ISSUE #364: Problem with S3Boto3StorageFile and a file greater than 5Mb

### LINKED ISSUE #449: Multi-part uploads using `s3boto3.py` are corrupt.
There is a bug in the multi-part upload mechanism. The entire buffer is uploaded as a 'part' every time the file is written to. I think this is because the temporary file (`self.file`) is not being truncated when `_flush_write_buffer()` is called.

Once the temporary buffer is larger than `AWS_S3_FILE_BUFFER_SIZE`, every subsequent write uploads the WHOLE file as a 'part'.

Here's an example:
```
pipenv run python demo.py
INFO:root:Size of 'test-file.txt' is 10485760 bytes. (10 megabytes)
INFO:root:Writing to <S3Boto3StorageFile: test-file.txt> in chunks of 1048576 bytes (1 megabyte).
INFO:root:Upload complete, checking size of file in S3 Bucket
INFO:root:Size of 's3://kx-tom-misc-test-bucket/test-file.txt' is 47185920 bytes. (45 megabytes)
```

I uploaded a 10 megabyte file in 1 megabyte chunks, and ended up with a 45 megabyte file in S3.

Instead of 2 parts of 5 megabytes each, I have parts sized `5 + 6 + 7 + 8 + 9 + 10 == 45`

Code for demo at: https://gist.github.com/tveastman/9d15076da4f4f0646c9ce4b0006be616

jnm:
This is a consolidation of #197 and #365 with merge conflicts resolved and `flake8` niggles addressed. All credit for the fixes belongs to @vinayinvicible and @tommwatson. 

Since the data-corruption problem is long-standing and severe, please consider merging these fixes and their included unit tests. If there are any issues with the proposed changes that make you hesitate to merge, please bring them up candidly—I'll work to address them ASAP!

I believe that this PR should:

* Close #169
* Fix #160
* Fix #364
* Fix #449

Additionally, #450 is related but cannot be closed by this PR because it also addresses #383, which is a separate issue about file _reading_.

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -74,6 +74,8 @@ class S3BotoStorageFile(File):
         if buffer_size is not None:
             self.buffer_size = buffer_size
         self._write_counter = 0
+        # file position of the latest part file uploaded
+        self._last_part_pos = 0
 
     @property
     def size(self):
@@ -123,9 +125,13 @@ class S3BotoStorageFile(File):
                 reduced_redundancy=self._storage.reduced_redundancy,
                 encrypt_key=self._storage.encryption,
             )
-        if self.buffer_size <= self._buffer_file_size:
+        if self.buffer_size <= self._file_part_size:
             self._flush_write_buffer()
         return super(S3BotoStorageFile, self).write(force_bytes(content), *args, **kwargs)
+
+    @property
+    def _file_part_size(self):
+        return self._buffer_file_size - self._last_part_pos
 
     @property
     def _buffer_file_size(self):
@@ -136,12 +142,15 @@ class S3BotoStorageFile(File):
         return length
 
     def _flush_write_buffer(self):
-        if self._buffer_file_size:
+        if self._file_part_size:
             self._write_counter += 1
-            self.file.seek(0)
+            pos = self.file.tell()
+            self.file.seek(self._last_part_pos)
             headers = self._storage.headers.copy()
             self._multipart.upload_part_from_file(
                 self.file, self._write_counter, headers=headers)
+            self.file.seek(pos)
+            self._last_part_pos = self._buffer_file_size
 
     def close(self):
         if self._is_dirty:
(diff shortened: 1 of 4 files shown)


jnm:
I [replaced](https://github.com/jschneier/django-storages/pull/504/commits/175ad806098fd1d139b6e1d2585fdf9295395e43) `method.assert_not_called()` with `self.assertFalse(method.called)` to restore Python 3.4 compatibility. The wondrous green check boxes have [appeared](https://travis-ci.org/jschneier/django-storages/builds/386008876)!

jschneier:
Taking a look at this one now.

jnm:
@jschneier, thanks for looking at this so quickly. You're going to hate me, but these changes introduce a memory leak because the temporary file is never truncated. I'm working on another PR right now to correct the problem. Sorry!

jschneier:
No hate here! Best to get all the bugs out. Seems like a pain to write a test to cover that.

jnm:
Thanks. Here's what I've got: #506. This time, I tried creating a 10GB ZIP on the fly and streaming it to S3 _before_ making the PR :) It seems to work well.

PR #508, opened 2018-06-07
Always check for security token

jlev:
fix for #282

PR #514, opened 2018-06-29
Fix AWS  security/session token not being set

alej0varas:
Changed a little the logic for setting `backends.s3boto.S3BotoStorage` and `backends.s3boto3.S3Boto3Storage` attributes `access_key`, `secret_key` and `security_token`. 
Currently if `access_key` or `secret_key` are set `security_token` is *not* set.
Code is tested on Lambda.

Fix https://github.com/jschneier/django-storages/issues/282
Related to https://github.com/jschneier/django-storages/pull/283 and https://github.com/jschneier/django-storages/pull/508

### DIFF
--- a/storages/backends/s3boto.py
+++ b/storages/backends/s3boto.py
@@ -14,7 +14,7 @@ from django.utils.encoding import (
 )
 from django.utils.six import BytesIO
 
-from storages.utils import clean_name, safe_join, setting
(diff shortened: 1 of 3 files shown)


jschneier:
Thanks! Merged in 9a92186f34d8a0a985fe7e5ceb61bc636f090632.

Issue #501, opened 2018-05-22
Overriding the default AWS ACL for specific fields


pieterdd:
I use `S3Boto3StorageBackend` with an implicit `AWS_DEFAULT_ACL` of `public-read`. That is the desired value for 90% of my uploads. However, I want one FileField in one model to upload with an ACL of `bucket-owner-read`.

What would be the best way to achieve this? Is there an option to override the default ACL on field level?

jschneier:
You should be able to pass `storage=s3boto3.S3Boto3Storage(default_acl='bucket-owner-read')` to your model field. Does that not work?

pieterdd:
I ended up going with a similar solution. I wrote a class inheriting from `S3Boto3Storage` that overrides the `default_acl` attribute and passing that as the `storage=` argument.

jschneier:
Sorry but why did you need to override at all?

pieterdd:
To make reuse slightly easier. This way I can apply the same class to other fields.

PR #516, opened 2018-07-04
Deprecate old arguments acl & bucket to S3Boto3Storage

jdufresne:
Deprecated in favor of `default_acl` and `bucket_name`. Encourage users to upgrade so the deprecated versions can eventually be removed.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -236,8 +236,21 @@ class S3Boto3Storage(Storage):
 
         # For backward-compatibility of old differing parameter names
         if acl is not None:
+            warnings.warn(
+                "The acl argument of S3Boto3Storage is deprecated. Use "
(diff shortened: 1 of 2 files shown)


jdufresne:
Thanks!

> how does this work with #524?

I'll need to rebase. I'll leave a message in that PR once it is ready.

jschneier:
Sounds good. Just looked at https://github.com/jschneier/django-storages/issues/501 <https://github.com/jschneier/django-storages/issues/501> maybe “acl” is actually a better name? “default" implies you can override it which isn’t the case right?

jdufresne:
That is fine by me.

Are yo suggesting "acl" should replace both "default_acl" & "bucket_acl" or only "acl"? Did you want me to handle that as part of this ticket, or separately?

PR #517, opened 2018-07-04
Deprecate the S3BotoStorage backend in favor of S3Boto3Storage

jdufresne:
Add a deprecation warning when using `S3BotoStorage` that suggests upgrading to `S3Boto3Storage`. Give the user something visible to see to inform them to change their code and warn them the backend will be removed in a future version.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 django-storages change log
 ==========================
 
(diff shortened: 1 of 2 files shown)


jschneier:
Plus one million.

jschneier:
Closing in favor of #555

PR #518, opened 2018-07-04
Remove long deprecated backend GSBotoStorage

jdufresne:
First deprecated in version 1.6, released on 2017-06-21. Removing the unsupported backend reduces testing and maintenance resources.


sww314:
Looks good. Can you also remove the documentation line: 
"It’s possible to access Google Cloud Storage in S3 compatibility mode using other libraries in django-storages, but this is the only library offering native support."

jdufresne:
Thanks for catching that. Done!

jschneier:
I think this is a good idea. My hope is to move to semantic versioning after 1.7 since many people seem to expect that nowadays. The reason I have not been adhering to it is this library was very out of date when first forked (still is in some ways). This will obviously be a breaking change.

What do you think of that plan @jdufresne @sww314?

jdufresne:
@jschneier That works for me. In my other open PRs, I've suggested a few other areas be deprecated or discouraged. Should we consider all of these for a 2.0-ish release?

jschneier:
I'm very happy to start issuing deprecations now. Since it'll probably take some time to get in the major breaking change I want (adjusting how settings is done) I'm cool with any other minor breaking changes we want to roll into 1.7 for now. Nothing is super firm -- it was just a thought I had (could do the same again in 1.8) -- I just don't want to create unnecessary surprises for users.

Personally when I upgrade packages on projects, even with minor version bumps, I read the release notes and go through the commit messages but for ecosystems like node (or even with pip/pipfile?) there can be automatic changes of some nature.

sww314:
I agree this probably a 2.0 release with a few more of the issues pulled in.

PR #520, opened 2018-07-04
Avoid silently changing settings; raise ImproperlyConfigured instead

jdufresne:
If a setting is considered invalid, raise an exception instead of silently changing it to some other value. Avoids confusion and forces the library user to be explicit.

For example, setting S3 location to a value with a leading slash previously stripped it. Instead, now instruct the user to use a valid value without the leading slash.

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 from django.utils.deconstruct import deconstructible
 from django.utils.encoding import force_bytes, smart_str
 
-from storages.utils import clean_name, safe_join, setting
+from storages.utils import check_location, clean_name, safe_join, setting
 
 try:
     from google.cloud.storage.client import Client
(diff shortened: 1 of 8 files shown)


jdufresne:
That would be great. Unfortunately, I believe a system check would require an entry in `INSTALLED_APPS` which doesn't exist for django-storages.

jschneier:
This is in the README:

Once that is done add storages to your INSTALLED_APPS and set DEFAULT_FILE_STORAGE to the backend of your choice. 

Am I missing something? This is admittedly a piece of Django with which I am not that familiar.

jdufresne:
Interesting. Thanks for pointing that out.

django-storages doesn't actually have a need for using `INSTALLED_APPS`. We can simplify the installation by removing it. I have been successfully using django-storages without `INSTALLED_APPS` on projects for years. Using an app would be required for things like models, migrations, templates, and templatetags. The configuration used, `DEFAULT_FILE_STORAGE`, can be any valid Python path even if it isn't a configured app. With that, I think it would be better to have simpler install instructions. I created PR #547 to do so.

jschneier:
I like system checks generally, maybe it is something to consider. I will merge your pull request for now.

PR #524, opened 2018-07-06
Calculate settings when storage is instantiated; not imported

jdufresne:
Makes the storage classes usable with Django's `override_settings`. For example, projects can now change the S3 location & bucket when running their tests. Affects the following backends:

- azure_storage
- gcloud
- gs
- s3boto
- s3boto3

The remaining storage backends do not need to be altered as they do no calculate settings at import time.

The class `BaseStorage` was created to use a common pattern to initialize settings for various backends. Users of these backends can expect them to work consistently. This class is fully backwards compatible. Settings can now be set:

- In settings.py
- Using override_settings
- As a class variable
- As an argument to __init__

Closes #498

Completes all necessary backends and adds tests.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -9,11 +9,11 @@ from azure.storage.blob import BlobPermissions, ContentSettings
 from azure.storage.blob.blockblobservice import BlockBlobService
 from django.core.exceptions import SuspiciousOperation
 from django.core.files.base import File
-from django.core.files.storage import Storage
 from django.utils import timezone
 from django.utils.deconstruct import deconstructible
 from django.utils.encoding import filepath_to_uri, force_bytes
 
+from storages.base import BaseStorage
 from storages.utils import (
     clean_name, get_available_overwrite_name, safe_join, setting,
 )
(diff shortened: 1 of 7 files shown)


sww314:
@jdufresne This looks good, but it will not merge cleanly after I merged on of your other changesets. Can you update?

jdufresne:
Thanks @sww314 I have rebased on to the latest master.

jdufresne:
Rebased to resolve merge conflicts.

sww314:
@jschneier These changes look good to me. Can you review as well?

jschneier:
Looking now. This will close what I consider by far the biggest issue in the original design of the library.

jdufresne:
I have rebased to resolve all merge conflicts. Ready for additional review. Thanks!

sww314:
@jschneier @jdufresne Now that 1.7 is out. Maybe we can get this merged for version 2?

jdufresne:
I've rebased this to the latest master. Feedback welcome.

yozlet:
@jschneier Bumping this PR - it would be very useful for those of us having to initialise credentials from other sources (e.g. reading Google Cloud creds from an env var rather than a file).

sushifan:
@jschneier, any indication as to when this might make it into an official release?  Would be extremely useful as we often need to switch between multiple sets of storage settings on the fly.  Thanks!

jschneier:
Hi,

Thanks for the interest. I’ve been playing work catch up after coming back from vacation this week. I’m going to dedicate one day to django-storages this weekend and this is one of my top priorities.

yozlet:
@sww314 Are you able to merge this? (Note that Travis last tested this PR several months ago, so it would be good to check they still pass.)

jdufresne:
> Note that Travis last tested this PR several months ago, so it would be good to check they still pass.

Rebased. Travis is still green.

jdufresne:
Hi, is there still interest in this PR? I'll happily rebase if it will be reviewed for inclusion. Just let me know.

sushifan:
> Hi, is there still interest in this PR? I'll happily rebase if it will be reviewed for inclusion. Just let me know.

I've been using the following workaround this entire time, but the original idea of this PR still seems like a great, essential feature...  My workaround is defining multiple storage backends and using them as appropriate:

```
from storages.backends.s3boto3 import S3Boto3Storage


class PrivateS3OverwriteStorage(S3Boto3Storage):
    """Only specify params that are different from those in Django settings."""
    def __init__(self):
        super().__init__(file_overwrite=True)


class PublicS3OverwriteStorage(S3Boto3Storage):
    """Only specify params that are different from those in Django settings."""
    def __init__(self):
        super().__init__(default_acl='public-read', file_overwrite=True, querystring_auth=False)
```

jschneier:
Yes, still interested. This greatly improves testability.

jdufresne:
Great! I have rebased.

jschneier:
Can you please change the method name to `get_default_settings`? Good to go after that. Thanks for your patience as usual.

jdufresne:
:+1: done.

terencehonles:
@jdufresne @jschneier I was just updating an unrelated PR, but I was wondering should `BaseStorage` provide a default for `get_default_settings` which returns an empty `{}`? Right now this would break for any custom storages, but if the base storage was customized to not raise `ImproperlyConfigured` but a deprecated warning it would be a soft fail until the next major release.

Since it will only take me a couple moments to write and I'm already updating an existing PR I'll just create a PR for this, and we can go from there.

jdufresne:
> Right now this would break for any custom storages,

Custom storages aren't yet inheriting from `BaseStorage` as it is brand new.

But I'd also be fine with an implementation that return `{}`.

terencehonles:
@jdufresne I was just going to update my comment with that, sorry, disregard my previous message.

I'm not sure an empty `{}` is warranted because it's likely that a storage is providing _some_ options, and it should use the `get_default_settings` for that? Or do you suspect people would create a new storage based off the `BaseStorage` that would have no settings? I guess for maybe a private source/single use storage that might be the case.

PR #531, opened 2018-07-13
Azure > 1.0

Peter-Slump:
Hi Josh,

I've created a storage back-end for the Azure library > 1.0 
In this implementation a new back-end is created (azure1). However it should also be possible to merge the changes into the original Azure back-end. In that case we should create some if-checks to find out which implementation to use. 

What do you prefer, do you want me to merge the support for > 1.0 library into the current Azure back-end or do you want to stay with this second Azure back-end?

Looking forward to hear from you.


sww314:
Closed in favor of #565

Issue #381, opened 2017-08-14
Change AWS_DEFAULT_ACL to 'private'


jonashaag:
1. For obvious security reasons
2. Because it doesn't work out of the box with the default S3 tutorial (which doesn't grant publish permissions), with no good error message ("Access Denied", where a good message would be "Access to make file `public-read` denied")

jschneier:
I agree, will have to go in a 2.0 release. I've been lax about breaking changes because they are usually smaller and not in widely used backends. This would not be that in any way.

robatwave:
By default it shouldn't set ACL at all, and defer to the bucket's ACL? Or as a non-breaking change, there should be an option to not explicitly set an ACL on the object.

robatwave:
Just to follow up: looks like we can work around this by explicitly setting `AWS_DEFAULT_ACL` to `None` in settings.

jschneier:
We will now warn about the insecure defaults with a recommendation to update.

PR #535, opened 2018-07-18
Warn about upcoming change to AWS_DEFAULT_ACL; allow None in AWS_DEFAULT_ACL

jcushman:
**Background**

As discussed in #381, the default setting of "public-read" for `AWS_DEFAULT_ACL` is insecure -- if a user sets up a bucket that is private by default, they should have to explicitly opt in for django-storages to override that choice and upload globally readable files. By default uploads should use the bucket's settings.

@jschneier suggested in that bug that the change should be made in django-storages 2.0, since it would be a breaking change for many users.

@robatwave commented that one can currently opt into the proposed behavior by setting `AWS_DEFAULT_ACL=None` explicitly. This does not work for me; with that setting I get `'NoneType' object has no attribute 'parts'` in `S3Boto3StorageFile.close`.

@ticosax sent a pull request last year, #429, that implements the breaking change but currently has some tests failing.

It would be great to take action on this without waiting for the decision to ship django-storages 2.0. It's likely that lots of users are currently using django-storages in a way that unintentionally leaks data -- it's not something where it's harmless to wait.

**This PR**

This is a change that can be made immediately without waiting for 2.0. It gets `AWS_DEFAULT_ACL=None` working, and adds a warning that the default behavior will be changing in 2.0.

This PR is based on the work @ticosax did and includes 3 commits:

* (1) Incorporate the changes from @ticosax in #429. 
* (2) Fix two issues with that PR, adding a second check for `None` and tweaking a test. This gets `AWS_DEFAULT_ACL=None` working for me.
* (3) Back out the breaking changes from #429 for now, and instead print a warning if the user has no explicit `AWS_DEFAULT_ACL` setting. The warning includes instructions for how (and why) a user can either explicitly retain the existing behavior if it's actually what they want, with `AWS_DEFAULT_ACL="public-read"`, or prepare for 2.0 by opting into the new behavior with `AWS_DEFAULT_ACL=None`.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -194,8 +194,8 @@ class S3Boto3Storage(Storage):
     object_parameters = setting('AWS_S3_OBJECT_PARAMETERS', {})
     bucket_name = setting('AWS_STORAGE_BUCKET_NAME')
     auto_create_bucket = setting('AWS_AUTO_CREATE_BUCKET', False)
-    default_acl = setting('AWS_DEFAULT_ACL', 'public-read')
-    bucket_acl = setting('AWS_BUCKET_ACL', default_acl)
+    default_acl = setting('AWS_DEFAULT_ACL', None)
+    bucket_acl = setting('AWS_BUCKET_ACL', None)
     querystring_auth = setting('AWS_QUERYSTRING_AUTH', True)
     querystring_expire = setting('AWS_QUERYSTRING_EXPIRE', 3600)
     signature_version = setting('AWS_S3_SIGNATURE_VERSION')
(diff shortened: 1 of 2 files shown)


ticosax:
well done

jcushman:
@sww314, just a nudge -- is there anything else I can do to assist with getting this merged? Thanks!

sww314:
@jcushman Looks ok to me, but I am not actively using the AWS storages, so I would like someone that uses the boto/boto3 to chime in.

jschneier:
Thanks so much. I've come to the conclusion that my decision earlier on this was wrong and it should have been addressed immediately. Going to merge and cut 2.0 with this fixed on Friday. For some of the other changes I want to make we can wait to 3.0 which can always come quicker.

PR #538, opened 2018-07-25
Ensure content_type is str in S3Boto3Storage

chris48s:
If `content_type` in `S3Boto3Storage._save()` is expressed as `bytes` instead of `str`, this will cause `botocore` to throw a validation error e.g: `Invalid type for parameter ContentType, value: b'text/javascript', type: <class 'bytes'>, valid types: <class 'str'>: ParamValidationError`.

This patch fixes this issue by ensuring `content_type` is a `str` before handing the value off to `boto3`.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -431,6 +431,8 @@ class S3Boto3Storage(Storage):
         _type, encoding = mimetypes.guess_type(name)
         content_type = getattr(content, 'content_type', None)
         content_type = content_type or _type or self.default_content_type
+        if type(content_type) == bytes:
+            content_type = content_type.decode('utf8')
 
         # setting the content_type in the key object is not enough.
         parameters.update({'ContentType': content_type})
--- a/tests/test_s3boto3.py
+++ b/tests/test_s3boto3.py
@@ -109,6 +109,25 @@ class S3Boto3StorageTests(S3Boto3TestCase):
             }
         )
 
+    def test_bytes_content_type(self):
+        """
+        Test saving a file with a Bytes content type.
+        """
+        name = 'test_data.json'
+        content = ContentFile('data')
+        content.content_type = b'application/json'
+        self.storage.save(name, content)
+        self.storage.bucket.Object.assert_called_once_with(name)
+
+        obj = self.storage.bucket.Object.return_value
+        obj.upload_fileobj.assert_called_with(
+            content.file,
+            ExtraArgs={
+                'ContentType': u'application/json',
+                'ACL': self.storage.default_acl,
+            }
+        )
+
     def test_storage_save_gzipped(self):
         """
         Test saving a gzipped file


jschneier:
Hi and thanks for opening this PR. I see you have referenced this from another issue, how did this come about? It almost seems more like a programming error.

chris48s:
Hi @jschneier

> how did this come about?

Its been a few weeks since I looked at this, but I hit this problem trying to run `collectstatic` while deploying an application to AWS Lambda with the static files hosted on S3. `collectstatic` is being run via `zappa manage`. I think the calls to `django-storages` are being made from `django-pipeline`. We declared a storage class folowing the example in https://django-pipeline.readthedocs.io/en/latest/storages.html#using-with-other-storages :


```python
from django.contrib.staticfiles.storage import ManifestFilesMixin
from pipeline.storage import PipelineMixin
from storages.backends.s3boto import S3BotoStorage


class StaticStorage(PipelineMixin, ManifestFilesMixin, S3Boto3Storage):
    location = 'static'
```

and then in the settings

```python
STATICFILES_STORAGE = 'myapp.myfile.StaticStorage'
```

I've included a stacktrace, in case that helps:

```
Invalid type for parameter ContentType, value: b'text/javascript', type: <class 'bytes'>, valid types: <class 'str'>: ParamValidationError
  File "/var/task/handler.py", line 567, in lambda_handler
  return LambdaHandler.lambda_handler(event, context)
  File "/var/task/handler.py", line 240, in lambda_handler
  return handler.handler(event, context)
  File "/var/task/handler.py", line 403, in handler
  management.call_command(*event['manage'].split(' '))
  File "/var/task/django/core/management/__init__.py", line 130, in call_command
  return command.execute(*args, **defaults)
  File "/var/task/django/core/management/base.py", line 345, in execute
  output = self.handle(*args, **options)
  File "/var/task/django/contrib/staticfiles/management/commands/collectstatic.py", line 193, in handle
  collected = self.collect()
  File "/var/task/django/contrib/staticfiles/management/commands/collectstatic.py", line 139, in collect
  for original_path, processed_path, processed in processor:
  File "/var/task/pipeline/storage.py", line 33, in post_process
  packager.pack_javascripts(package)
  File "/var/task/pipeline/packager.py", line 112, in pack_javascripts
  return self.pack(package, self.compressor.compress_js, js_compressed, templates=package.templates, **kwargs)
  File "/var/task/pipeline/packager.py", line 107, in pack
  self.save_file(output_filename, content)
  File "/var/task/pipeline/packager.py", line 118, in save_file
  return self.storage.save(path, ContentFile(smart_bytes(content)))
  File "/var/task/django/core/files/storage.py", line 54, in save
  return self._save(name, content)
  File "/var/task/storages/backends/s3boto3.py", line 452, in _save
  self._save_content(obj, content, parameters=parameters)
  File "/var/task/storages/backends/s3boto3.py", line 467, in _save_content
  obj.upload_fileobj(content, ExtraArgs=put_parameters)
  File "/var/runtime/boto3/s3/inject.py", line 621, in object_upload_fileobj
  ExtraArgs=ExtraArgs, Callback=Callback, Config=Config)
  File "/var/runtime/boto3/s3/inject.py", line 539, in upload_fileobj
  return future.result()
  File "/var/task/s3transfer/futures.py", line 73, in result
  return self._coordinator.result()
  File "/var/task/s3transfer/futures.py", line 233, in result
  raise self._exception
  File "/var/task/s3transfer/tasks.py", line 126, in __call__
  return self._execute_main(kwargs)
  File "/var/task/s3transfer/tasks.py", line 150, in _execute_main
  return_value = self._main(**kwargs)
  File "/var/task/s3transfer/upload.py", line 692, in _main
  client.put_object(Bucket=bucket, Key=key, Body=body, **extra_args)
  File "/var/runtime/botocore/client.py", line 314, in _api_call
  return self._make_api_call(operation_name, kwargs)
  File "/var/runtime/botocore/client.py", line 586, in _make_api_call
  api_params, operation_model, context=request_context)
  File "/var/runtime/botocore/client.py", line 621, in _convert_to_request_dict
  api_params, operation_model)
  File "/var/runtime/botocore/validate.py", line 291, in serialize_to_request
  raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter ContentType, value: b'text/javascript', type: <class 'bytes'>, valid types: <class 'str'>
```

> It almost seems more like a programming error.

Since `.save()` may be passed an arbitary `ContentFile` object, it seems like ensuring that this value is of a valid type for the relevant storage API before passing it on to that API is desirable behaviour here.


Thanks

jdufresne:
I agree with @jschneier

This seems like an error elsewhere. IMO, `bytes` shouldn't have been passed in the first place. This looks like a workaround for some other bug.

chris48s:
> This seems like an error elsewhere. IMO

Are you able to explain a bit more about that @jdufresne ? Is it because of the way I have written the test case that simulates the condition, or is there something else?

Thanks

jdufresne:
Yes. Why would code ever do:

```py
        content = ContentFile('data')
        content.content_type = b'application/json'
```

Instead of:

```py
        content = ContentFile('data')
        content.content_type = 'application/json'
```

?

In other words, why is code ever assigning a bytes to `.content_type` instead of a string? To me, whatever is doing that seems like the source of the bug. The two types shouldn't be used interchangeably. Do you have additional details about what is causing this?

chris48s:
OK. I chose to write the test in that way because it was the most convenient way to get a `bytes` value into that function. I'm not somewhere I can conveniently debug this right now (I had the stacktrace saved), but I'll try and get back into it next week and update with more detail on how this condition is happening.

chris48s:
Hi. Thanks for checking in. I investigated a couple of things, but the environment I'm seeing this in has too many moving parts. I need some more time to chase this through the dependency tree and work out a minimal reproducible example. I will circle back to this at some point as I do want to find the solution to this but if you want to close this PR in the meantime and get it off your list of notifications feel free. I can re-open a PR or issue later if needed.

jschneier:
Sounds good 👍

PR #553, opened 2018-08-12
Deprecate AWS_PRELOAD_METADATA

jschneier:


blitzd:
Hey - we were using this option as we found without it things like filebrowser-safe were just too slow. Any known alternatives for this scenario?

jschneier:
Honestly not opposed to keeping this option, was curious what people were using it for since the boto3 version was broken for a while before it was updated I kind of assumed it wasn’t used. It’s also undocumented but if it is being used by the community we can keep it and document it. The biggest downside is making the codebase a bit more complicated, a larger testing suite would help with that issue.

blitzd:
Ah - I was actually hoping you had an alternative!

While this does speed up things like filebrowser where there are otherwise a lot of roundtrips, it also leads to some issues in a load balanced environment where file uploads done on one node can show up as directories or not at all on other nodes. They do still work perfectly fine as file objects if you use the link, but often times when filebrowser is used for CMS like activities to pick a graphic file you aren't able to 'pick' anything it thinks is a directory. I think this stems from the _entries object only ever being created once per instance, so I'm going to look into maybe adding a way to occasionally resync it. If it ends up working out for us I'll look at adding tests as well.

jschneier:
Concurrency issues is a very good reason to move ahead with the deprecation. I don’t have another suggestion off-hand but it certainly feels like something where you want to have a cache in the middle somehow.

jschneier:
Okay searching on GitHub shows there is some use so I am going to revert those commits and will add docs + warning about these concurrency issues.

PR #554, opened 2018-08-12
Fixed #513 -- Properly truncate available names when file_overwrite = True

### LINKED ISSUE #513: Respect max_length when AWS_S3_FILE_OVERWRITE=True
`get_available_name` method of `S3Boto3Storage`  doesn't respect `max_length`.

The Django Storage's `get_available_name` method truncates the file name if it's length is more than `max_length`.

When `AWS_S3_FILE_OVERWRITE=True` and len(file_name) > max_length, this results in Database error - `DataError: (1406, "Data too long for column 'file_field' at row 1")`.

```
    def get_available_name(self, name, max_length=None):
        """Overwrite existing file with the same name."""
        if self.file_overwrite:
            name = self._clean_name(name)
           >>  ** CHANGE REQUIRED HERE: max_length VALIDATION  **
            return name
        return super(S3Boto3Storage, self).get_available_name(name, max_length)
```

Validation for `max_length` should be added to the overridden method, when file_overwrite=True.

jschneier:


jschneier:
Need to fix for other backends but what do we think about this approach?

jschneier:
@nitely is `AzureStorage` affected by this? My reading is yes but there is some special `_AZURE_MAX_LEN` indirection in there and just want to confirm.

nitely:
@jschneier I'm afraid it is. `_AZURE_MAX_LEN` just makes sure the name doesn't exceed azure's name length limit. Without it, it would raise later when uploading the file (hopefully, but I didn't try it).

PR #556, opened 2018-08-12
Remove unnecessary install step from Travis CI

jdufresne:
Upgrading pip, setuptools, and wheel is unnecessary. It only adds time and network traffic to the CI build.


jschneier:
Travis was a nightmare last night.

PR #557, opened 2018-08-12
Remove suffix "storage" from module names that have it

jdufresne:
Will result in a more consistent naming style for Python modules and backends. The old names are still available as aliases, but produce a warning that they will be removed in django-storages 2.0.


jdufresne:
I closed this because it occurred to me that the name of the Azure product is ["Azure Storage". So, IIUC, the name is `azure_storage.py` due to the name of the product, not because it is a Django storage backend. Is that right?

I'll reopen for now.

PR #561, opened 2018-08-17
Make sure that we read the file from the beginning in Google Cloud Storage backend

m1kola:
Makes the same effect as with s3 storage:

https://github.com/jschneier/django-storages/blob/7757ff0ee62e52d2c67fcfe90858a3c646ad0de1/storages/backends/s3boto3.py#L512

Without rewinding the file to the beginning, it's possible that we will try to upload 0 bytes in case if some code reads the file before Django calls the save method on backend.


m1kola:
I updated asserts in tests. Should be ready for review now.

jschneier:
This was fixed in #481. Nice that there is `rewind=True` so I have updated the code to use that in #581.

PR #562, opened 2018-08-20
Make sure that we read the file from the beginning in Azure Storage backend

m1kola:
The same as #561, but for Azure Storage backend

I'm not familiar with Azure Storage and I'm not using it. Also there are no tests for this backend, so please, review this carefully :)

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -88,6 +88,7 @@ class AzureStorage(Storage):
         else:
             content_type = mimetypes.guess_type(name)[0]
 
+        content.seek(0)
         if hasattr(content, 'chunks'):
             content_data = b''.join(chunk for chunk in content.chunks())
         else:


storages/backends/azure_storage.py
```diff
@@ -88,6 +88,7 @@ def _save(self, name, content):
         else:
             content_type = mimetypes.guess_type(name)[0]
 
+        content.seek(0)
         if hasattr(content, 'chunks'):
             content_data = b''.join(chunk for chunk in content.chunks())
```
m1kola: As far as I understand `content` is a file-like object, but I have no idea about this case. Is it something Azure-specific? Is it ok to call `seek` on `content ` before this?

jschneier:
Azure storage was massively rewritten and merged in since then. However this does appear to be a real issue and I have opened #582 to resolve it. Thanks for the PRs.

PR #318, opened 2017-05-16
Upgrade azure

guydou:
This pull request:
1. using the recent azure-storage lib
2. writing tests to the azure backend
3. Implementing streaming write for large files

PR #565, opened 2018-08-22
Azure backend

nitely:
This PR supersedes #318. Implements:

* Upgraded azure dependencies (`azure-storage` is no longer the recommended lib, this depends on `azure` and `azure-storage-blob` instead)
* Support for streams to avoid hitting disk (although Django may hit the disk if regular form file is used)
* Proper file name checking
* `listdir` method to list files by virtual path
* Several settings that were missing in previous PR: timeout, upload_max_conn, overwrite_files, location and default_content_type. See docs for the settings explanation.
* Integration tests using [azurite](https://github.com/Azure/Azurite) (oficial emulated Azure).
* Docs explaining how to use public and private storage.


gilgamezh:
:tada:

jedie:
Will it then be v1.7? Please write here when it comes out!

jschneier:
@jedie version 1.7 is now live.

PR #569, opened 2018-08-26
S3Boto3Storage: use UNSIGNED signature version for public urls

### LINKED ISSUE #254: AWS_QUERYSTRING_AUTH = False still requires credentials to generate static URLs
When using a public S3 bucket for static files and `AWS_QUERYSTRING_AUTH = False`, it would be advantageous to not require AWS access keys for django-storages to generate static urls.

As it stands, this setting only strips the URL after the credentials are created and added. If the credentials aren't present, we get a `NoCredentialsError` even though this particular function shouldn't need credentials to work if static files are public.

I'm not 100% on all the places Django calls `url()` or if I'm missing something where the credentials would be required anyway (I understand they're important for operations like uploads); does this make sense to implement?

[storages/backends/s3boto3.py line 576:](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L576)
```
        url = self.bucket.meta.client.generate_presigned_url('get_object', Params=params,
                                                             ExpiresIn=expire)
        if self.querystring_auth:
            return url
        return self._strip_signing_parameters(url)
```

jschneier:
Fixes #254

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -67,11 +67,6 @@ Available are numerous settings. It should be especially noted the following:
             'CacheControl': 'max-age=86400',
         }
 
-``AWS_QUERYSTRING_AUTH`` (optional; default is ``True``)
-    Setting ``AWS_QUERYSTRING_AUTH`` to ``False`` to remove query parameter
-    authentication from generated URLs. This can be useful if your S3 buckets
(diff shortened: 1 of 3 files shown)


jschneier:
We need integration tests. This breaks file upload since there is only one client config used. Will have to look into it more.

Issue #425, opened 2017-11-10
Google Cloud Storage urls getting expired


gbrennon:
How to prevent ExpiredToken for files url?

My files are public and I can visit them only with the name of the file, but django-storages continues to generate urls that expire...

How can I do this? I want to be able to send uploaded files url to people

helloagain-dev:
I have the same issue.
The S3 Storage has an option to disable this.

It would be great to have an option for the url call to get a signed URL or the public URL as some of my files are public and some are private.

ThunderTrent:
I have the same issues -- and would desire that option as well.

jschneier:
This has been fixed in #570.

PR #570, opened 2018-08-27
Gcloud url fixes to handle public and private files 

### LINKED ISSUE #491: Google Cloud backend not a very good replacement for Django's Media or Static handling
The Google cloud backend isn't a very good candidate for Django's Media or Static files storage backend without some additional configuration and studying the sources. Ideally, the docs would talk a little bit about issues such as #463 #413 

Here's my solution:

```python
"""
GoogleCloudStorage extensions suitable for handing Django's
Static and Media files.

Requires following settings:
MEDIA_URL, GS_MEDIA_BUCKET_NAME
STATIC_URL, GS_STATIC_BUCKET_NAME

In addition to
https://django-storages.readthedocs.io/en/latest/backends/gcloud.html
"""
from django.conf import settings
from storages.backends.gcloud import GoogleCloudStorage
from storages.utils import setting
from urllib.parse import urljoin


class GoogleCloudMediaStorage(GoogleCloudStorage):
    """GoogleCloudStorage suitable for Django's Media files."""

    def __init__(self, *args, **kwargs):
        if not settings.MEDIA_URL:
            raise Exception('MEDIA_URL has not been configured')
        kwargs['bucket_name'] = setting('GS_MEDIA_BUCKET_NAME', strict=True)
        super(GoogleCloudMediaStorage, self).__init__(*args, **kwargs)

    def url(self, name):
        """.url that doesn't call Google."""
        return urljoin(settings.MEDIA_URL, name)


class GoogleCloudStaticStorage(GoogleCloudStorage):
    """GoogleCloudStorage suitable for Django's Static files"""

    def __init__(self, *args, **kwargs):
        if not settings.STATIC_URL:
            raise Exception('STATIC_URL has not been configured')
        kwargs['bucket_name'] = setting('GS_STATIC_BUCKET_NAME', strict=True)
        super(GoogleCloudStaticStorage, self).__init__(*args, **kwargs)

    def url(self, name):
        """.url that doesn't call Google."""
        return urljoin(settings.STATIC_URL, name)
```

sww314:
For public files use blob.public_url. This does not require a network trip to Google, so much faster. 
For non-public files, it will return a signed url. The default authentication is set to 1 day but is configurable with GS_EXPIRES_IN. 

It relies on the GS_DEFAULT_ACL to tell if the files will be public. Actually, checking the ACL with Google would be slow. It does not use the GS_CREATE_BUCKET_ACL on purpose because public buckets are a bad idea and lead to a lot of security leaks. 

Docs + tests are updated.

Does not currently support a custom url (did not before either), can be added as an additional feature.

Fixes #491 and #425

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -138,7 +138,7 @@ Sets Cache-Control HTTP header for the file, more about HTTP caching can be foun
 Subdirectory in which the files will be stored.
 Defaults to the root of the bucket.
 
-``GS_EXPIRES_ON`` (optional: default is ``''``)
+``GS_EXPIRES_ON`` (optional: default is ``timedelta(seconds=86400)``)
 
 The time that a generated URL is valid before expiration. The default is 1 day.
 Public files will return a url that does not expire.
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -143,7 +143,7 @@ Sets Cache-Control HTTP header for the file, more about HTTP caching can be foun
 Subdirectory in which the files will be stored.
 Defaults to the root of the bucket.
 
-``GS_EXPIRES_ON`` (optional: default is ``timedelta(seconds=86400)``)
+``GS_EXPIRES_IN`` (optional: default is ``timedelta(seconds=86400)``)
 
 The time that a generated URL is valid before expiration. The default is 1 day.
 Public files will return a url that does not expire. Files will be signed by
@@ -152,7 +152,7 @@ the credentials provided to django-storages (See GS_CREDENTIALS).
 Note: Default Google Compute Engine (GCE) Service accounts are
 `unable to sign urls <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
 
-The ``GS_EXPIRES_ON`` value is handled by the underlying `Google library  <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
+The ``GS_EXPIRES_IN`` value is handled by the underlying `Google library  <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
 It supports `timedelta`, `datetime`, or `integer` seconds since epoch time.
 
 
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -143,7 +143,7 @@ Sets Cache-Control HTTP header for the file, more about HTTP caching can be foun
 Subdirectory in which the files will be stored.
 Defaults to the root of the bucket.
 
-``GS_EXPIRES_IN`` (optional: default is ``timedelta(seconds=86400)``)
+``GS_EXPIRATION`` (optional: default is ``timedelta(seconds=86400)``)
 
 The time that a generated URL is valid before expiration. The default is 1 day.
 Public files will return a url that does not expire. Files will be signed by
@@ -152,7 +152,7 @@ the credentials provided to django-storages (See GS_CREDENTIALS).
 Note: Default Google Compute Engine (GCE) Service accounts are
 `unable to sign urls <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
 
-The ``GS_EXPIRES_IN`` value is handled by the underlying `Google library  <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
+The ``GS_EXPIRATION`` value is handled by the underlying `Google library  <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
 It supports `timedelta`, `datetime`, or `integer` seconds since epoch time.
 
 
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -113,6 +113,8 @@ translated.)
 For most cases, the blob will need to be set to the ``publicRead`` ACL in order for the file to viewed.
 If GS_DEFAULT_ACL is not set, the blob will have the default permissions set by the bucket.
 
+``publicRead`` files will return a public - non-expiring url.
+
 
 ``GS_FILE_CHARSET`` (optional)
 
@@ -135,6 +137,12 @@ Sets Cache-Control HTTP header for the file, more about HTTP caching can be foun
 
 Subdirectory in which the files will be stored.
 Defaults to the root of the bucket.
+
+``GS_EXPIRES_ON`` (optional: default is ``''``)
+
+The time that a generated URL is valid before expiration. The default is 1 day.
+Public files will return a url that does not expire.
+
 
 Usage
 -----
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -147,13 +147,13 @@ Defaults to the root of the bucket.
 
 The time that a generated URL is valid before expiration. The default is 1 day.
 Public files will return a url that does not expire. Files will be signed by
-the default credentials provided to the client.
+the credentials provided to django-storages (See GS_CREDENTIALS).
 
 Note: Default Google Compute Engine (GCE) Service accounts are
 `unable to sign urls <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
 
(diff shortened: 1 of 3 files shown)


jschneier:
How about checking the expiration time rather than the acl? I feel like there are a bunch of reasons that is better.

sww314:
Using `projectPrivate` with `public_url` does not work - well not the way they would want it to. It returns the public_url and it will work for the developer (because they have access to the bucket), but it will not work for anyone else.

Explcitly settting the files to public is the most common case for profile images, static files etc. This flow makes it clear you are doing that and handles it quickly (unlike previous version).

I added a default GS_EXPIRES_ON, so if people use the GCS default of `projectPrivate` it will work. In this case, when users upgrade:
1. projectPrivate files will now actually work, but they did not before.
2. publicBuckets need to add the GS_DEFAULT_ACL or they will get signed urls (not ideal). Public buckets are a security risk should not be encouraged. 

The current version is broken so backwards compatability is not critical. I can write better documentation as a different pull request. 

Finally, I can switch the default signed vs public acl check, but just wanted to explain my thinking.

storages/backends/gcloud.py
```diff
@@ -249,10 +252,17 @@ def get_created_time(self, name):
...
+        """
+        Return public url or a signed url for the Blob.
+        This DOES NOT check for existance of Blob - that makes codes too slow
+        for many use cases.
+        """
         name = self._normalize_name(clean_name(name))
-        blob = self._get_blob(self._encode_name(name))
-        return blob.public_url
+        blob = self.bucket.blob(self._encode_name(name))
+        if (self.default_acl == 'publicRead'):
```
jschneier: Are we covering all possibilities in the [possible ACL list](https://cloud.google.com/storage/docs/access-control/lists) with this logic?

If so, let's remove the parens. Otherwise, what changes are necessary?

docs/backends/gcloud.rst
```diff
@@ -136,6 +138,13 @@ Sets Cache-Control HTTP header for the file, more about HTTP caching can be foun
 Subdirectory in which the files will be stored.
 Defaults to the root of the bucket.
 
+``GS_EXPIRES_ON`` (optional: default is ``timedelta(seconds=86400)``)
+
+The time that a generated URL is valid before expiration. The default is 1 day.
+Public files will return a url that does not expire. Files will be signed by
+the default credentials provided to the client.
```
jschneier: From my reading of the source this **MUST** be a `timedelta`, can we note that?

Also, the name, how about `GS_EXPIRES_IN`? The `ON` language is maybe a bit awkward?

storages/backends/gcloud.py
```diff
@@ -249,10 +252,17 @@ def get_created_time(self, name):
...
+            return blob.public_url
+
+        return blob.generate_signed_url(expiration=self.expires_on)
```
jschneier: `expiration` isn't a `kwarg` and is in fact the only mandatory so let's remove the `expiration=`.

storages/backends/gcloud.py
```diff
@@ -87,6 +88,8 @@ class GoogleCloudStorage(Storage):
     auto_create_acl = setting('GS_AUTO_CREATE_ACL', 'projectPrivate')
     default_acl = setting('GS_DEFAULT_ACL')
 
+    expires_on = setting('GS_EXPIRES_ON', timedelta(seconds=86400))
```
jschneier: We should reflect the variable name if we change the setting name.

storages/backends/gcloud.py
sww314: Technically there is publicReadWrite that allows you to create a bucket that anyone can write to - this seems dangerous, but I can add the check.

jschneier:
I think that ACL is lunacy so I won’t even mention it. What about authenticated read?

docs/backends/gcloud.rst
sww314: The underlying library supports a datetime, timedelta, or an integer number of seconds. I will change it to ON vs IN.

storages/backends/gcloud.py
sww314: authenticatedRead only works for Google accounts that have been added to the Blob. Since we have no way on knowing if the user is one of those authenticated people - the safe thing is to send the signed url. 
Basically, if you are going to really use Google ACL and user lists, you will have to use the Google API more directly.

jschneier:
Okay. That’s fine. Just trying to think through if there is a reason to have public-read and to still want to have expiration time. If not then this approach is good. Also nice because users won't have to set 2 different settings to get unsigned public urls.

docs/backends/gcloud.rst
jschneier: This is the doc of that param:

```   
:type expiration: :class:`int`, :class:`long`, :class:`datetime.datetime`,
:class:`datetime.timedelta`
:param expiration: When the signed URL should expire.
```
which gets passed to `get_expiration_seconds` which is:
```python
def get_expiration_seconds(expiration):
    """Convert 'expiration' to a number of seconds in the future.

    :type expiration: int, long, datetime.datetime, datetime.timedelta
    :param expiration: When the signed URL should expire.

    :raises TypeError: When expiration is not an integer.

    :rtype: int
    :returns: a timestamp as an absolute number of seconds.
    """
    # If it's a timedelta, add it to `now` in UTC.
    if isinstance(expiration, datetime.timedelta):
        now = NOW().replace(tzinfo=_helpers.UTC)
        expiration = now + expiration

    # If it's a datetime, convert to a timestamp.
    if isinstance(expiration, datetime.datetime):
        micros = _helpers._microseconds_from_datetime(expiration)
        expiration = micros // 10**6

    if not isinstance(expiration, six.integer_types):
        raise TypeError('Expected an integer timestamp, datetime, or '
                        'timedelta. Got %s' % type(expiration))
    return expiration
```
which I think interprets the integer as an epoch time.

sww314: Good catch!

docs/backends/gcloud.rst
```diff
@@ -136,6 +143,19 @@ Sets Cache-Control HTTP header for the file, more about HTTP caching can be foun
 Subdirectory in which the files will be stored.
 Defaults to the root of the bucket.
 
+``GS_EXPIRES_ON`` (optional: default is ``timedelta(seconds=86400)``)
```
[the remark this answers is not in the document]
sww314: Docs are updated. The JSON service account file is handled already by GS_CREDENTIALS. I do not see any use case for 2 different service accounts - inside the same instance of django-storages.

docs/backends/gcloud.rst
sww314: Updated in docs.

docs/backends/gcloud.rst
```diff
@@ -136,6 +143,19 @@ Sets Cache-Control HTTP header for the file, more about HTTP caching can be foun
...
+
+The time that a generated URL is valid before expiration. The default is 1 day.
+Public files will return a url that does not expire. Files will be signed by
+the credentials provided to django-storages (See GS_CREDENTIALS).
+
+Note: Default Google Compute Engine (GCE) Service accounts are
+`unable to sign urls <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
+
+The ``GS_EXPIRES_IN`` value is handled by the underlying `Google library  <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
+It supports `timedelta`, `datetime`, or `integer` seconds since epoch time.
```
jschneier: Kind of:

`timedelta` is added to `.utcnow()` while `date` and `int` are literal times. It would be pretty weird to be using `date` and `int` in this library correctly.

storages/backends/gcloud.py
```diff
@@ -249,10 +252,17 @@ def get_created_time(self, name):
...
+        """
+        Return public url or a signed url for the Blob.
+        This DOES NOT check for existance of Blob - that makes codes too slow
+        for many use cases.
+        """
         name = self._normalize_name(clean_name(name))
-        blob = self._get_blob(self._encode_name(name))
-        return blob.public_url
+        blob = self.bucket.blob(self._encode_name(name))
+        if self.default_acl == 'publicRead':
```
jschneier: My nitpick on style would be to either:

a) put a blank line above the `if` and eliminate the one below it
b) use a ternary with no newline (a bit ugly i suppose with the function call)

docs/backends/gcloud.rst
sww314: Shrug... maybe you want the URL to expire on Christmas in 2020. I expect most people will use timedelta.

storages/backends/gcloud.py
sww314: Fixed.

docs/backends/gcloud.rst
jschneier: Yeah but this is where `expires_on` vs `expires_in` comes into play. If I put `timedelta(hours=4)` then that is an `expires_in` scenario. If I put `2020-12-25` then that is `expires_on`. It seems very confusing.

sww314: That is why I originally had it at expires_on because it expires_on (now + timedelta) or an absolute time.... 

I would also argue this is a very minor point because without this change the current version is broken. 
My guess it 99% of people set a timedelta, but I do not think we should block the underlying Google functionality by disallowing the other options.

PR #575, opened 2018-08-31
Google: Only verify bucket if auto_create is True

jschneier:


storages/backends/gcloud.py
```diff
@@ -131,17 +130,15 @@ def _get_or_create_bucket(self, name):
         """
         Retrieves a bucket if it exists, otherwise creates it.
         """
+        if not self.auto_create_bucket:
+            return Bucket(client=self.client, name=name)
```
stanislavb: Alternative syntax:
```suggestion
            return self.client.bucket(name)
```

stanislavb:
I was going to suggest such a change to get rid of storage.buckets.get permission requirement, but found that here it is! It also supersedes #412

jschneier:
Merged #718 which takes this further.

PR #591, opened 2018-09-06
Avoid AWS_DEFAULT_ACL warning when overridden as a class variable

jdufresne:
Prevents warning spam to django-storages users that have overridden the insecure setting as a class variable.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -277,7 +277,7 @@ class S3Boto3Storage(Storage):
             self.config = Config(**kwargs)
 
         # warn about upcoming change in default AWS_DEFAULT_ACL setting
-        if not hasattr(django_settings, 'AWS_DEFAULT_ACL'):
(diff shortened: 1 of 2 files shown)


jschneier:
Not perfect but better than nothing I suppose.

jdufresne:
@jschneier What do you think about a releasing 1.7.2 to include this fix? I run my project's tests with warnings enabled so avoiding this large amount of noise would be nice.

davepeck:
I ran across this issue with a client who was explicitly overriding `default_acl` in a class variable anyway; I have to say that the warning was very confusing! To make sense of it, I eventually spelunked into `s3boto3.py` and did a `git blame`, which led me here.

@jschneier Is there any reason we shouldn't issue a 1.7.2 release that contains this fix?

jschneier:
Sorry all I’ve started a new job recently and things have fallen behind.
Going to release a new version this weekend.

jschneier:
@davepeck this has now been released, sorry for the delay.

davepeck:
@jschneier Wow — thanks!

PR #598, opened 2018-09-17
Add tox directory to isort skip list

jdufresne:
Like flake8, when running `isort --apply`, shouldn't affect the `.tox` directory.


jdufresne:
isort now does this by default.

PR #618, opened 2018-10-18
Fixing wrong documentation

arpan29:
Copied from this location and spent half an hour to just figure out why the module not found error was there. Realized that the documentation missed out on the correct module name.


jschneier:
Thanks.

PR #631, opened 2018-12-07
Use tox configuration option "extras"

jdufresne:
https://tox.readthedocs.io/en/latest/config.html#conf-extras

Allows simplifying the configuration. Can reuse the version strings from
setup.py, allowing it to be a single source of truth.


jschneier:
Opened #751 with this and will merge if it passes. Thanks.

PR #632, opened 2018-12-07
Update amazon-S3.rst

martialp:
Source: https://github.com/jschneier/django-storages/pull/173#issuecomment-232460528

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -28,6 +28,8 @@ To allow ``django-admin.py`` collectstatic to automatically put your static file
 ``AWS_SECRET_ACCESS_KEY``
     Your Amazon Web Services secret access key, as a string.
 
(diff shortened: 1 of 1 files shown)


docs/backends/amazon-S3.rst
```diff
@@ -28,6 +28,8 @@ To allow ``django-admin.py`` collectstatic to automatically put your static file
 ``AWS_SECRET_ACCESS_KEY``
     Your Amazon Web Services secret access key, as a string.
 
+      Note that if AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are not set, boto internally looks up iam credentials.
```
jdufresne: The AWS docs always capitalizes IAM. I think we should follow that convention.

https://aws.amazon.com/iam/

---

You can put backticks around the settings so they render as code:

```rst
``AWS_ACCESS_KEY_ID``
```

---

boto -> boto3 as it is the preferred library.

PR #641, opened 2018-12-20
publicRead ACL

mj8894:
- handling defaultACL when creating a bucket file using write operation.


mj8894:
I agree.

However, the idea here was to apply default acl to newly created files using write instead of save option.

mj8894:
Made Changes to apply default acl only if a new blob is being uploaded to bucket, not affecting ACLs of the already available objects.

cupojoe:
@sww314 Why hasn't this being merged? Anything we can do to get it in? In order for the library to respect the default ACL I had to do a pretty ugly workaround to use save instead of the recommended write.

jschneier:
Fixed by #756.

PR #656, opened 2019-01-22
Add support for AZURE_CONNECTION_STRING

### LINKED ISSUE #643: custom domain for azure

taymonbeal:
This allows arbitrary connection settings to be specified, as long as the Azure Python library supports them.

Not sure how to test this, since the existing tests mock this part out.

Fixes #643.


nitely:
this is fixed in #658

PR #658, opened 2019-01-25
Support all blobstorage settings

### LINKED ISSUE #643: custom domain for azure

nitely:
fixes #643

Changes:

- Add support for all blostorage settings
- Update docs

> AZURE_CUSTOM_DOMAIN is important to fetch azurite from a linked container

### DIFF
--- a/docs/backends/azure.rst
+++ b/docs/backends/azure.rst
@@ -76,8 +76,6 @@ Set the default storage (i.e: for media files) and the static storage
 
 The following settings are available:
 
-    is_emulated = setting('AZURE_EMULATED_MODE', False)
-
 ``AZURE_ACCOUNT_NAME``
 
     This setting is the Windows Azure Storage Account name, which in many cases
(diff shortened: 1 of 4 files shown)


TheArtOfPour:
Great changes! Can you explain how to configure the CDN url with these new variables?

nitely:
Yeah, use the custom_domain setting for that. Here are the docs https://docs.microsoft.com/en-us/azure/storage/blobs/storage-custom-domain-name

malderete:
We should merge this PR. It improves the integration with Azure library's service.

jschneier:
Hi this looks okay. My only question is about the required azure library bump, are there breaking changes from 3 to 4?

nitely:
> My only question is about the required azure library bump, are there breaking changes from 3 to 4?

I double checked, all settings are there[0], so the version bump is not required. I temporarily pinned the version and all tests pass as well.

Not sure why I bumped them, maybe the changes did not in `CloudStorageAccount` or I had an old azure version :man_shrugging: 

[0] https://github.com/Azure/azure-storage-python/blob/v1.3.1-blob/azure-storage-blob/azure/storage/blob/blockblobservice.py#L96

davidfischer:
Firstly, this is great work.

I'm having some trouble getting this to work with an Azure CDN setup as opposed to a blob service custom domain. From what I can tell when I turned up debugging, it looks like the custom domain is used as an API endpoint.

However, when used with Azure CDN, the custom domain resolves to `BUCKETNAME.azureedge.net.` rather than `BUCKETNAME.blob.core.windows.net` and then the API calls seem to fail.

Am I misunderstanding what the `custom_domain` setting is supposed to do? Should it work with a CDN setup?

nitely:
Interesting. No, it won't work, because as you said, it will try to upload the file to the custom domain (cdn) and fail. I think we should add two internal services, one to generate the URLs using the custom_domain (cdn) and another one to upload/download through blob.core. I've to try whether that plays nice with the endpoint_suffix, though. I'll try it tomorrow and ping you so you can try it as well @davidfischer

nitely:
I made some changes and this has worked for me:

* create a new file
* read the new file
* access the new file with through CDN (using the `file.url` property)

Can you give this a try @davidfischer

davidfischer:
The changes worked for me. I tried:

- listing directories
- testing whether files exist
- getting the URL of a file (using `storage.url('path/to/file')`) - the custom domain was used in this case

Maxyme:
Hi, what is the progress on this? This is really useful for using Azurite inside of docker-compose and setting the AZURE_CUSTOM_DOMAIN to point to that container.
Thank you for this PR.

akshaybabloo:
Any update on this?

nitely:
It's up to @jschneier to merge it. I'm guessing he's waiting for the next minor (v1.8) release. We have been using this for a few months now and it has worked well so far.

jschneier:
Ugh I screwed up resolving the conflict in the GUI and now am not sure how to undo it.

jschneier:
Moved this to #750. Sorry for that and thanks again @nitely. Would you be interested in helping to maintain the Azure backend?

Issue #606, opened 2018-09-24
Error upload file. Use zappa+ lambda + s3 + django2 + RDS. 


davidllauce:
**When I save in local it keeps me well in S3 but when it is running with lamda I get the following error.**
`
Request Method: | POST
-- | --
https://s5h4s4t98f.execute-api.us-west-2.amazonaws.com/dev/admin/publicidad/publicidad/add/
2.1.1
ClientError
An error occurred (InvalidToken) when calling the PutObject operation: The provided token is malformed or otherwise invalid.
/var/runtime/botocore/client.py in _make_api_call, line 612
/var/lang/bin/python3.6
3.6.1
['/var/task',  '/var/runtime/awslambda',  '/var/runtime',  '/var/lang/lib/python36.zip',  '/var/lang/lib/python3.6',  '/var/lang/lib/python3.6/lib-dynload',  '/var/lang/lib/python3.6/site-packages',  '/var/task/setuptools-39.1.0-py3.6.egg',  '/var/task']
Lun, 24 Sep 2018 00:27:27 -0500

`

mongkok:
Try to downgrade the installed version, it works for me.

```sh
pip install django-storages==1.6.6
```

jschneier:
How can I reproduce the issue?

dpretty:
I suppose the solution needs to be a django setting e.g. `AWS_SECURITY_TOKEN_IGNORE_ENVIRONMENT` which, when set to `True` won't try and load the security token / session token from environment variables?

While AWS Lambda also populates the environment with `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`, django-storages allows these to be overridden in django settings. My setup generally has these loaded from the environment, but with a different variable name (e.g. `AWS_APP_ACCESS_KEY_ID`). However, it doesn't feel right setting `AWS_SESSION_TOKEN` or `AWS_SECURITY_TOKEN` in django settings as these will change over time, so probably we should provide a django setting which instead allows the behaviour of loading the security token from the environment to be overridden.

jschneier:
Thanks for the investigation. I’m trying to figure out the cleanest way for this to work for everyone. My use case is obviously different but it’s clear that there are a lot of different way this can be setup. As an initial step I am leaning towards partially reverting b13efd92b3bf3e9967b8e7819224bfcf9. Would you be willing to write up some docs for some ways to configure access?

[...]

PR #665, opened 2019-02-08
small change for better extending the S3BotoStorage class

thesunlover:
relates to #606 
This one will allow to extend the base class and configure how the credentials are stored/managed/refreshed.
Useful in case of Django deployed on AWS Lambda (15 min live) when using Zappa as a deploy tool. Zappa is configured to use aws-role for the update/deploy actions.
And AWS roles rotate their security credentials each 1h to 12hours, depending on the configured values.

check comments 
https://github.com/jschneier/django-storages/issues/606#issuecomment-461762380
https://github.com/jschneier/django-storages/issues/606#issuecomment-461774432

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,6 +36,7 @@ By order of apparition, thanks:
     * Alex Watt (Google Cloud Storage patch)
     * Jumpei Yoshimura (S3 docs)
     * Jon Dufresne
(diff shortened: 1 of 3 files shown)


jschneier:
You can already override those exact methods and it should work I think?

PR #670, opened 2019-02-18
Adds more of the supported libcloud driver init params to the provider settings dict

richardARPANET:
Previously the provider settings would not allow for these params and so django-storages could not function against a locally running server.

Here are the params in libcloud https://github.com/apache/libcloud/blob/bc097ffe5b04990e0ca302231388ee7d5d6fe911/libcloud/common/base.py#L902

---

I'm actually not a fan of this implementation as it couples django-storages more to the current implementation of libcloud. But am unsure of other effects which could occur from just passing all additional params from provider django settings dict into the `Driver()` init, as I'm new to this codebase.


richardARPANET:
@jschneier Any chance you could take a look at this PR?

richardARPANET:
Is this project no longer being maintained?

jschneier:
I’m transitioning to a new role starting in mid-April. I’m going to try and get this library to good place and then hand over the reins. Sorry for the delay here.

PR #671, opened 2019-02-20
Fix Azure name cleanup

### LINKED ISSUE #609: Azure backend strips some non-alphabetic characters from the file name
https://github.com/jschneier/django-storages/blame/master/storages/backends/azure_storage.py#L109

Is there any reason for stripping characters other than:
* `a`-`z` `A`-`Z` `0`-`9`
* `_`
* `-`
* `.`
* `/`

?

As much as I learn from the comment, url-escaping will do for special characters that occur inbetween .

nitely:
fixes #609

Changes:

* remove stripping special characters from file name
* escape URL reserved and unsafe characters on `url` method

Notes:

* This is consistent with the other storages
* Django does the same name cleaning as the previous code, anyway
* Django name cleaning can be bypassed by overriding `get_valid_name` or calling `default_storage.save` directly. This was not possible with the old code
* Since this is a superset of what was previously allowed, it's a minor breaking change
* Files uploaded with the previous code will still work

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 import mimetypes
-import re
 from datetime import datetime, timedelta
 from tempfile import SpooledTemporaryFile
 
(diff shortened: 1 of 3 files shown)


jschneier:
@nitely is this still necessary with your other branch now merged?

nitely:
Yes, this removes some silly restrictions in file names. I made anew PR with all conflicts resolved #752

PR #676, opened 2019-03-02
Allow S3Transfer config to be customised

michalc:
A typical use for this would be to control number of threads, or even disabling threads. For example, some transfers are unlikely to be faster than the transfer to/from S3, and so triggering threads is unlikely to be helpful.

The setting AWS_S3_TRANSFER_CONFIG is used twice so that S3Boto3Storage and S3Boto3StorageFile, which can be used directly in client code, can both be customised.


jschneier:
Superseded by #1280. I guess 4 years later is better than never...

Issue #680, opened 2019-03-06
Azure backend pulls broad and outdated dependencies


tiloc:
I am struggling getting the Azure backend to work with my other code, due to its dependencies. It pulls in the "azure" package, which is a wrapper to pull in a wide range of Azure SDKs. It hasn't been updated in quite a bit and thus conflicts with some of my code which has newer Azure SDK dependencies. It also pulls in much more than seems to be needed for the storage backend.
Could this be limited to use the "azure-common" package instead? Aside from "azure-storage-blob" this seems to be the only dependency which is really required.

PR #681, opened 2019-03-06
Limit and update required Azure packages

tiloc:
Limit and update required Azure packages. See issue #680

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires=['Django>=1.11'],
     extras_require={
-        'azure': ['azure>=3.0.0', 'azure-storage-blob>=1.3.1'],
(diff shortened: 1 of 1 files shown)


dazzag24:
+1 for this.  Requiring the base azure package pulls in far too many azure sub modules that are not needed. This is making our CI tests slower than they need to be.
Thanks

setup.py
```diff
@@ -15,7 +15,7 @@ def read(filename):
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires=['Django>=1.11'],
     extras_require={
-        'azure': ['azure>=3.0.0', 'azure-storage-blob>=1.3.1'],
+        'azure': ['azure-common', 'azure-storage-blob>=1.3.1'],
```
nitely: `azure-common` should be removed. `azure-storage-blob` already depends on it (so it will install it)

akshaybabloo: True

jschneier: I took this into consideration in #684

PR #688, opened 2019-04-08
Update azure_storage.py

zypriafl:
Use AZURE_CONNECTION_TIMEOUT_SECS instead of hardcoded value of 10 seconds


storages/backends/azure_storage.py
```diff
@@ -48,7 +48,7 @@ def _get_file(self):
                 blob_name=self._path,
                 stream=file,
                 max_connections=1,
-                timeout=10)
+                timeout=self.timeout)
```
nitely: This should be `self._storage.timeout`

PR #697, opened 2019-05-10
DigitalOceanSpacesStorage

mcueto:
I was implementing this package in order to use DigitalOcean Spaces and i lost 2 days figuring out how exactly the settings must be set to work, so i decided to create an Storage Backend called `DigitalOceanSpacesStorage` to keep clear when an user is using DigitalOcean Instead of Digital Ocean.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,6 +36,7 @@ By order of apparition, thanks:
     * Alex Watt (Google Cloud Storage patch)
     * Jumpei Yoshimura (S3 docs)
     * Jon Dufresne
+    * Marcelo Cueto (DigitalOceanSpacesStorage)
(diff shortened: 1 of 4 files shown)


sww314:
@mcueto thanks for the contribution! Looks like the lint and some of the tests are failing. Once those are fixed, I will review in more detail.

chriswedgwood:
Hi both I have taken this branch and created a new pull request #741 to fix the tests and flake8 issues
If there is a better way to do things?
thanks
Chris

jschneier:
Sorry, how did you lose 2 days? I don't think this adds very much value & adds a lot of new code. I'd rather expand the docs.

PR #698, opened 2019-05-12
Google Storage Backend: settings ACLs when file is uploaded

stmos:
There is no need to make a separate call to set ACLs on a file (blob). ACLs can be passed as a parameter to the "upload_from_file" function.

On a side note, the reason I started looking at this code is that over the last few days I started getting errors when uploading files to Google cloud: "503 GET https://www.googleapis.com/storage/v1/b/BUCKET_NAME/o/FILE_PATH/acl: Backend Error"

I don't have a clear explanation on why this was happening, I am guessing that setting ACLs right after the file was uploaded may have caused some contention issues on Google side. (the issue was totally random)

With the change proposed in this PR I don't see any more issues when uploading files.

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -175,9 +175,7 @@ class GoogleCloudStorage(Storage):
         file = GoogleCloudFile(encoded_name, 'rw', self)
         file.blob.cache_control = self.cache_control
         file.blob.upload_from_file(content, rewind=True, size=content.size,
-                                   content_type=file.mime_type)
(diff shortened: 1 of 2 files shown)


sww314:
@stmos thanks for this. 
The tests are not passing - can you take a look at them. 

This change also subtly changes the way things work. It always sets the ACL which may be unset.
You need to check if the default_acl is set before using it.

stmos:
Tests look good now. Thanks!

PR #718, opened 2019-06-25
Google: Only verify bucket if auto_create is True

stanislavb:
Motivation for this change is to send fewer requests to Google Storage
API. We do not check existence of bucket unless auto_create_bucket is
configured or exists('') is called.

When creating bucket, instead of making two requests to get and create,
we just create and look for conflict.

Based on the work in #575 and supersedes #412.

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -15,9 +15,8 @@ from storages.utils import (
 )
 
 try:
-    from google.cloud.storage.client import Client
-    from google.cloud.storage.blob import Blob
-    from google.cloud.exceptions import NotFound
+    from google.cloud.storage import Blob, Client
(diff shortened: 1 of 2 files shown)


storages/backends/gcloud.py
```diff
@@ -191,9 +190,9 @@ def delete(self, name):
     def exists(self, name):
         if not name:  # root element aka the bucket
             try:
-                self.bucket
+                self.client.get_bucket(self.bucket)
                 return True
-            except ImproperlyConfigured:
+            except (ImproperlyConfigured, NotFound):
```
stanislavb: ImproperlyConfigured can no longer be caught here, I think.
```suggestion
            except NotFound:
```

jschneier:
Looks great, thanks.

PR #723, opened 2019-07-08
Azure: Adds missing parameter sp (permission) to Azure signature

### LINKED ISSUE #705: Invalid sas_token was generated
https://github.com/jschneier/django-storages/blob/b441b74a17a46eb87ee4b10f60774b9a080c0fe1/storages/backends/azure_storage.py#L260

I think this should be corrected as follows.

sas_token = self.service.generate_blob_shared_access_signature(
                self.azure_container, name, permission=BlobPermissions.READ, expiry=self._expire_at(expire))

This problem is caused by
https://github.com/Azure/azure-storage-python/issues/595

hbaltz:
Hello,

 Microsoft adds new parameters to its function generate_blob_shared_access_signature (to see more details: [Azure storage python issue 595](https://github.com/Azure/azure-storage-python/issues/595))

 Without the permission parameter in generate_blob_shared_access_signature, the URL generated by Azure sdk is incorrect and we have the following error : `sp is mandatory. cannot be empty`

 This PR should correct this and fix #705

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -257,7 +257,7 @@ class AzureStorage(Storage):
         make_blob_url_kwargs = {}
         if expire:
             sas_token = self.service.generate_blob_shared_access_signature(
-                self.azure_container, name, BlobPermissions.READ, expiry=self._expire_at(expire))
+                self.azure_container, name, permission=BlobPermissions.READ, expiry=self._expire_at(expire))
             make_blob_url_kwargs['sas_token'] = sas_token
 
         if self.azure_protocol:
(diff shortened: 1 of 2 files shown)


frankie567:
+1 This problem prevents to retrieve the URL of the files in Azure.

felixschul:
@jschneier Do you intend on making a release on PyPi for this? I currently integrate the master-branch, since I require this fix. Or am I doing something wrong? The last tag/release seems to be in September 2018? Thanks for your awesome work BTW!

jschneier:
Yes I plan on it but as usual open source is hard.

PR #728, opened 2019-07-24
Basic azure CDN support

cmberryau:
Tested and works so far for cheapest MS CDN option.. no support for Verizon prem with SAS keys.


jschneier:
This is contained with #750.

PR #745, opened 2019-08-29
Update dropbox.py

btywoniuk:
Added settings parameters as in s3boto3.py that way its easier to modify them when creating a  subclass


storages/backends/dropbox.py
```diff
@@ -66,17 +66,18 @@ def _set_file(self, value):
 @deconstructible
 class DropBoxStorage(Storage):
     """DropBox Storage class for Django pluggable storage system."""
+    location = setting('DROPBOX_ROOT_PATH', '/')
+    oauth2_access_token = setting('DROPBOX_OAUTH2_TOKEN')
+    timeout = setting('DROPBOX_TIMEOUT' _DEFAULT_TIMEOUT)
```
jschneier: Ugh I missed the comma here in the resolution of the conflicts.

PR #746, opened 2019-09-04
SFTPStorage: Fix reopen file

igerko:
Default Django implementation uses os.path.exists and open(), which is wrong.


jschneier:
Thanks.

PR #750, opened 2019-09-09
Azure: Support all blobstorage settings

### LINKED ISSUE #643: custom domain for azure

jschneier:
Non messed up rebased version of #658.

Closes #643. 
Closes #656.


jschneier:
Hey @nitely in the `url` function there is a check for `if self.azure_protocol`. That appears to always be truthy, no?

nitely:
Hey. You are right, there's an unnecessary check there. The value should be passed as `protocol=self.azure_protocol` instead.

PR #709, opened 2019-05-25
Drop support for Python 2

jdufresne:
Python 2 support is nearing end of life. At the end of this year, it will no longer be supported, including for security issues.
    
Django has already dropped support for Python 2 in Django 2.0. Follow upstream to be more compatible with the master branch as tests now pass.
    
For more details, see https://pythonclock.org/
    
All code changes and clean ups that were noticed were done. Changes include:
    
- Remove str/unicode compatibility shims.
- Remove encoding cookies from source files. Python 3 interprets all files as utf-8 by default.
- Remove unnecessary `__future__` imports. The features are now built-in.
- Use modern and simpler `super()` syntax.
- Update the wheel configuration to reflect it is no longer universal.
- Update trove classifiers.
- And more …
    
Also remove supported for end of life Python 3.4. It went end of life 2019-03-18.

PR #759, opened 2019-09-10
Added error handling at import django.utils.six module. #758

### LINKED ISSUE #758: Support django 3.0
I tried django-storages 1.7.2 with [Django 3.0a1](https://pypi.org/project/Django/3.0a1/).
I found that django-storage uses the _**django.utils.six**_ module which [will be removed in django 3.0](https://docs.djangoproject.com/en/dev/releases/3.0/#removed-private-python-2-compatibility-apis).

> django.utils.six - Remove usage of this vendored library or switch to [six](https://pypi.org/project/six/).

The module was used in the following places:
- storages
  - backends
    - s3boto3.py
    - apache_libcloud.py
    - sftpstorage.py
    - ftp.py
- tests
  - test_s3boto.py
  - test_s3boto3.py

yukihira1992:
I fixed #758.
If django-storages ends support for Python2, close this.

### DIFF
--- a/storages/backends/apache_libcloud.py
+++ b/storages/backends/apache_libcloud.py
@@ -9,8 +9,13 @@ from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import File
 from django.core.files.storage import Storage
 from django.utils.deconstruct import deconstructible
-from django.utils.six import string_types
-from django.utils.six.moves.urllib.parse import urljoin
+
+try:
+    from django.utils.six import string_types
+    from django.utils.six.moves.urllib.parse import urljoin
+except ImportError:
+    string_types = str
+    from urllib.parse import urljoin
 
 try:
     from libcloud.storage.providers import get_driver
--- a/storages/backends/ftp.py
+++ b/storages/backends/ftp.py
@@ -24,9 +24,13 @@ from django.core.exceptions import ImproperlyConfigured
 from django.core.files.base import File
 from django.core.files.storage import Storage
 from django.utils.deconstruct import deconstructible
-from django.utils.six.moves.urllib import parse as urlparse
 
 from storages.utils import setting
+
+try:
+    from django.utils.six.moves.urllib import parse as urlparse
+except ImportError:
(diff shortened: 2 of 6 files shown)


jschneier:
Thanks for this, I'm not quite sure why they removed `six` before 1.11 is no longer supported. I think a better way to solve this is to do `try` `except` rather than depending on `six`.

Want to take a crack at that?

yukihira1992:
Removed `six` and added error handling.

jdufresne:
I think it would be overall simpler maintenance to follow Django's lead and drop support for Python 2. It is very close to EOL anyway. It will required fewer compatibility shims and workarounds. Work for this has started #709

codingjoe:
Thanks for addressing this issue. I guess it blocked some people from testing the Django alpha.

> Thanks for this, I'm not quite sure why they removed `six` before 1.11 is no longer supported. I think a better way to solve this is to do `try` `except` rather than depending on `six`.

@jschneier I understand the confusion, yet `six` has undergone the regular deprecation cycle, which wasn't required, considering that it was never part of the official Django API.

> I think it would be overall simpler maintenance to follow Django's lead and drop support for Python 2. It is very close to EOL anyway. It will required fewer compatibility shims and workarounds. Work for this has started #709

@jdufresne I second that. I believe there is no need to wait until 1.11 hits EOL in April, since 1.11 extended support does not cover Python 2.7 since it's EOL is in December. So one should only use Django 1.11 on Python 3 after December.

It's absolutely admirable to keep support Python 2 for that long, but I guess it's time for django-storages v2.0.0 ;)

jschneier:
This library saw such a dearth of development for such a long time that I am attempting to put in all the minimum necessary features before doing something drastic and breaking backwards combat too string. When I release 2.0 I think we will also move to strict semantic versioning - given where we started I found it impractical to start there. I understand the EOL timeline, however.

codingjoe:
@jschneier I hear you. If you need any help, you just say the word. Thanks for maintaining this beast, much appreciated.

PR #760, opened 2019-09-11
Update gcloud.rst

iarambula:
Had I really hard time figuring out what packages I needed to install.


dcecile:
The original pip install command worked for me without this change 🤔 

Was there an error?

jschneier:
The extra should work fine. If the packages change then we'd have update it in a bunch of places.

PR #767, opened 2019-09-25
Run pyupgrade across codebase to automate cleanups

jdufresne:
https://github.com/asottile/pyupgrade

pyupgrade is a tool to automatically upgrade syntax to use features and
conventions of newer versions of Python.


jschneier:
Thanks. Not sure if `{}` formatting is better than `%s`. f-strings clearly are.

Issue #784, opened 2019-11-01
Add support for azure-storage-blob>=12.0.0


larserikgk:
Yesterday version 12.0.0 of [`azure-storage-blob`](https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-storage-blob_12.0.0) was released.

First of all I would like to request pinning versions of required libraries so new releases of these do not instantly break django-storages if they have any breaking changes. In my opinion it is much better to pin the requirements and update manually when needed/requested.

My nightly build using django-storages for Azure failed, and the first issue I found was the renaming of `BlobPermissions` to `BlobSasPermissions`. This must be updated in [azure_storage.py](https://github.com/jschneier/django-storages/blob/master/storages/backends/azure_storage.py#L8)

I have not looked through the rest of the changes yet, but there might be more issues hidden in the release notes. I will try to find the issues and submit a PR.

An immediate quick fix would be to pin `azure-storage-blob` to version `2.1.0`. The [`azure`] module uses `1.3.1`, so that is an option as well.

Release notes for azure-storage-blob 12.0.0:
* [12.0.0](https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-storage-blob_12.0.0)
* [12.0.0b4](https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-storage-blob_12.0.0b4)
* [12.0.0b3](https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-storage-blob_12.0.0b3)
* [12.0.0b2](https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-storage-blob_12.0.0b2)
* [12.0.0b1](https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-storage-blob_12.0.0b1)

Bastien-Brd:
+1 👍

larserikgk:
There are more changes than just renaming, so I suggest a requirement pin until someone has time to update the code according to 12.0.0 of [azure-storage-blob]. Please see #785

jschneier:
Yeah, many changes. For now I will just pin to `>1.3.1<2.1`. I'm not a fan of pinning altogether.

jschneier:
Sorry, let me clarify. Since most things (this library excepted) now use semantic versioning I think that pinning to version ranges is acceptable, spinning to specific versions is not though.

I'm a bit surprised at how much churn this in the relevant libraries for django-storages.

matthewgdv:
[...]

PR #785, opened 2019-11-04
Pin dependency of azure storage blob to revert breakage by version 12

larserikgk:
This is a temporary fix for #784


rik:
Latest version before 12.0.0 is 2.1.0. I'd suggest using a range rather than a strict version.

larserikgk:
Fair point @rik. I pinned to 1.3.1 since that is what the `azure` module wants, but I agree with you.

dmpe:
@jschneier can you please take a look on this ?
thanks

jschneier:
Thanks.

PR #801, opened 2019-12-14
DigitalOcean Storages Improvements and customisation

chriswedgwood:
Hi I've carried on from the #697 pull request and fixed the test and flake8.
I wasn't able to do a pull request against https://github.com/mcueto/django-storages so I have done it against master. Please advise if there is a better way :)

Thanks
 Chris


jschneier:
I don't think this adds a ton of value while also increasing the amount of code quite a bit. I don't see how, e.g all those tests are worth it when they basically just duplicate the boto3 tests.

Expanding of the docs, if there are DO specific things, would be fine.

Issue #435, opened 2017-11-30
S3Boto3StorageFile close() should mimic FileSystemStorage.close() for empty files


Masterdanielsan:
**S3Boto3StorageFile** close() vs. **FileSystemStorage**.close()
The following code will fail to write a file in the former case but not in the latter:

Setup: 'some/path/to/file.txt' doesn't exist in S3 or on File system in either case

```
file = default_storage.open('some/path/to/file.txt', 'w')
file.close()
```

One shouldn't have to call file.write() before file.close() in order to create an empty file.

This would make S3Boto3StorageFile consistent with a regular python open() command as well as the FileSystemStorage.open() command.

PR #816, opened 2020-01-15
Create empty file if no bytes written to S3 file in 'w' mode

chrisrink10:
This PR adds functionality to `S3Boto3StorageFile` that will create an empty file if no bytes are written to S3 after opening a file in `w` mode, similar to how `FileSystemStorage` and Python's native `open` work.

Let me know if I'm missing anything or if you have any questions. Thank you!

Fixes https://github.com/jschneier/django-storages/issues/435

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -38,6 +38,7 @@ By order of apparition, thanks:
     * Jon Dufresne
     * Rodrigo Gadea (Dropbox fixes)
     * Martey Dodoo
+    * Chris Rink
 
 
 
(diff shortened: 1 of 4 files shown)


chrisrink10:
Looks like my remaining build failure is due to extra warnings being thrown on Python 3.8 for `tests/test_s3boto3.py::S3Boto3Storagetests::test_deprecated_bucket`. I'm not exactly clear how to proceed with fixing this issue. I don't think my changes caused this.

jdufresne:
Why should this be optional? Shouldn't the backend always follow the upstream behavior of `FileSystemStorage`?

chrisrink10:
@jdufresne Completely agree it should not be optional in an ideal world. I was concerned it would be a breaking change though, since that's not how it works right now. If people think it should be standard behavior I'm happy to make it that way (it's easier for me).

cc @jschneier

blargoner:
@jschneier Any chance this can be merged? I love empty files.

jschneier:
Please remove the setting.

chrisrink10:
@jschneier Thanks for the reply. I removed the setting.

Looks like the Python 3.8 build issue is resolved now.

blargoner:
@jschneier Now that the setting is removed, any chance this can be merged? Thanks!

PR #837, opened 2020-02-12
Fix GCS signed urls when using custom endpoint.

martey:
Fix the generation of signed URLs in Google Cloud Storage when using a custom endpoint (see #648). The previous behavior generates an invalid URL due to the inclusion of the GCS bucket name (e.g. "https://custom_endpoint/bucket_name/filename?signed" instead of "https://custom_endpoint/filename?signed").

Since the "generate_signed_url" method does not natively support custom endpoints, create a signed URL with the standard endpoint and manually replace it (and the bucket name) with the custom endpoint.

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -17,7 +17,7 @@ from storages.utils import (
 
 try:
     from google.cloud.storage import Blob, Client
-    from google.cloud.storage.blob import _quote
+    from google.cloud.storage.blob import _API_ACCESS_ENDPOINT, _quote
     from google.cloud.exceptions import Conflict, NotFound
 except ImportError:
     raise ImproperlyConfigured("Could not load Google Cloud Storage bindings.\n"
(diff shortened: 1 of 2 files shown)


jschneier:
Is there a bug on the library about this?

martey:
> Is there a bug on the library about this?

https://github.com/googleapis/python-storage/issues/11. That issue doesn't have much information, but the same issue on the C++ version of the library (https://github.com/googleapis/google-cloud-cpp/issues/3328) makes it clear that "bucket-bound hostnames" are custom endpoints.

jschneier:
Looks like this was fixed in the underlying Python library (although with kind of bizarre param names).

martey:
Sorry for the delay on responding to this.

The change in google-cloud-storage doesn't affect this PR because it is only for v4 signed URLs and custom endpoints currently use v2 signed URLs (see https://github.com/jschneier/django-storages/pull/648#issuecomment-485151358 for more information on why this is). In addition, [it only exists in google-cloud-storage >= 1.27.0](https://github.com/googleapis/python-storage/blob/master/CHANGELOG.md#1270-2020-04-01).

martey:
Closing this because I believe my other pull request (see above) is a more elegant solution.

PR #587, opened 2018-09-05
[s3] Add support for signing CloudFront URLs

terencehonles:
Fixes #456. Similar to #484 but uses AWS CloudFront signed URLS ([See AWS docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html))

This is probably the right way to approach the problem in #484, but if that solution works that might be an alternative (not sure how @mattjegan configured S3+CloundFront)

Issue #165, opened 2016-07-03
Custom domain and query string auth


exp0nge:
I see that the query string auth feature does not work when a custom domain is provided. At least that's what it seems like from the source. Is this intentional? Is there a workaround to allow both features to work?

jonathan-golorry:
Technically the comments in settings imply this "just use this domain plus the path", but I agree that this is an issue that needs to be addressed.

exp0nge:
Are folks just not using auth with custom domains then?

jonathan-golorry:
Probably. I'm running two different s3 locations for static and media files and Django settings are rarely well documented, so I'm not really sure what settings are being used for what. I think it might be possible to adjust MEDIA_URL or STATIC_URL without setting AWS_S3_CUSTOM_DOMAIN and have it work. I had to set STATIC_URL for django.contrib.staticfiles, but I went with the default domain for other media instead of messing with MEDIA_URL.

tchaumeny:
Hi!
I just came across this problem too, it isn't possible to use a CUSTOM_DOMAIN for private media stored using S3 (the auth parameters just won't be added to querystring indeed). Is this because of some restriction in AWS / Boto ?

vchrisb:
If a ```AWS_S3_CUSTOM_DOMAIN``` is specified, presigned URLs are not generated:
https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L567-L569

tchaumeny:
@vchrisb Exactly the boto backend you are pointing to doesn't make it possible to use signed URLs with custom domain, although it is something possible (see for instance http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html#private-content-creating-signed-url-canned-policy-procedure, paragraph 2 with custom domain examples)

Jaime5:
This was my hacky solution below:


```python
"""
Note that below for media file objects I requested they not have a custom domain.

This is because they need to be signed in order to be viewed. 
Django-Storages only supports signing on files not set with a custom domain
"""

class StaticStorage(S3Boto3Storage):
    default_acl = None  # don't want any ACL applied ever.
    location = settings.AWS_STATIC_LOCATION

class PublicMediaStorage(S3Boto3Storage):
    location = settings.AWS_PUBLIC_MEDIA_LOCATION
    file_overwrite = False  # use case important
    default_acl = None  # don't want any ACL applied ever.
[...]

PR #839, opened 2020-02-14
Allow using AWS_QUERYSTRING_AUTH & AWS_S3_CUSTOM_DOMAIN at the same time

cpoetter:
Hi, 

I am currently trying to mock S3 in development with the help of https://github.com/adobe/S3Mock. Unfortunately I can't use signed URLs and a custom domain at the same time. This _should_ fix that.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -671,9 +671,7 @@ class S3Boto3Storage(Storage):
     def url(self, name, parameters=None, expire=None):
         # Preserve the trailing slash after normalizing the path.
         name = self._normalize_name(self._clean_name(name))
-        if self.custom_domain:
-            return "{}//{}/{}".format(self.url_protocol,
-                                      self.custom_domain, filepath_to_uri(name))
+
         if expire is None:
             expire = self.querystring_expire
 
@@ -682,6 +680,19 @@ class S3Boto3Storage(Storage):
         params['Key'] = self._encode_name(name)
         url = self.bucket.meta.client.generate_presigned_url('get_object', Params=params,
                                                              ExpiresIn=expire)
+
+        if self.custom_domain:
+            # Key parameter can't be empty. Use "/" and remove it later.
+            params['Key'] = '/'
+            root_url_signed = self.bucket.meta.client.generate_presigned_url('get_object',
+                                                                             Params=params,
+                                                                             ExpiresIn=expire)
+            # Remove signing parameter and previouly added key "/".
+            root_url = self._strip_signing_parameters(root_url_signed)[:-1]
+            # Replace bucket domain with custom domain.
+            custom_url = "{}//{}/".format(self.url_protocol, self.custom_domain)
+            url = url.replace(root_url, custom_url)
+
         if self.querystring_auth:
             return url
         return self._strip_signing_parameters(url)


terencehonles:
I think this is related/fixed by #587? I'm not sure if you have tried using that @cpoetter

nathando:
hi, can someone try to merge this in. https://github.com/jschneier/django-storages/pull/587 only solved issue if we use CloudFront. If we want to use our own proxy solution like nginx this is important

terencehonles:
> hi, can someone try to merge this in. #587 only solved issue if we use CloudFront. If we want to use our own proxy solution like nginx this is important

Looks like it's merged now

rahulkp220:
Is this fixed yet?

terencehonles:
> Is this fixed yet?

@rahulkp220 you can probably use this https://github.com/jschneier/django-storages/pull/587#issuecomment-628681209 from #587

rahulkp220:
Ohkay, I will have a look at it.

cjnething:
The suggestion to use #587 doesn't work for us, it would be great if this could be merged in 🙂

terencehonles:
> The suggestion to use #587 doesn't work for us, it would be great if this could be merged in

@cjnething there are merge conflicts and if it indeed doesn't work for the author, the author will need to update this PR so it can be considered (and to be clear, I would not be the person considering it as I only suggested a PR that I wrote may fix the issue if these are independent issues the author and the maintainer(s) would need to approve this PR).

terencehonles:
Based on the more detailed report in https://github.com/jschneier/django-storages/issues/165#issuecomment-653936017 I understand a little more what this is trying to fix (not sure if it's really the same problem others are running into), but if @cpoetter has a chance to update this PR I added some comments about the different states that need to be accounted for https://github.com/jschneier/django-storages/issues/165#issuecomment-653943359

cpoetter:
The merge conflict _should_ be resolved ✅

terencehonles:
> The merge conflict _should_ be resolved

Just a heads up, there are two `if self.custom_domain` blocks (yours is the second). The first always returns therefore the one you added will do nothing, and you'll want to adjust the PR to fix that.

kut:
Hey folks won't replacing the S3 URL with the custom URL after signing cause the signature to be invalid?

terencehonles:
The signature is not signing the whole URL but part of it. The part of it that's signed is not changing. I believe this PR still needs an update based on my last comment.

rissson:
@cpoetter are you still willing to work on this?

dvail:
Is this something that the maintainers would still want merged in? We have similar use cases, both with S3Mock and using nginx as a reverse proxy, and the fix in https://github.com/jschneier/django-storages/issues/165#issuecomment-810166563 seems to work for now.

I might be able to work a PR to get this into the main library though if it is something people still want.

jschneier:
> Is this something that the maintainers would still want merged in? We have similar use cases, both with S3Mock and using nginx as a reverse proxy, and the fix in [#165 (comment)](https://github.com/jschneier/django-storages/issues/165#issuecomment-810166563) seems to work for now.
> 
> I might be able to work a PR to get this into the main library though if it is something people still want.
> 
> Edit: Actually looking into this a bit more and it looks like this functionality will cause a possible breaking change if others are using a custom domain when `AWS_CLOUDFRONT_KEY_ID` is set, as the two code paths construct the URL in two different ways. A safer fix might be to introduce a new configuration setting, e.g. `AWS_REVERSE_PROXY_URL`, but that would be yet another option to maintain and complicate this code further.

@dvail Yes, I will take something like this since it seems there is a fair amount of demand and I think it will ease integrating with other S3-compatible services.

That comment seems more elegant than this method, mind opening a fix?

jschneier:
Hello, I'd still like to get this in and am thinking of something like the following which incorporates seemingly the best of the proposed solutions. However, I don't fully understand why `Key` is getting applied and then stripped off. If someone wants to take this patch and push it forwards a bit, I would be happy to merge.

```diff
diff --git a/storages/backends/s3boto3.py b/storages/backends/s3boto3.py
index 7231750..fb79c30 100644
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -562,33 +562,53 @@ class S3Boto3Storage(CompressStorageMixin, BaseStorage):
         split_url = split_url._replace(query='&'.join(joined_qs))
         return split_url.geturl()
 
+    def _get_custom_domain_url(self, name, params, expire):
+        url = '{}//{}/{}{}'.format(
+            self.url_protocol,
+            self.custom_domain,
+            filepath_to_uri(name),
+            ('?' + urlencode(params)) if params else '',
+        )
+
+        if self.querystring_auth:
+            if self.cloudfront_signer:
+                expiration = datetime.utcnow() + timedelta(seconds=expire)
+                url = self.cloudfront_signer.generate_presigned_url(
+                    url,
+                    date_less_than=expiration
+                )
+            else:
+                # TODO
+                url = self.bucket.meta.client.generate_presigned_url(
+                    'get_object',
+                    Params=params,
+                    ExpiresIn=expire,
+                    HttpMethod=http_method
+                )
+
+        return url
+
     def url(self, name, parameters=None, expire=None, http_method=None):
-        # Preserve the trailing slash after normalizing the path.
         name = self._normalize_name(clean_name(name))
         params = parameters.copy() if parameters else {}
+
         if expire is None:
             expire = self.querystring_expire
 
         if self.custom_domain:
-            url = '{}//{}/{}{}'.format(
-                self.url_protocol,
-                self.custom_domain,
-                filepath_to_uri(name),
-                '?{}'.format(urlencode(params)) if params else '',
-            )
-
-            if self.querystring_auth and self.cloudfront_signer:
-                expiration = datetime.utcnow() + timedelta(seconds=expire)
-                return self.cloudfront_signer.generate_presigned_url(url, date_less_than=expiration)
-
-            return url
+            return self._get_custom_domain_url(name, params, expire)
 
         params['Bucket'] = self.bucket.name
         params['Key'] = name
-        url = self.bucket.meta.client.generate_presigned_url('get_object', Params=params,
-                                                             ExpiresIn=expire, HttpMethod=http_method)
+        url = self.bucket.meta.client.generate_presigned_url(
+            'get_object', Params=params,
+            ExpiresIn=expire,
+            HttpMethod=http_method
+        )
+
         if self.querystring_auth:
             return url
+
         return self._strip_signing_parameters(url)
 
     def get_available_name(self, name, max_length=None):
```

jschneier:
Sorry folks this didn't make it into the most recent version still. It would be very helpful if someone could list out the configuration they want to have INCLUDING the DNS / domain mapping because the info here conflicts in several ways for me.

dopry:
this is my use case... 

given 
- AWS_S3_ENDPOINT_URL is any public AWS s3 service
- static.mysite.com is a CNAME which points to the bucket director or is proxied through any CDN capable of fronting the object store and passing through the signed query string such as Cloudflare, Fastly, Edgecast, Cloudfront, etc. 

```python
class CustomStorage(S3Boto3Storage):
    bucket_name = 'example'
    custom_domain = 'static.mysite.com'
    url_protocol = 'http'
    querystring_auth = True
    default_acl = "private"
```

amoralesc:
Commenting here to report that there is a current hack of replacing the endpoint URL after it has been generated, but only works with Signature Version 2. More info here https://github.com/jschneier/django-storages/issues/165#issuecomment-1920497275.

danilo-botelho:
After going to great lengths to make it work, I finally got it with just _AWS_S3_PROXIES_ for a local setup:

```
AWS_S3_ENDPOINT_URL = "http://localhost:9000"
AWS_S3_PROXIES = {"http": env("S3_ENDPOINT")}
```

**S3_ENDPOINT** environment var points to my local docker Minio server (minio:9000).

With a global _AWS_QUERYSTRING_AUTH = False_ I can control url signing with _OPTIONS_:

 ```
"BACKEND": "storages.backends.s3boto3.S3Boto3Storage",
 "OPTIONS": {"bucket_name": MINIO_BUCKET_NAME, "querystring_auth": True}
```

PR #846, opened 2020-02-28
Add note about not being able to use Uniform access control on google cloud bucket

devxpy:
Hello, thanks for this amazing project.

I just wanted to add a small note to the docs. 

I spent a few hours trying to figure this out, so I thought this would help others who might have set Uniform access control on their bucket.


jschneier:
Thanks!

Issue #851, opened 2020-03-03
Subclassing SFTPStorage


DoctorJohn:
I'd like to subclass the SFTPStorage class so that I can have multiple SFTP backends. That's already conveniently possible with most of the other backends, as their settings can be provided using class variables.

The SFTPStorage class, however, only works with global settings or parameters passed to it's `__init__` method. Thus subclassing requires overriding `__init__`, which doesn't feel as clean as providing class variables.

Is this intended? Otherwise I could create an PR which would adjust the SFTPStorage to allow convenient subclassing similar to the other backends.

jschneier:
Sure, would absolutely accept a PR

DoctorJohn:
@jschneier I prepared a PR (#852), it's ready to be reviewed.

jschneier:
Closed via #852

PR #852, opened 2020-03-03
Allow SFTPStorage to be configured via class variables

DoctorJohn:
This PR makes the SFTPStorage class configurable via class variables (similar to other backends). That way subclassing is more convenient. (#851).

As far as I can tell, these changes are backwards compatible.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -44,6 +44,7 @@ By order of apparition, thanks:
     * Manuel Kaufmann (humitos)
     * Taras Petriichuk (Dropbox write_mode option)
     * Zoe Liao (S3 docs)
+    * Jonathan Ehwald
 
 
 Extra thanks to Marty for adding this in Django,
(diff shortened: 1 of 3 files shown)


jschneier:
Can you please rebase

jschneier:
It looks good, will merge after

jdufresne:
I think this should hold off until #524 lands, and then use the approach developed there. That way we don't end up with the same issue in the SFTPStorage backend.

DoctorJohn:
@jschneier I rebased, however @jdufresne has a point. After #524 landed, I could alter this PR or drop it and create a new one, that has the same affect but uses "the style" introduced in #524.

jdufresne:
The other PR has landed, this could be rebased and updated if you'd like.

jschneier:
@DoctorJohn are you interested in updating this? I'm going to make another release imminently.

DoctorJohn:
@jschneier I just updated the SFTPStorage backend to use the settings mechanism introduced in #524 while keeping everything backwards compatible.

Issue #853, opened 2020-03-05
Add support to specify `http_method` on `storage.url`


humitos:
When using AWS S3 backend storage, `storage.url` uses [`generate_presigned_url`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.generate_presigned_url) behind the scenes to generate a valid URL with the querystring attribute required to access directly to the file with a proper signature and more.

S3 signature is based on HTTP method and since `storage.url` is not passing the `HttpMethod` to `generate_presigned_url`, it uses the "method's model" by default (which in this case is GET because of `get_object`).

It would be good if `storage.url` could accept `http_method` (defaulting to GET) in a similar way as it accept `parameters` and `expire` and [pass it along to `generate_presigned_url`](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L683-L684) so `storage.url` can return valid URLs when using other HTTP methods different than GET.

PR #854, opened 2020-03-05
Add S3 support for `http_method` in `storage.url`

humitos:
I went ahead and wrote a proposal for #853.

Allow to pass `http_method` to `storage.url` in order to calculate the signature accordingly. By default, it uses ~~'GET'~~ `None` to keep backward compatibility.

Please, let me know if this makes sense to you and suggest any changes you consider appropriate.


jschneier:
Looks great, thanks

PR #857, opened 2020-03-07
Correct name of warning filter "all" → "always"

jdufresne:
Per the Python documentation, there is no "all" filter. It looks like
"always" was intended:

https://docs.python.org/3/library/warnings.html#the-warnings-filter

The filter "all" happens to work as it is an undocumented alias, but
might as well used the documented name.

https://github.com/python/cpython/blob/v3.8.2/Lib/warnings.py#L244

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,7 @@ envlist =
 [testenv]
 setenv =
 	DJANGO_SETTINGS_MODULE = tests.settings
-	PYTHONWARNINGS = all
+	PYTHONWARNINGS = always
 	PYTHONDONTWRITEBYTECODE = 1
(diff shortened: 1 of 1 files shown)


jschneier:
Just as a sanity check - any idea how long back the history goes?

jdufresne:
The use of "all" was introduced in cecb101a8eacf3dda708e251471890941fc994a8 (Mar 12, 2018). Is that what you're looking for? Let me know if I can provide any other information.

jschneier:
Just the normal sort of backwards compat breaking hesitation.

jdufresne:
This change is to `tox.ini` which is only used internally by django-storages for testing only and is not exposed as part of the outward API. So I don't think there is any concern about backwards compatibility.

If I misunderstood, could you provide more detail?

PR #868, opened 2020-03-31
GCS: Retrying using an exponential backoff

### LINKED ISSUE #867: GCS: Retrying failed requests
Working with a GoogleCloudStorage API may fail due to some transient server side errors, such as `TooManyRequests`, `ServiceUnavailable`, `InternalServerError`, etc. Retrying failed requests is [recommended by Google](https://cloud.google.com/storage/docs/key-terms#immutability) for some [specific errors](https://cloud.google.com/apis/design/errors#error_retries). As an example, _gsutil tool_, which is written in Python, has it's own [retry handling strategy](https://cloud.google.com/storage/docs/gsutil/addlhelp/RetryHandlingStrategy#retry-strategy_1) (more information on what errors are considered to be retryable can be found there). These errors could be handled with an [exponential backoff](https://cloud.google.com/iot/docs/how-tos/exponential-backoff)  retrying algorithm. In case, if that does not help, it would still prevent the client from making too many requests and and retrying indefinitely.
Beside that, aforementioned _gsutil tool_ also retries on transient network errors, and I consider the one specified in @and3rson's issue jschneier/django-storages#735 to be one of those. Using a standard Google's retry handler allows to handle such exceptions by appending the list of exceptions that must be handled.

defined-user:
Resolves #867

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -42,6 +42,7 @@ By order of apparition, thanks:
     * Shaung Cheng (S3 docs)
     * Andrew Perry (Bug fixes in SFTPStorage)
     * Manuel Kaufmann (humitos)
+    * Andrii Skoryk (Google Cloud Storage)
 
 
 Extra thanks to Marty for adding this in Django,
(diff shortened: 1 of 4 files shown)


jakepoz:
@jschneier Can you please review? We have been using this in production for over a month successfully.

PR #873, opened 2020-04-14
Write mode option for Dropbox

petriichuk:
Adding the possibility to overwrite existing files at Dropbox instead of allowing Django change it's name.

### DIFF
--- a/docs/backends/dropbox.rst
+++ b/docs/backends/dropbox.rst
@@ -26,13 +26,9 @@ To use DropBoxStorage set::
       Timeout in seconds for making requests to the API. If ``None``, the client will wait forever.
       The default is ``100`` seconds which is the current default in the official SDK.
 
-``DROPBOX_FILE_OVERWRITE`` (optional)
-      Overwrite an existing file when it has the same name as the file being uploaded.
-      Otherwise, rename it. Default is ``False``
-
 ``DROPBOX_WRITE_MODE`` (optional)
       Allow to set Dropbox WriteMode strategy.
-      Default is ``overwrite``
+      Default is ``add``. Read more at https://dropbox-sdk-python.readthedocs.io/en/latest/api/files.html#dropbox.files.CommitInfo.mode
 
 .. _`tutorial`: https://www.dropbox.com/developers/documentation/python#tutorial
 .. _`Dropbox SDK for Python`: https://www.dropbox.com/developers/documentation/python#tutorial
--- a/docs/backends/dropbox.rst
+++ b/docs/backends/dropbox.rst
@@ -26,5 +26,13 @@ To use DropBoxStorage set::
       Timeout in seconds for making requests to the API. If ``None``, the client will wait forever.
       The default is ``100`` seconds which is the current default in the official SDK.
 
+``DROPBOX_FILE_OVERWRITE`` (optional)
+      Overwrite an existing file when it has the same name as the file being uploaded.
+      Otherwise, rename it. Default is ``False``
+
+``DROPBOX_WRITE_MODE`` (optional)
+      Allow to set Dropbox WriteMode strategy.
+      Default is ``overwrite``
+
 .. _`tutorial`: https://www.dropbox.com/developers/documentation/python#tutorial
 .. _`Dropbox SDK for Python`: https://www.dropbox.com/developers/documentation/python#tutorial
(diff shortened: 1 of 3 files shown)


storages/backends/dropbox.py
```diff
@@ -89,6 +93,9 @@ def delete(self, name):
         self.client.files_delete(self._full_path(name))
 
     def exists(self, name):
+        if self.overwrite:
```
jschneier, line 96: This is wrong.

storages/backends/dropbox.py
```diff
@@ -132,7 +139,7 @@ def _open(self, name, mode='rb'):
     def _save(self, name, content):
         content.open()
         if content.size <= self.CHUNK_SIZE:
-            self.client.files_upload(content.read(), self._full_path(name))
+            self.client.files_upload(content.read(), self._full_path(name), mode=WriteMode('overwrite'))
```
jschneier, line 142: The mode isn't constant. It needs to depend on what the user wants.

petriichuk: django checks for exists, if it's true - it renames, so the mode does not actually metter

petriichuk: @jschneier so, can I then add WriteMode as an option?

storages/backends/dropbox.py
petriichuk: so how then to make Django to not rename? any suggestions?

jschneier: Look at the other backends. They override `get_available_name`.

petriichuk: updated

storages/backends/dropbox.py
petriichuk: @jschneier updated, made as an option

storages/backends/dropbox.py
```diff
@@ -158,3 +166,24 @@ def _chunked_upload(self, content, dest_path):
                     content.read(self.CHUNK_SIZE), cursor
                 )
                 cursor.offset = content.tell()
+
+    def _clean_name(self, name):
```
jschneier, line 170: I think you can just use `self._full_path` instead here since that seems to work.

docs/backends/dropbox.rst
```diff
@@ -26,5 +26,13 @@ To use DropBoxStorage set::
       Timeout in seconds for making requests to the API. If ``None``, the client will wait forever.
       The default is ``100`` seconds which is the current default in the official SDK.
 
+``DROPBOX_FILE_OVERWRITE`` (optional)
```
jschneier, line 29: No reason to include this. It's the same as setting `DROPBOX_WRITE_MODE = 'overwrite'`.

docs/backends/dropbox.rst
```diff
@@ -26,5 +26,13 @@ To use DropBoxStorage set::
...
+      Overwrite an existing file when it has the same name as the file being uploaded.
+      Otherwise, rename it. Default is ``False``
+
+``DROPBOX_WRITE_MODE`` (optional)
```
jschneier, line 33: Can you default this to whatever the current default is (not overwrite). Don't want to break backwards compat. Can you also link out to the Dropbox docs for this property.

storages/backends/dropbox.py
```diff
@@ -158,3 +166,24 @@ def _chunked_upload(self, content, dest_path):
...
+        # a workaround here.
+        if name.endswith('/') and not clean_name.endswith('/'):
+            # Add a trailing slash as it was stripped.
+            clean_name += '/'
+        return clean_name
+
+    def get_available_name(self, name, max_length=None):
+        """Overwrite existing file with the same name."""
+        name = self._clean_name(name)
+        if self.file_overwrite:
```
jschneier, line 187: Here you can just check if `self.write_mode == 'overwrite'`.

petriichuk: updated

storages/backends/dropbox.py
petriichuk: done

docs/backends/dropbox.rst
petriichuk: changed to 'add'

storages/backends/dropbox.py
petriichuk: @jschneier can you review please ?

jschneier:
Looks great.

petriichuk:
> Looks great.
> 
> @petriichuk did you test it?

of course, I already use it.

Issue #879, opened 2020-04-28
S3 provider SSE-C not working since parameters (key) is not passed to Boto3


greenstatic:
When we enable SSE-C for the S3 provider using the following settings:
```python3
AWS_S3_OBJECT_PARAMETERS = {
    "SSECustomerAlgorithm": "AES256",
    "SSECustomerKey": "<redacted key>",
    "SSECustomerKeyMD5": "<redacted md5 key>"
}
```

We can successfully save a file to S3 (Minio) but when we try to retrieve it, we get a generic error.
The file however is normally accessible using the aws-cli [when supplied with the correct encryption keys](https://docs.minio.io/docs/how-to-use-minio-s-server-side-encryption-with-aws-cli).
Meaning django-storages handles saving files using encryption just fine, but doesn't read them correctly.

I dived into the code and found 2 issues when reading files from S3.
1. The first HEAD request (https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L74) does not contain the necessary parameters (i.e. `AWS_S3_OBJECT_PARAMETERS` that contain the necessary headers). 
Adding the encryption keys to the HEAD request is [required as per AWS S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html):
> If you encrypt an object by using server-side encryption with customer-provided encryption keys (SSE-C) when you store the object in Amazon S3, then when you retrieve the metadata from the object, you must use the following headers: 
> * x-amz-server-side​-encryption​-customer-algorithm
> * x-amz-server-side​-encryption​-customer-key
> * x-amz-server-side​-encryption​-customer-key-MD5

2. Once 1. is fixed, we still get an error because when we fetch the contents of the file (https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L100), we do so again without passing params (`AWS_S3_OBJECT_PARAMETERS`) to boto3

PR #880, opened 2020-04-28
Fix error on S3 provider file retrieval when SSE-C is enabled

greenstatic:
This is a simple recommendation what a fix for #879 can be. 
This works for my scenario but further testing/code review is necessary since I am not too familiar with the project to see if this breaks any other functionality.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -71,7 +71,7 @@ class S3Boto3StorageFile(File):
         self.obj = storage.bucket.Object(storage._encode_name(name))
         if 'w' not in mode:
             # Force early RAII-style exception if object does not exist
-            self.obj.load()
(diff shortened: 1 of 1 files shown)


jschneier:
This fixes your issue?

greenstatic:
Correct this fixes the issue #879. This can be tested manually by setting up Minio and adding the `AWS_S3_OBJECT_PARAMETERS` mentioned in the issue. Then uploading a file (you can see in Minio that it is encrypted) and requesting it in a view to display the contents.

greenstatic:
Any updates on this?

jschneier:
Hi. My hesitation with this is that I'm not sure (haven't tried) the impact of sending all parameters in these situations. People set many object parameters while using this library.

jschneier:
@greenstatic not sure why I didn't take this last year. If you rebase, I will merge.

PR #885, opened 2020-05-29
Sign CloudFront urls only in case querystring_auth is enabled

Jihadik:
## Summary
At the moment, all URLs are being signed, even if `querystring_auth` is set to `False`.
This is needed in a scenario when a user have a mix of private and public content distributed via CloudFront and they don't want to sign public urls.

## Test Plan
Updated tests to cover the scenario.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -740,7 +740,7 @@ class S3Boto3Storage(BaseStorage):
             url = "{}//{}/{}".format(
                 self.url_protocol, self.custom_domain, filepath_to_uri(name))
 
-            if self.cloudfront_signer:
+            if self.querystring_auth and self.cloudfront_signer:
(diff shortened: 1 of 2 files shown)


Jihadik:
@terencehonles, could you take a look at the PR?

terencehonles:
@Jihadik I don't personally have a mix of private and public objects that use the same storage object. For my use case I have another storage object that does not have a cloudfront signer configured. For your PR how are you turning off and on the cloudfront signing? Is there a reason you don't just not configure the signer? If you still feel strongly that the signing should be turned on/off by a boolean setting I'm not sure it should be re-using the `querystring_auth` property.

Jihadik:
@terencehonles, it's exactly the same logic as it was before: https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L755-L757 So I think it's quite consistent.

PR #888, opened 2020-06-03
Update amazon-S3.rst

lahnstrom:
After having trouble accessing images uploaded to region eu-north-1. I found out that the old signature was used, even though the documentation states that the new one is used by default. That the default signature version is still v2 is clarified in the following issue: https://github.com/boto/boto3/issues/2417

Please feel free to change the wording or decline the PR if this change is not appropriate.


jschneier:
Thanks.

PR #892, opened 2020-06-06
Introduce black for automatic code formatting

jdufresne:
From https://github.com/psf/black

> Black is the uncompromising Python code formatter. By using it, you
> agree to cede control over minutiae of hand-formatting. In return,
> Black gives you speed, determinism, and freedom from pycodestyle
> nagging about formatting. You will save time and mental energy for
> more important matters.

Using the community de-facto standard formatter makes it easier for
contributors as it removes all guesswork from the project's preferred
code style.

The flake8 and isort configurations have been updated to be compatible.


jschneier:
I'm fine introducing `black`. Can you add some language about how to work with it locally? This is going to end up being a pain for contributors (and me).

Also, has all updates needed for `travis` been applied?

jdufresne:
> Can you add some language about how to work with it locally? 

I have updated the contributing section of the README with this information. Let me know if you think more should be added or reworded. I'm happy to continue working that section until you think it looks good.

> This is going to end up being a pain for contributors (and me).

The hope is that -- in the long run -- this will make contributing easier as the formatting will be handled automatically by the tools and neither the contributor nor the reviewer will need to think about it. I hope that ends up being the case.

> Also, has all updates needed for travis been applied?

Travis already runs the `flake8` tox environment which now includes balck, so I don't believe it requires any changes, let me know if you're looking for something else here, though.

jdufresne:
Rebased to resolve conflicts.

jdufresne:
Simplified isort configuration using `profile = black`.

PR #896, opened 2020-06-11
Fixed typo in S3Boto3StorageFile.__init__()

LincolnPuzey:
Fixed typo in `S3Boto3StorageFile.__init__()` that prevented files opened in "r" mode from being read.

lambda has a higher precedence than if-else, so `self._force_mode` was set to a function that returned
another lambda when `'b'` is not in `mode`.
Also added regression test.


LincolnPuzey:
If someone can suggest how to improve the unit test so some actual content is returned by the `.read()` call, I can update the PR.

PR #898, opened 2020-06-14
[azure] Add method to set ContentDisposition

drnextgis:
So far we couldn't do that.

### DIFF
--- a/docs/backends/azure.rst
+++ b/docs/backends/azure.rst
@@ -165,3 +165,8 @@ The following settings are available:
 
     A variable to set the Cache-Control HTTP response header. E.g. 
     ``AZURE_CACHE_CONTROL = "public,max-age=31536000,immutable"``
+
+``AZURE_OBJECT_PARAMETERS``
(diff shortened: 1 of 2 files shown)


jschneier:
Can you add something similar to [get_object_parameters](https://github.com/jschneier/django-storages/blob/1c6004cc68d8160ba51df1513810ef80251ad4c4/storages/backends/s3boto3.py#L502). Probably should have been a little smarter about the name so it would be consistent across backends...you could just call it `get_object_parameters`.

drnextgis:
@jschneier thanks for the prompt reply. Done and rebased.

drnextgis:
Hey, should I do something else to have it being merged?

JeffreyCA:
Thank you for this! For future reference, the accepted key values for `AZURE_OBJECT_PARAMETERS` are: `content_type`, `content_encoding`, `content_language`, `content_disposition`, `cache_control`, and `content_md5`.

PR #899, opened 2020-06-18
Handle empty S3 folders

francoisfreitag:
Users can create empty folders on S3, by creating an object which name
ends with a forward slash, for example `folder/`. Using the `listdir`
operation on an empty folder returns `[], ["."]`. This is unexpected,
there are no file named `.` in the folder.

The reason for its presence is AWS S3 returning the following content
when listing the content of the directory:
```
{
    'Contents': [
        {... 'Key': 'folder/', ...}
    ],
   ...
}
```
Then, `posixpath.relpath("folder/", "folder/")` returns ".".

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,6 +45,7 @@ By order of apparition, thanks:
     * Taras Petriichuk (Dropbox write_mode option)
     * Zoe Liao (S3 docs)
     * Jonathan Ehwald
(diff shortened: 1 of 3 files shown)


jschneier:
Merged in 5a82c01eb51892acdd1f8125bfd3031b1acf64e4

francoisfreitag:
Thanks!

PR #906, opened 2020-07-16
Travis: test on Ubuntu 18 instead of 16

nikolas:


nikolas:
The travis failure here is only failing on flake8, which master is also failing on.

jdufresne:
> The travis failure here is only failing on flake8, which master is also failing on.

Fixed in #908

PR #907, opened 2020-07-17
Update amazon-S3.rst

amyy28:
Since collectstatic is a part of the command, it should be put inside the code block.


jschneier:
Thanks! Pushed as 353490e843721dfef70135081439c04b75196e41.

Issue #909, opened 2020-07-24
gcloud backend and App Engine Standard: credentials config problem for buckets with uniform access


przem8k:
I recently integrated django-storages to store user-uploaded files in my personal project on App Engine Standard.

Here's what **does work** perfectly: if the bucket used to store files has **fine-grained access policy** and I set `GS_DEFAULT_ACL = 'publicRead'`,everything "just works". In particular, I don't seem to need to set `GOOGLE_APPLICATION_CREDENTIALS` – the credentials present in the App Engine environment seem to be enough to authenticate the requests.

However, this does **not** seem to work if the bucket has **uniform access policy** (which could make sense to use, I want all my files in this bucket to be world-readable):

 - if with uniform access policy we set `GS_DEFAULT_ACL = ‘publicRead’`, the requests to GCS fail with 4xx, because we're trying to pass an ACL in the request (while the bucket is configured for uniform access policy)
 - if with uniform access policy we set `GS_DEFAULT_ACL = None`, then django-storages seem to assume that we need to use local credentials to sign the URL and fails with the error below

```
File "/env/lib/python3.7/site-packages/google/cloud/storage/_signing.py", line 55, in ensure_signed_credentials
"details.".format(type(credentials), SERVICE_ACCOUNT_URL)
AttributeError: you need a private key to sign credentials.the credentials you are currently using <class 'google.auth.compute_engine.credentials.Credentials'>
```

It seems that to make it possible to use buckets with uniform access policy w/o local credentials (using the ones present in the App Engine environment) it should be possible to configure django-storages so that it doesn't try to sign anything while making the request (like with `GS_DEFAULT_ACL = ‘publicRead’`) but **also** doesn't pass the ACL when making the GCS request.

**in the meantime, the workaround I found to work is just to use a bucket with fine-grained access control** – I wrote some more notes on this @ https://pnote.eu/notes/django-app-engine-user-uploaded-files/ .

sakimyto:
I had exactly the same problem. Thank you.

jschneier:
Hello, I believe #952, which was just merged, fixes this. Please confirm.

sakimyto:
> Hello, I believe #952, which was just merged, fixes this. Please confirm.

Looks great. Thanks.

Issue #783, opened 2019-11-01
Setting GS_DEFAULT_ACL to 'publicRead' returns a 400 Bad Request


ryanovas:
So I'm just trying to get my urls to stop expiring. My bucket on Google is set to public read, and if I remove the query parameters from the url of an expired link then it loads fine. I read the documentation that said to set GS_DEFAULT_ACL = 'publicRead' to return urls without the expiry. So I tried this and now I get 400 errors from the google library whenever I try to upload a file. Here is my code, followed by the error I get:

**This is my settings.py**
```
# Django Storages Config to use Google Cloud
from google.oauth2 import service_account

GS_BUCKET_NAME = 'bucket-name'
DEFAULT_FILE_STORAGE = 'storages.backends.gcloud.GoogleCloudStorage'
GS_PROJECT_ID = 'project-id'
GS_DEFAULT_ACL = 'publicRead'
GS_CREDENTIALS = service_account.Credentials.from_service_account_file(
    os.path.join(BASE_DIR, 'assets/auth/google-storage.json')
)
```
**Here is the model in question (though all file uploads fail not just this one):**

```
def path_and_rename_rdoc(instance, filename):
    ext = filename.split('.')[-1]
    filename = '{}.{}'.format(uuid4().hex, ext)
    return os.path.join('document_images', filename)

def validate_file_extension(value):
    ext = os.path.splitext(value.name)[1]
    valid_extensions = ['.jpg','.png','.svg']
    if not ext in valid_extensions:
        raise ValidationError('File not supported!')

class RequiredDocument(models.Model):
    image = models.FileField(upload_to=path_and_rename_rdoc, null=False, validators=[validate_file_extension])
```
**and here's the error:**

`google.api_core.exceptions.BadRequest: 400 POST https://www.googleapis.com/upload/storage/v1/b/bucket-name/o?uploadType=multipart&predefinedAcl=publicRead: ('Request failed with status code', 400, 'Expected one of', <HTTPStatus.OK: 200>)
`

I must be doing something dumb and not realizing it, as it works fine if I remove GS_DEFAULT_ACL. It just has expiring links even though the bucket is public. I'm using version 1.7.2, and my service account has Storage Admin permissions.

Thanks for the help!

podstava:
Same issue pointed out today. Was trying to figure out how to speed up image downloading - setting default_acl='publicRead' works, but now i can`t upload images

podstava:
[...]

PR #910, opened 2020-07-26
Fix support for Google Cloud buckets with uniform access control

dulacp:
To support Uniform permissions buckets on Google Cloud Storage, we need to keep `GS_DEFAULT_ACL` to `None`, but it forces each url to be signed, which is useless since the uniform permission is usually meant to give world read access. This new parameter solves this use case reported in #783, #846 and #909.

Cheers

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -127,7 +127,15 @@ a signed (expiring) url.
 
 .. note::
     When using this setting, make sure you have ``fine-grained`` access control enabled on your bucket, 
-    as opposed to ``Uniform`` access control, or else, file  uploads will return with HTTP 400.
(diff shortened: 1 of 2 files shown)


mands:
Bumping this PR - anything needed to do to get it merged? Would really help people using GCS buckets with uniform access policies.

mands:
Have updated and recreated the PR in #952

jschneier:
I merged in #952, thanks for the initial patch.

MeRuslan:
Hello!
I've got a bucket with uniform access at my hands, now django uploads the file, but fails to set the model field value.
I.e. the image gets uploaded, but ImageField value remains None.

MeRuslan:
i have opened a new issue: https://github.com/jschneier/django-storages/issues/962

Issue #708, opened 2019-05-22
ContentFile with str content throws TypeError


radimsuckr:
Hello, I've run into problems when I tried saving Django `ContentFile`. When I save the file with, for example, `'hello'`, it throws following exception: `TypeError: Unicode-objects must be encoded before hashing`.

However everything works fine with `bytes`. Below is traceback from Django shell. I've tried it on Django 1.11 and 2.2 with latest versions of both boto3 and django-storages.

Traceback:
```
>>> S3Boto3Storage().save('test.txt', ContentFile('hello'))
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/django/core/files/storage.py", line 52, in save
    return self._save(name, content)
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/storages/backends/s3boto3.py", line 506, in _save
    self._save_content(obj, content, parameters=parameters)
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/storages/backends/s3boto3.py", line 521, in _save_content
    obj.upload_fileobj(content, ExtraArgs=put_parameters)
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/boto3/s3/inject.py", line 621, in object_upload_fileobj
    ExtraArgs=ExtraArgs, Callback=Callback, Config=Config)
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/boto3/s3/inject.py", line 539, in upload_fileobj
    return future.result()
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/s3transfer/futures.py", line 106, in result
    return self._coordinator.result()
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/s3transfer/futures.py", line 265, in result
    raise self._exception
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/s3transfer/tasks.py", line 126, in __call__
    return self._execute_main(kwargs)
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/s3transfer/tasks.py", line 150, in _execute_main
    return_value = self._main(**kwargs)
  File "/home/grelek/projects/djstorages-test/venv/lib/python3.7/site-packages/s3transfer/upload.py", line 692, in _main
[...]

Issue #843, opened 2020-02-24
Unicode-objects must be encoded before hashing - update from 1.8 to 1.9.1


IrinaAlexandra:
So we've recently updated from version to 1.8 to version 1.9.1. Now we have a `TypeError: Unicode-objects must be encoded before hashing` when saving to `s3`. 

Code: 
```
file = ContentFile("")
document.save_csv_file(file) #writes to csv file using csv library
example_django_model.file.save(file_name, file, save=False)
```

I then proceed to receive the same error as in #708 . Even though that issue was closed a long time ago (with the solution of doing ContentFile(b'')), the code above was working fine in version 1.8. What has changed in 1.9.1 that is now causing this encoding error?

LincolnPuzey:
Did you also update your botocore/boto3 version?

On 2018-11-30 `botocore` published [version 1.12.57](https://github.com/boto/botocore/blob/develop/CHANGELOG.rst#11257) with the change "bugfix:s3: Add md5 header injection to new operations that require it"

I haven't looked at the actual code that changed there but I suspect that botocore is now calculating md5 hashes of uploads.

mertcangokgoz:
any Update?

jschneier:
Fixed by #911

PR #911, opened 2020-07-27
Fix Issue #708: Save files with string content

LincolnPuzey:
Wraps file objects open in string mode so they produce bytes and can be passed to `obj.upload_fileobj()`


LincolnPuzey:
This fixes #708, #843.

@jschneier Can I get a review on this?

You can see my new tests that reproduce the issue, fail in the build on commit c3d00d9.

The latest build is just flake8 failing which is also happening on master.

storages/backends/s3boto3.py
```diff
@@ -440,6 +440,10 @@ def _save(self, name, content):
         name = self._normalize_name(cleaned_name)
         params = self._get_write_parameters(name, content)
 
+        # wrap content so read() always returns bytes. This is required for passing it
+        # to obj.upload_fileobj() or self._compress_content()
+        content = ReadBytesWrapper(content)
```
LincolnPuzey, line 445: This is the core part of the fix. This approach has the benefits of not making any extra read() or seek() calls to the file object.

storages/utils.py
```diff
@@ -117,3 +117,30 @@ def get_available_overwrite_name(name, max_length):
             'allows sufficient "max_length".' % name
         )
     return os.path.join(dir_name, "{}{}".format(file_root, file_ext))
+
+
+class ReadBytesWrapper:
+    """
+    A wrapper for a file-like object, that makes read() always returns bytes.
+    """
+    def __init__(self, file, encoding="utf-8"):
```
LincolnPuzey, line 126: Currently any string files will be encoded to utf-8 when saved.

I'm not sure about this. It could try to look up the `file.encoding` and use that.

LincolnPuzey: I think that's a good idea given how the `encoding` property is documented here https://docs.python.org/3/library/io.html#io.TextIOBase.encoding

LincolnPuzey:
@jschneier Can I get this reviewed? Most of the diff is in tests. The new tests in `S3Boto3StorageTestsWithMoto` reproduce the original error and show its fixed.

LincolnPuzey:
@jschneier Still waiting for a review

LincolnPuzey:
Also fixes #961

wiktr:
@LincolnPuzey I see there are conflicts, could you please resolve them?
@jschneier please take a look at your earliest convenience. I think it's worth resolving this and related tickets.

I hit the described issue using versions:

```
boto3==1.16.56
botocore==1.19.56
django-storages==1.11.1
```

Would be lovely if django storages worked with s3 and boto3 out of the box, without workarounds.

LincolnPuzey:
I'm happy to resolve the conflicts if @jschneier or another maintainer can indicate they're willing to review and merge the PR.

pauloxnet:
Thanks for the PR , it fix a problem we have in our projects.
I can image all maintainers are busy, but I hope one of them can review this PR and then merge.

pauloxnet (approved):

dvf:
Any news here folks? The docs use an example that breaks too:

image

mertcangokgoz:
Any update?

hugo-reveni:
@LincolnPuzey here is another one interested in this topic, as we're having problems with text files uploaded from Admin using django-import-export v3.2.0

Any merge prevision?

jschneier:
Thanks, `moto` is a big improvement. Seems boto has grown several more features in the last few years that necessitated a few more tweaks.

hugo-reveni:
Thanks a lot @jschneier, we will wait for the next release with this improvement 👏🏽

jschneier:
No need to wait; I released it already.

hugo-reveni:
@jschneier sorry I didn't find new releases in Github tags section like past ones but now I see it was in [here](https://pypi.org/project/django-storages/). Thanks!

jschneier:
Ah I forgot to apply the tag, fixed. Thanks.

dvf:
It's people like @jschneier that make OSS awesome. Thanks Josh ❤️

PR #908, opened 2020-07-21
Update to be compatible with isort 5.* series

jdufresne:
* not_skip is deprecated (files no longer skipped
* --recursive is deprecated (now the default)

PR #912, opened 2020-07-27
Pin isort and flake8 dependencies to major versions.

LincolnPuzey:
isort5 contains some breaking changes causing tests to fail on master. This is a quick fix, it should be upgraded to 5 at some point.


jdufresne:
The upgrade is handled in #908. IMO, we should apply the upgrade rather than pinning to an old version.

Issue #895, opened 2020-06-07
Support Temporary Profile Credentials


rotherfuchs:
Assuming `AWS_S3_SESSION_PROFILE` has been set in `settings.py` we could add this to `s3boto.py`, line `362+`:

```
if settings('AWS_S3_SESSION_PROFILE'):
    session = boto3.Session(profile_name=settings('AWS_S3_SESSION_PROFILE'))
    self._connections.connection = session.resource(
        's3',
        aws_access_key_id=session.get_credentials().access_key,
        aws_secret_access_key=session.get_credentials().secret_key,
        aws_session_token=session.get_credentials().token,
        region_name=self.region_name,
        use_ssl=self.use_ssl,
        endpoint_url=self.endpoint_url,
        config=self.config,
        verify=self.verify,
    )
```

jschneier:
Fixed by #919

Issue #774, opened 2019-10-09
Add AWS_PROFILE settings


witold-gren:
If you work in a few projects better way (locally) is usage AWS profiles (`IAM Roles`) configured in awscli. You usage one credentials account (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`) and much more AWS profiles.

witold-gren:
Actually it's a better solution is usage `sts` client to start usage IAM Role without configuring `awscli`, We must only set two extra params `AWS_ROLE_ARN ` and `AWS_ROLE_SESSION_NAME `:
```
client = boto3.client('sts')
response = client.assume_role(
    RoleArn= AWS_ROLE_ARN, 
    RoleSessionName= AWS_ROLE_SESSION_NAME)
session = boto3.Session(
    aws_access_key_id=response['Credentials']['AccessKeyId'],
    aws_secret_access_key=response['Credentials']['SecretAccessKey'],
    aws_session_token=response['Credentials']['SessionToken'])
```

jschneier:
Fixed by #919

PR #919, opened 2020-08-25
Allow the use of AWS profiles for S3 access

dan-hook:
This uses the setting AWS_S3_SESSION_PROFILE as an alternative to using static credentials in AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.  
This addresses the spirit of https://github.com/jschneier/django-storages/issues/774 and implements the suggestion in https://github.com/jschneier/django-storages/issues/895

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,6 +45,7 @@ By order of apparition, thanks:
     * Taras Petriichuk (Dropbox write_mode option)
     * Zoe Liao (S3 docs)
     * Jonathan Ehwald
+    * Dan Hook
 
 
 Extra thanks to Marty for adding this in Django,
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -32,7 +32,14 @@ To allow ``django-admin.py`` collectstatic to automatically put your static file
 
 .. note::
 
-      If ``AWS_ACCESS_KEY_ID`` and ``AWS_SECRET_ACCESS_KEY`` are not set, boto3 internally looks up IAM credentials.
+      If ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_S3_SESSION_PROFILE`` are not set, boto3 internally looks up IAM credentials.
+
+``AWS_S3_SESSION_PROFILE``
+    The AWS profile to use instead of ``AWS_ACCESS_KEY_ID`` and ``AWS_SECRET_ACCESS_KEY``
+
+.. note::
+      If this is set, then ``AWS_ACCESS_KEY_ID`` and ``AWS_SECRET_ACCESS_KEY`` are ignored
+
 
 ``AWS_STORAGE_BUCKET_NAME``
     Your Amazon Web Services storage bucket name, as a string.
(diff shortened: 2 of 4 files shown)


jschneier:
This looks close. If you [look at the underlying source](https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/session.html#Session.resource) you can see that I don't think we need to branch on whether or not profile name exists. You can just always pass it in and if it's `None` it won't matter. Similar for the `AWS_ACCESS_KEY_ID` and other parameters.

Just need to have the docs be very particular about what is used.

dan-hook:
You're right that it isn't necessary to branch on the existence of the profile name, but that changes what the code does in the scenario where AWS_S3_SESSION_PROFILE is set and AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY are set.  

Here's the code I think you're proposing:
```
session = boto3.Session(profile_name=setting('AWS_S3_SESSION_PROFILE'))
self._connections.connection = session.resource(
         's3',
         aws_access_key_id=self.access_key,
         aws_secret_access_key=self.secret_key,
         aws_session_token=self.security_token,
         region_name=self.region_name,
         use_ssl=self.use_ssl,
         endpoint_url=self.endpoint_url,
         config=self.config,
         verify=self.verify,
     )
```
In the docs I said that if AWS_S3_SESSION_PROFILE is set, then the key variables are ignored, but using the code above reverses that.  In the short term, it seems that if someone has all of these variables set, then the intention is that AWS_S3_SESSION_PROFILE would take precedence (since it's newer).  

I think we have the following options:

1. AWS_S3_SESSION_PROFILE takes precedence over the key variables.
2. The key variables take precedence over AWS_S3_SESSION_PROFILE
3. We raise an error if AWS_S3_SESSION_PROFILE and either of the key variables are set.

I prefer 1 or 3, because with 2 we have to consider what happens when AWS_S3_SESSION_PROFILE is set and only one of the key variables is set.

jschneier:
Hmm, maybe #3 should be a Check? Raising would probably be okay too.

The authenticating to S3 docs need a bit of a cleanup with all the additional options that have been added.

dan-hook:
Eventually S3Boto3Storage falls through to looking at environment variables for the access key id and the secret key, and boto3 will prefer to use the access key given as an argument rather than the profile.  I would need to change a lot of the code in order to make sure that `self.access_key == None` when AWS_S3_SESSION_PROFILE is set.  
I've added a check to raise an error if both AWS_S3_SESSION_PROFILE and either of the AWS_S3_ key settings are present.

dan-hook:
Any additional comments?  I need to condition on the existence of the session name to make sure that keys aren't used in the 
creation of the resource.

dan-hook:
Pinging about this PR again.  I don't know how to get rid of the if self.session_profile, but if your prefer I could move the session creation into it's own private method or a property, which would at least clean up the connection method.

jschneier:
Can you move to a method and add a comment? I'm itching to DRY it and there are good reasons to avoid doing so that can only be captured via comment.

I think this looks good in general and will take it.

dan-hook:
I added a method to create a session and removed the conditional logic from the connection property.  For an unknown reason, the py37-django22 build failed.  Everything else passed, so I'm guessing that's a transient issue with travis-ci.  I added another commit with a comment.  Hopefully the build will go smoothly this time.

dan-hook:
@jschneier Pinging about this PR again.  Do you like this refactoring?

dan-hook:
Pinging again.  https://github.com/jschneier/django-storages/pull/956/ would create a conflict with this PR, so it would be nice to get this merged before then.

llostris:
@jschneier Any chance this could get a review? 🙏 (Or is there anything we can do to help push it across the line? :))

dan-hook:
@llostris Josh has been missing from github for the entire year.  I'm somewhat concerned.

dan-hook:
@jschneier Welcome back!

dan-hook:
Pinging on this again.  Anything left to do on this to get it merged?

dan-hook:
Looks like there was some activity in the repo on June 6th, so I'll ping on this again.  Any further changes needed to get this ready to merge?

szi:
This would be really useful to have for me as well. Hoping this PR will merge some point soon.

anaerobeth:
Neat feature. Looking forward to using this.

jschneier:
I like this most recent take and am happy to take it.

PR #930, opened 2020-09-08
Calculate settings for S3Boto3StorageFile when instantiated, not imported

brianhelba:
This allows the `storages.backends.s3boto3` module to be cleanly imported before Django settings are configured. This further supports the changes from #524.

Removing `S3Boto3StorageFile.buffer_size` as a class variable does not affect the API surface of `S3Boto3StorageFile`, as `buffer_size` was always accessible as an instance variable.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -75,7 +75,6 @@ else:
 
 @deconstructible
 class S3Boto3StorageFile(File):
-
     """
     The default file object used by the S3Boto3Storage backend.
 
@@ -92,7 +91,6 @@ class S3Boto3StorageFile(File):
     order to properly write the file to S3. Be sure to close the file
     in your application.
     """
-    buffer_size = setting('AWS_S3_FILE_BUFFER_SIZE', 5242880)
 
     def __init__(self, name, mode, storage, buffer_size=None):
         if 'r' in mode and 'w' in mode:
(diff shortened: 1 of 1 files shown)


brianhelba:
@jschneier @jdufresne I'd welcome your feedback, as this supports #524.

brianhelba:
Ping @jschneier @jdufresne 

Is there anything I should do to support review of this PR?

jdufresne:
The change looks good to me. Could a test be added to demonstrate the effect of the change?

brianhelba:
In the test environment, the Django settings are always implicitly configured on-demand, since the `DJANGO_SETTINGS_MODULE` environment variable is set.

The proof that this PR works is that from a clean Python shell (where `DJANGO_SETTINGS_MODULE` is *not* set),
```python
from storages.backends import s3boto3
```
raises
```
django.core.exceptions.ImproperlyConfigured: Requested setting AWS_S3_FILE_BUFFER_SIZE, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```
before this change and successfully imports after this change.

I *could* add an additional Tox test environment where `DJANGO_SETTINGS_MODULE` is not set, and where the test code attempts to import all of the the modules within `storages`. However, that seems like a lot of additional test infrastructure to maintain. On the other hand, if everyone agrees that being able to cleanly import is an important feature (since it is a major goal of #524), maybe having a test to prevent a regression is worth it.

I'd be happy to write the extra testing if the project maintainers want it.

jdufresne:
The following test demonstrates the change, IMO. It fails before the fix and passes aftewards.

```py
    def test_override_file_buffer_size(self):
        f = s3boto3.S3Boto3StorageFile("foo.txt", "rb", mock.MagicMock())
        self.assertEqual(f.buffer_size, 5242880)
        with override_settings(AWS_S3_FILE_BUFFER_SIZE=1024):
            f = s3boto3.S3Boto3StorageFile("foo.txt", "rb", mock.MagicMock())
            self.assertEqual(f.buffer_size, 1024)
```

PR #938, opened 2020-09-23
Check for `NoSuchKey` error response in `exists()`

deadbeef404:
I'm trying to determine why I'm seeing some files overwriting other files when they have the same name.

I feel like this might be the source of the issue (perhaps due to previous code catching a network error or some other error) but I don't quite have the depth of knowledge to validate that.

Does this suggested change look like it might address a bug?


jschneier:
Shouldn't it be `NoSuchKey`? https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

jschneier (approved):

PR #949, opened 2020-10-29
docs: Add missing import for S3 backend

chaws:
Documentation on S3 is missing `from django.core.files.base import ContentFile` import here:

Screenshot from 2020-10-29 12-01-08


jschneier:
Thanks.

PR #952, opened 2020-11-10
Add setting to disable signed URLs for GCS uniform buckets

mands:
Rebased version of existing PR by @dulacp in #910 that fixes merge conflict and should pass CI tests.

As per docs in #910 
>  To support Uniform permissions buckets on Google Cloud Storage, we need to keep GS_DEFAULT_ACL to None, but it forces each url to be signed, which is useless since the uniform permission is usually meant to give world read access. This new parameter solves this use case reported in #783, #846 and #909.


jschneier:
Looks great. Thanks.

PR #970, opened 2020-12-21
[gcloud] Add configurable blob parameters

### LINKED ISSUE #510: No way to upload files with headers using Google Cloud Storage
I'm using Google Cloud Storage and currently have to either manually or script the setting of my Content-Disposition headers. The amazon module supports this but GCS does not?
I believe it should be implemented in a similar way, 
```python
GS_HEADERS = {
    'Expires': 'Thu, 15 Apr 2010 20:00:00 GMT',
    'Cache-Control': 'max-age=86400',
    'Content-Disposition': 'attachment',
}
```

### LINKED ISSUE #663: specify additional metadata for google storage
I'd really like to be able to set arbitrary metadata/headers when new files get uploaded to google storage. Ideally, i'd have arbitrary keyword args on my `FileField` so I could, for example, set the `content-type` and `content-disposition` fields during the upload. It's not clear to me if that's supported by the google api, though. 

This one seems related, https://github.com/jschneier/django-storages/issues/231

jschneier:
Resolves #663 
Resolves #510

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -38,24 +38,29 @@ Alternatively, you can use the setting `GS_CREDENTIALS` as described below.
 
 Getting Started
 ---------------
-Set the default storage and bucket name in your settings.py file:
+Set the default storage and bucket name in your settings.py file
 
-::
+.. code-block:: python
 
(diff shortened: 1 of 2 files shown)


denizdogan:
Is this PR ready to go? Looking forward to overriding `content_type`.

pokidovea:
Hi! What about a new version with these changes?

PR #980, opened 2021-02-08
Add gzip support Google Cloud Storage

jkevingutierrez:
Add CompressMixins that can be reused across all storages.

Add the Gzip logic needed for Cloud Storage


tests/test_gcloud.py
```diff
@@ -10,6 +11,7 @@
 from google.cloud.storage.blob import Blob
 
 from storages.backends import gcloud
+from tests.test_s3boto3 import NonSeekableContentFile
```
jschneier, line 14: Can you create a `tests/util.py` and import from that. I don't want to cross-import the various backend's test files into each other.

storages/compress.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class CompressStorageMixin():
+    def _compress_content(self, content):
+        """Gzip a given string content."""
+        return GzipCompressionWrapper(content)
+
+
+class CompressFileMixin():
+    def _compress_file(self):
```
jschneier, line 13: IMO this should take in `mode` and `file`.

storages/backends/gcloud.py
```diff
@@ -109,6 +112,14 @@ def get_default_settings(self):
             "default_acl": setting('GS_DEFAULT_ACL'),
             "querystring_auth": setting('GS_QUERYSTRING_AUTH', True),
             "expiration": setting('GS_EXPIRATION', timedelta(seconds=86400)),
+            "gzip": setting('GS_IS_GZIPPED', False),
+            "gzip_content_types": setting('GZIP_CONTENT_TYPES', (
+                'text/css',
+                'text/javascript',
+                'application/javascript',
+                'application/x-javascript',
+                'image/svg+xml',
```
jschneier, line 121: `wasm` is in the docs but missing from the list.

storages/backends/gcloud.py
```diff
@@ -56,6 +57,8 @@ def _get_file(self):
                 self._is_dirty = False
                 self.blob.download_to_file(self._file)
                 self._file.seek(0)
+            if self._storage.gzip and self.blob.content_encoding == 'gzip':
+                self._file = self._compress_file()
```
jschneier, line 61: Doesn't this actually decompress rather than compress? If the content_encoding is gzip already and we are getting the file and all.

This might be buggy actually, not sure. Why would `self._storage.gzip` matter in this case.

storages/backends/gcloud.py
```diff
@@ -76,7 +79,7 @@ def write(self, content):
         if 'w' not in self._mode:
             raise AttributeError("File was not opened in write mode.")
         self._is_dirty = True
-        return super().write(force_bytes(content))
+        return super().write(to_bytes(content))
```
jschneier, line 82: 👍

storages/backends/gcloud.py
```diff
@@ -168,10 +179,15 @@ def _save(self, name, content):
...
+        if self.gzip and upload_params.get(CONTENT_TYPE) in self.gzip_content_types:
+            content = self._compress_content(content)
+            file_object.blob.content_encoding = 'gzip'
+            upload_params[CONTENT_ENCODING] = 'gzip'
+
         for prop, val in blob_params.items():
             setattr(file_object.blob, prop, val)
 
-        file_object.blob.upload_from_file(content, rewind=True, size=content.size, **upload_params)
+        file_object.blob.upload_from_file(content, rewind=True, size=getattr(content, 'size', None), **upload_params)
```
jschneier, line 190: Why is this change necessary?

storages/backends/gcloud.py
```diff
@@ -168,10 +179,15 @@ def _save(self, name, content):
         if CONTENT_TYPE not in blob_params:
             upload_params[CONTENT_TYPE] = file_object.mime_type
 
+        if self.gzip and upload_params.get(CONTENT_TYPE) in self.gzip_content_types:
+            content = self._compress_content(content)
+            file_object.blob.content_encoding = 'gzip'
+            upload_params[CONTENT_ENCODING] = 'gzip'
+
```
jschneier, line 186: Can you explain how I should think about `upload_params` vs `blob_params` in the context of this set of changes.

jschneier (changes requested):
A few questions & a few concerns.

storages/backends/gcloud.py
jkevingutierrez: Good catch! As you can see in the CompressFileMixin the `_compress_file` method is actually decompressing the file. I just gave it the wrong name.

storages/backends/gcloud.py
```diff
@@ -168,10 +178,14 @@ def _save(self, name, content):
         if CONTENT_TYPE not in blob_params:
             upload_params[CONTENT_TYPE] = file_object.mime_type
 
+        if self.gzip and upload_params.get(CONTENT_TYPE) in self.gzip_content_types:
+            content = self._compress_content(content)
+            file_object.blob.content_encoding = 'gzip'
```
jschneier, line 183: Alternatively I would just write

```python
blob_params['content_encoding'] = 'gzip'
```

since it ends up being set just below.

storages/backends/gcloud.py
```diff
@@ -168,10 +178,14 @@ def _save(self, name, content):
         if CONTENT_TYPE not in blob_params:
             upload_params[CONTENT_TYPE] = file_object.mime_type
 
+        if self.gzip and upload_params.get(CONTENT_TYPE) in self.gzip_content_types:
```
jschneier, line 181: I'm still not sure about this check.

If the user chooses to put `content_type` in `get_object_parameters`, then `upload_params[CONTENT_TYPE]` would never be set so this wouldn't be hit.

Maybe what we actually need above is

```python
upload_params[CONTENT_TYPE] = blob_params.pop(CONTENT_TYPE, file_object.mime_type)
```

Then I think to mirror S3 what we want is to check

```python
if (self.gzip and 'content_encoding' not in blob_params
       and upload_params[CONTENT_TYPE] in self.gzip_content_types):
  ....
```

What do you think?

jschneier (commented):
Couple more comments.

storages/backends/gcloud.py
jkevingutierrez: Yeah, good one. Thanks!

storages/backends/gcloud.py
jkevingutierrez: Fixed. Thanks!

storages/compress.py
```diff
@@ -0,0 +1,14 @@
+from gzip import GzipFile
+
+from storages.utils import GzipCompressionWrapper
```
jschneier, line 3: Can you also move the `GzipCompressionWrapper` into this file.

jkevingutierrez: Sure, good call

jschneier (approved):

PR #990, opened 2021-03-11
Add AWS_S3_ACCESS_KEY_ID and AWS_S3_SECRET_ACCESS_KEY alternative settings to S3 docs

solastley:
I noticed that the docs don't currently mention that you can use more specific settings variable names `AWS_S3_ACCESS_KEY_ID` and `AWS_S3_SECRET_ACCESS_KEY` to configure your S3 access. This is useful in situations where you would like to use a separate access key to e.g. send emails via Amazon SES. I checked in the source code, and it seems that these variable names [are supported](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L246-L247).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,6 +45,7 @@ By order of apparition, thanks:
     * Taras Petriichuk (Dropbox write_mode option)
     * Zoe Liao (S3 docs)
     * Jonathan Ehwald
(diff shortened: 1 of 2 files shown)


jschneier:
Resolved by #919

PR #994, opened 2021-03-22
GoogleCloud: Use V4 signed URLs

martey:
Update URL signing version from v2 to v4. Increase minimum google-cloud-storage library version so that custom endpoints work
properly.

This supersedes #837 - moving to v4 signed URLs and using `bucket_bound_hostname` will fix the bug with invalid signed URLs being generated without having to resort to the hackish solution used in that pull request.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ boto3 =
 dropbox =
     dropbox >= 7.2.1
 google =
-    google-cloud-storage >= 1.15.0
+    google-cloud-storage >= 1.27.0
 libcloud =
     apache-libcloud
 sftp =
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -282,11 +282,14 @@ class GoogleCloudStorage(BaseStorage):
                 quoted_name=_quote(name, safe=b"/~"),
             )
         elif not self.custom_endpoint:
-            return blob.generate_signed_url(self.expiration)
+            return blob.generate_signed_url(
+                expiration=self.expiration, version="v4"
+            )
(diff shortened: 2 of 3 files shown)


martey:
@jschneier I would appreciate your feedback on this - this solves the bug with using signed URLs with a custom endpoint on GCS, but it isn't as hackish as my earlier attempt.

jschneier:
@martey I think the fix is good but am concerned about breaking backwards compat.

I recall when I upgraded S3 to use `s3v4` alone I had to revert because people had persisted the output of `url` to the DB. I also realize that GCloud doesn't allow URLs to hang around for longer than 7 days anymore.

What do you think?

jschneier:
Yeah okay let’s take this. Can you rebase?

jschneier:
Does this require the higher google-cloud-storage library? It’s fine, just want to document if so.

martey:
I have rebased this.

This pull request does require at least google-cloud-storage 1.27. [That version](https://github.com/googleapis/python-storage/blob/main/CHANGELOG.md#1270-2020-04-01) adds custom endpoint/hostname support for v4 signed URLs. Before that version, it wasn't possible to natively create a signed URL with a custom endpoint (which is why my previous fix for signed URLs and custom endpoints in #837 created a signed URL and then manually added the custom endpoint).

Since Google is now recommending that users use v4 signing instead of v2 signing (see the first "Important" note on https://cloud.google.com/storage/docs/access-control/signed-urls-v2), and this fixes a bug (using signed URLs and a custom endpoint fails), I personally don't see an issue with this being a breaking change.

PR #1003, opened 2021-04-20
Convert string to bytes if needed

dannysauer:
When django-storages is used from an app which is configured by env variables
fed into dynaconf (or probably other situations), it's not easy to directly
coerce a string representation of the signing key into a bytes-like object; it
comes in as a string.  This detects that situation and encodes the string.


jschneier:
Thanks, looks great. Hopefully merging kicks CI.

jschneier:
Guess we could have used Django's `force_bytes` or similar.

dannysauer:
My Django skills are very weak. :)

PR #1004, opened 2021-04-20
Handle parameters for presigned CloudFront

### LINKED ISSUE #997: parameters are ignored if s3 custom URL is provided
https://github.com/jschneier/django-storages/blob/770332b598712da27ecdba75c9e202ad6a1a8722/storages/backends/s3boto3.py#L554-L578

If additional parameters are passed in to the url method when a custom_url is in use (like when S3 is behind CloudFront), they're ignored on the generated URL.  They're already passed in as a dict, so they could easily be passed in to `urllib.parse.urlunparse` instead of using a format string when the URL is built (which might be a good fix anyway).

dannysauer:
Prior code discarded parameters on pre-signed cloudfront URLs. This code fixes that, using the botocore methods which validate parameters allowed on S3 get object calls.

Also set expiration time header/parameter for both cloudfront and regular S3 signed URLs, since that seems to make sense for cache purposes; cache expires when url expires.

Closes #997

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -572,7 +572,7 @@ class S3Boto3Storage(BaseStorage):
                 self._s3_service_model['operations']['GetObject'], sm)
         serializer = botocore.serialize.create_serializer('rest-json')
         url_dict = serializer.serialize_to_request(params, om)
-        return urlencode(url_dict['query_string'])
+        return urlencode(url_dict['query_string'], safe='/",')
 
     def url(self, name, parameters=None, expire=None, http_method=None):
         # Preserve the trailing slash after normalizing the path.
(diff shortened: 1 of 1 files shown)


dannysauer:
FWIW, this was written under a bit of duress.  It's working for me in production with signed CloudFront URLs, but I'm very open to structure / stylistic changes to make it fit in with the other code better or just suck less. :)

It's ultimately not a lot of real change - moved some existing code earlier in a subroutine, changed how the customdomain URL is constructed, and added a method to aws-ify the S3 parameters for customdomain use.  That last part is derived from what the regular "build signed url" function ultimately does for S3 (which took a *ton* of fidlding around for what boils down to just 3-4 lines of code), except that I essentially hard-coded GetObject as the operation since I have a hard time imagining people using other operations in the context of this module.

dannysauer:
force push because rebased on current master.

jschneier:
Ended up taking #1054 which I imagine addresses your problem. Please re-open if it does not.

PR #1015, opened 2021-05-20
Fix compatability with Django safe_join and Windows (leading "/" on file path)

johnvorsten:
Even though you pass the upload_to argument which gives a unix-style path (something like /save/path) the django utility makes sure the root of that save path matches the base path for the operating system :(. The dropbox SDK does not like a drive prefixed to the desired save path (AKA C:/path/file.name is not a valid save directory for dropbox).


wild-boomerang:
Are you sure that `import os` and
```
class SuspiciousFileOperation(Exception):
    pass
```
are not needed to add?
As for me, your code didn't work until I added this code.
Thanks anyway! It works for me.

storages/backends/dropbox.py
```diff
@@ -86,8 +86,26 @@ def __init__(self, oauth2_access_token=oauth2_access_token, root_path=location,
...
         if name == '/':
-            name = ''
-        return safe_join(self.root_path, name).replace('\\', '/')
+            name = ''        
+        # If the machine is windows do not append the drive letter to file path
+        if os.name == 'nt':
+            final_path = os.path.join(self.root_path, name).replace('\\', '/')
+            
+            # Separator on linux system
+            sep = '//'
```
jschneier, line 95: Isn't `sep` on Linux (Unix) just a single slash (`/`)?

PR #1024, opened 2021-06-23
Signed urls without network call

nielsole:
Lazily calling bucket.get_blob, thus avoiding the call to bucket.get_blob during url(), which should prevent blocking IO.
Fixes https://github.com/jschneier/django-storages/issues/413

Tested via `DJANGO_SETTINGS_MODULE=tests.settings pytest tests/test_gcloud.py`

I wanted to benchmark this before and after, to be sure I actually improved it, but haven't had the time to set this up so far.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,6 +45,7 @@ By order of apparition, thanks:
     * Taras Petriichuk (Dropbox write_mode option)
     * Zoe Liao (S3 docs)
     * Jonathan Ehwald
+    * Niels-Ole Kühl (Google Cloud Storage patch)
 
 
 Extra thanks to Marty for adding this in Django,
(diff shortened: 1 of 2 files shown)


nielsole:
I did a before/after performance test with ~40 signed URLs in a view and couldn't find _any_ performance difference o_O
Looks like I have some more validation to do.

sww314:
@nielsole I worked on this and ended up solving outside of `django-storages`. If you profile the the generation of the URL. We had a network call to refresh some credentials. We cached that, but then the majority of the time is actually in the crypto and generation of the key. 

We ended up caching the signed urls to improve performance. Even making 40 redis calls was much faster than generating the values.

PR #1026, opened 2021-07-08
Re-allow overrides of access_key_names

toddgardner:
This change:

b13efd9

swapped from using `self.access_key_names` to `S3Boto3Storage.access_key_names`; it's unclear if that was intentional or not, but that means it does not accept overrides; we relied on overrides to use some bucket specific keys provided by rook ceph.


jschneier:
Thanks. Not intentional.

PR #1032, opened 2021-07-28
Fixed storage location parsing and added ftps protocol

PotatoHD404:
If you pass a password that contains the `#` symbol the library crashes. So I fixed it.


jschneier:
Between #1320 & #1329 both of these fixes are now merged, thanks for this PR and sorry for not taking it at the time.

PR #1036, opened 2021-08-08
Add how to overwrite the default storage class

jacobitosuperstar:
Reading through the docs of Google Cloud Storage, there are some things missing about functionality that are able in the S3 section. So i though necessary to add them. 

The point is to know:
- How to overwrite the default storage class
- Know how to use several buckets in the same project

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -306,24 +306,26 @@ You can now use your custom storage class for default file storage in Django set
 
 Or you may want to upload files to different buckets depending on your needs::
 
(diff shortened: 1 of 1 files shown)


jacobitosuperstar:
I am working on a personal project and I'm using Google Cloud for it. There is a lack of documentation on the tools is a problem. There where a lot of things that i could only figure out when checking the source code of the project, and i though that i could be part of the solution. Hope my knowledge is good and it helps. Open to corrections and comments.

jschneier:
Docs have been updated to cover this use case.

PR #1041, opened 2021-08-12
S3: Avoid a ListParts request in multipart upload

craigds:
S3 returns the ETag for each part of the upload while we're PUTting
them. However, until now we were ignoring those responses, and instead
doing a separate ListParts (GET) request for the upload object, just
before POSTing the list of parts to the completion endpoint to finalise
the upload.

That extra GET request is unnecessary so this commit removes it.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -113,6 +113,7 @@ class S3Boto3StorageFile(File):
         self._raw_bytes_written = 0
         self._file = None
         self._multipart = None
+        self._parts = None
         # 5 MB is the minimum part size (if there is more than one part).
         # Amazon allows up to 10,000 parts.  The default supports uploads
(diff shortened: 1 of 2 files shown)


jschneier:
@craigds Thanks. I never set up hitting actual AWS during the tests so I must ask — how did you test this?

craigds:
I tested it inside a dev container which _was_ hitting s3, inserted a pdb breakpoint at the request and inspected the return value, and updated the mock to look similar.

Really the only thing that I had to verify is how the ETag for the uploaded part is included in the response ([docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html#API_UploadPart_ResponseSyntax))

jschneier:
Yep. Not saying it’s complicated, just want to hear how it was done (that it was done at all, really)

PR #1005, opened 2021-04-20
Add Django 3.2 trove classifier

jschneier:

PR #1051, opened 2021-09-06
ci: add github CI to replace travis CI

mariajgrimaldi:
**Description**

This migrates Travis CI configuration to Github actions. This approach was discussed [here](https://github.com/jschneier/django-storages/pull/1005#issuecomment-913050477) as a consequence of Travis CI changing support for open source projects.

**Concerns**

- [x] Should we keep the Travis CI configuration?
- [x] Which python version was being used for the quality checks?
- [x] On which events should we trigger testing?


mariajgrimaldi:
Hey! @jschneier 

I think Github Actions are not enabled

jschneier:
@mariajgrimaldi that isn't what I see but it's not clear to me why it isn't running.

jschneier:
@mariajgrimaldi for your questions

1. You can also ahead and delete the Travis stuff
2. For linting, it would be ideal if the minimum version of Python was used but I don't think it overly matters
3. Tests should run on PR open and PR update (push to PR). It should also run against trunk after merging.

mariajgrimaldi:
Thanks for the response @jschneier! I used python 3.6 for this restriction: https://github.com/jschneier/django-storages/blob/master/tox.ini#L55

Also, do you know why checks are not running? They do here: https://github.com/mariajgrimaldi/django-storages/pull/1/checks

Maybe this has to be enabled? 
image

jschneier:
Unclear, I think because I didn't "set it up" myself. I'm going to commit this and see what happens.

jschneier:
@mariajgrimaldi yup! Looks like it's working!

https://github.com/jschneier/django-storages/actions/runs/1209946412

mariajgrimaldi:
excellent!! @jschneier, thanks for the help. Can you make a release with these changes?

natabene:
@jschneier Just checking if you could look into doing a release with this changes some time soon. I would greatly appreciate, as it would really help our Django upgrade project for the Open edX Platform.

PR #1060, opened 2021-09-21
[gcloud] fix no_signed_url incorrectly evaluated when setting acl publicRead via object_parameters

### LINKED ISSUE #1059: [gcloud] Add configurable blob parameters - Bug
#970 did introduce the option to set `acl` in `object_parameters ` instead of `default_acl`.

But when setting `acl` to `publicRead`, the following code https://github.com/jschneier/django-storages/blob/26936b3f3894d819c558acd83bbad1bff4808ac3/storages/backends/gcloud.py#L275 isn't evaluated correctly and the url will be signed.

A workaround is to either keep using `default_acl = "publicRead"` or setting `querystring_auth = False`

vchrisb:
fixes #1059

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -272,7 +272,7 @@ class GoogleCloudStorage(BaseStorage):
         name = self._normalize_name(clean_name(name))
         blob = self.bucket.blob(name)
         no_signed_url = (
-            self.default_acl == 'publicRead' or not self.querystring_auth)
+            self.object_parameters.get('acl', self.default_acl) == 'publicRead' or not self.querystring_auth)
 
         if not self.custom_endpoint and no_signed_url:
             return blob.public_url


jschneier:
Thanks. It's also in `.close()`, sigh.

jschneier:
Probably should call `get_object_parameters` as well for correctness.

vchrisb:
I have a look. I guess it makes sense to move the handling of `cache_control` and `default_acl` to `get_object_parameters` and call `get_object_parameters` where necessary?

vchrisb:
@jschneier can you have a look at the proposed changes?

storages/backends/gcloud.py
```diff
@@ -83,7 +83,7 @@ def close(self):
             if self._is_dirty:
                 self.blob.upload_from_file(
                     self.file, rewind=True, content_type=self.mime_type,
-                    predefined_acl=self._storage.default_acl)
+                    predefined_acl=self._storage.object_parameters.get('acl', self._storage.default_acl))
```
jschneier, line 86: Should this call ‘get_object_parameters’?

storages/backends/gcloud.py
```diff
@@ -158,34 +158,39 @@ def _save(self, name, content):
...
+            warnings.warn(
+                'The GS_CACHE_CONTROL setting is deprecated. Use GS_OBJECT_PARAMETERS to set any '
+                'writable blob property or override GoogleCloudStorage.get_object_parameters to '
+                'vary the parameters per object.', DeprecationWarning
+            )
+            object_parameters['cache_control'] = self.cache_control
+        if 'acl' not in object_parameters:
+            object_parameters['acl'] = self.default_acl
+
+        return object_parameters
```
jschneier, line 193: The problem with putting this here is that people will override this method. Then the ACL won’t be set unless they call `super`.

storages/backends/gcloud.py
vchrisb: I haven't grasp how all things tie together and didn't want to make a bigger change here, especially without a testcase.
I've pushed an update.

storages/backends/gcloud.py
vchrisb: I added the `default_acl` logic to the code areas where object_parameters is called from.

PR #1061, opened 2021-09-21
Replace S3Boto3Storage._compress_content with a streaming implementation

vainu-arto:
The original version reads the entire content in one go, compresses and
places it into a buffer in memory. This limits both the possible size of
the file that can be saved and compressed, and also the performance of the
transfer.


jschneier:
This handles setting `mtime=0`, right?

vainu-arto:
Yes, mtime in the header will always be zero. In my understanding it isn't actually possible to set it to any other value when allowing zlib to generate the header itself.

vainu-arto:
I ran random sets of data through it and the stdlib GzipFile, decompressed (with GzipFile and system gzip) and compared the output between each other and the original. I couldn't find a way to make it produce output that could not be decompressed or would differ from the original.

jschneier:
Thanks

PR #1063, opened 2021-10-02
Azure: support user delegation key

dimbleby:
I just noticed that #805 had been merged!  Very happy to see it, thank you.

While that has been waiting, I have been running with a variation on the Azure storage code that adds support for [user delegation keys](https://docs.microsoft.com/en-us/rest/api/storageservices/get-user-delegation-key).  This allows us to hand out URLs after using a managed service identity to authenticate with the blob store, rather than an account key.  Once you have set up your Azure resources appropriately, the configuration looks something like this:
```python
from azure.identity import DefaultAzureCredential
AZURE_TOKEN_CREDENTIAL = DefaultAzureCredential()
AZURE_ACCOUNT_NAME = "myaccount"
AZURE_CONTAINER = "my-container"
...
```

Since getting a user delegation key requires use of the `BlobServiceClient` I've also refactored the creation of the `ContainerClient` ever so slightly to reuse that.  I must admit I had my doubts about the value of this refactoring when I came to polish this into an MR and realized that I'd have to fix up the resulting unit test breakage! but I do think it is the cleanest way.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -4,7 +4,7 @@ from tempfile import SpooledTemporaryFile
 
 from azure.core.exceptions import ResourceNotFoundError
 from azure.storage.blob import (
-    BlobClient, BlobSasPermissions, ContainerClient, ContentSettings,
+    BlobClient, BlobSasPermissions, BlobServiceClient, ContentSettings,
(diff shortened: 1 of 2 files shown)


jschneier:
Is there any need to update the documentation for this feature?

PR #1071, opened 2021-10-12
Add Azure `parameters` as kwargs for `storage.url`

fao89:
It enables using kwargs, such as `content_type`, `content_disposition`, ..., when generating a shared access signature for a blob
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/azure/storage/blob/_shared_access_signature.py#L482-L574


ipanova:
> How come we only want to run this `if expire`?
 
reading trough the docs expire is set to None by default which means it never expires. Any reason we cannot create token_sas when expire is not set?

ipanova:
I see, it boils down to public and private access https://django-storages.readthedocs.io/en/latest/backends/azure.html#private-vs-public-access; tldr in private access there is always an expiration set,  and in public access i guess one does not need any credentials.

Issue #944, opened 2020-10-10
[DigitalOcean] Private objects signing does not work with custom domain


lifenautjoe:
When using a custom domain on your DigitalOcean space, such as `cdn.peekalink.com`, enabled by setting `AWS_S3_CUSTOM_DOMAIN=cdn.peekalink.com`, when having a storage with the property `default_acl='private'`, no signature is emitted on the object url and it cannot be accessed.

This is because on the S3BotoStorage, the `url` method checks whether theres a custom domain, and if there is, it tries to sign the object url with the CloudFront Signer, thing which obviously does not work for DigitalOcean Spaces.

```python
       # On s3boto.py line 564
        if self.custom_domain:
            url = "{}//{}/{}".format(
                self.url_protocol, self.custom_domain, filepath_to_uri(name))

            if self.querystring_auth and self.cloudfront_signer:
                expiration = datetime.utcnow() + timedelta(seconds=expire)

                return self.cloudfront_signer.generate_presigned_url(url, date_less_than=expiration)

            return url
```

I'm trying to figure out how to best do this but the most sensible thing probably to do is to create a `digitalocean_signer` that handles this edge case and can be set on an storage `digitalocean_signer` attribute.

Opening this issue so that there's some evidence that this problem exists and serving private resources stored on DigitalOcean won't work until this is solved.

lifenautjoe:
I'm not sure anymore whether this is even supported by DigitalOcean.

https://ideas.digitalocean.com/ideas/DO-I-2914

ZuSe:
We are running into the same problem on OVH.
Basically the whole Storage is based in Swift 3, but works well with most of the S3 API functions.

It would be great if signatures would work for custom domains as well. Even though the token will change after 1,24 or whatever hours it would still save GB of bandwidth.

lifenautjoe:
According to some of the new replies on that issue raised for digital ocean, signing the object with the non-cdn url and then just replacing it by the cdn-url works.

This would imply that we could "hotfix" this in this library.

I don't have some free space to do this at the moment, so feel free to give it a try.

awhileback:
Arriving here from google researching this on mixing public/private objects: throwing in an idea I'm about to test...

Could this not be solved by changing the storage class without having to fork/change the code?

    @deconstructible
    class PrivStorage(S3Boto3Storage):

        custom_domain = None
        bucket_name = settings.AWS_PRIVSTORAGE_BUCKET_NAME
[...]

Issue #395, opened 2017-09-23
Digital Ocean Spaces support


paramono:
Digital Ocean launched a new object storage service called Spaces. Are there any plans to support Spaces in this package?

destos:
You can use the existing `S3BotoStorage` class as their API is "the same" or close.

EdNutting:
Here's some example config for you - this works with current stable version of Django storages:

| Name                          | Description                                                                  | Example value                                                      |
|-------------------------------|------------------------------------------------------------------------------|--------------------------------------------------------------------|
| AWS\_STORAGE\_BUCKET\_NAME    | The name of the storage bucket.                                              | mysite                                                            |
| AWS\_S3\_ENDPOINT\_URL        | The endpoint url excluding the bucket name.                                  | https://nyc3.digitaloceanspaces.com                                |
| AWS\_S3\_OBJECT\_PARAMETERS   | Default params for objects.                                                  | { 'CacheControl': 'max-age=86400' }                                |
| AWS\_LOCATION                 | The folder within the space to store files.                                  | mysite                                                            |
| AWS\_S3\_SIGNATURE\_VERSION   | Signature version. DigitalOcean only support v2 for pre-signed urls.         | s3=s3v2 (Version 2), s3v4=s3v4 (Version 4)                         |
| AWS\_STATIC\_LOCATION         | The folder within the space to store static files.                           | '%s/static' % AWS_LOCATION                                         |
| STATICFILES\_STORAGE          | The storage backend to use for static files.                                 | 'op\_storage.storage\_backends.StaticStorage'                      |
| STATIC\_URL                   | The base URL for generating URLs to static files included in rendered pages. | "https://%s/%s/" % (AWS\_S3\_ENDPOINT\_URL, AWS\_STATIC\_LOCATION) |
| AWS\_PUBLIC\_MEDIA\_LOCATION  | The folder within the space to store public media files.                     | '%s/media/public' % AWS_LOCATION                                   |
| DEFAULT\_FILE\_STORAGE        | The storage backend to use for public media files.                           | 'op\_storage.storage\_backends.PublicMediaStorage'                 |
| AWS\_PRIVATE\_MEDIA\_LOCATION | The folder within the space to store private media files.                    | '%s/media/private' % AWS_LOCATION                                  |
| PRIVATE\_FILE\_STORAGE        | The storage backend to use for private media files.                          | 'op\_storage.storage_backends.PrivateMediaStorage'                 |

### Private Configuration

[...]

PR #1075, opened 2021-10-15
Docs for multiple buckets with varying properties, Digital Ocean signed URLs with CDN domain

awhileback:
from guide posted in #944, also referenced in #395

I double checked and cleaned up the example template filter, but I don't use DO Spaces and would rather not pay 5 bucks to test this if possible, perhaps @lifenautjoe can verify that this all works on his DO account.

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -408,7 +408,7 @@ You could also pass different characteristics to different buckets::
 
     s3_priv_storage = PrivStorage()
 
-The above setup would use two buckets, one that could use an ``AWS_S3_CUSTOM_DOMAIN`` specified in your settings for static files in one folder and media files in another folder of the same bucket, and a separate private bucket that would ignore the custom domain setting and use signed S3 urls to store private media files in a separate bucket.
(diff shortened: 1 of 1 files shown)


paulkuhle:
I followed the steps in the updated docs and can verify that they work perfectly. 

One minor thing: I had to also add the storage bucket name to the replace function, given that django-storages generates the link in this format (`https://fra1.digitaloceanspaces.com/example-cdn` instead of `https://example-cdn.fra1.digitaloceanspaces.com`).

This is a quick example of how this would look like in a graphene resolver (I imagine the same principle applies to template filters, DRF endpoints, etc.):
```python
def resolve_file(parent, info):
    url = parent.file.url
    if settings.AWS_S3_ENDPOINT_URL in url:
        cdn_domain = 'https://cdn.example.com'
        new_url = url.replace(
            f"{settings.AWS_S3_ENDPOINT_URL}/{settings.AWS_STORAGE_BUCKET_NAME}", cdn_domain)
        return new_url
    return url
```

It would be great if we could get this merged soon. :+1:

jambonrose:
Thank you for writing these docs.

Unfortunately, the [Digital Ocean docs](https://docs.digitalocean.com/products/spaces/how-to/use-aws-sdks/#calling-s3-commands) now has a section that states: 

> Using presigned URLs does not allow transferred files to be cached when using the Spaces CDN. Attempting to do so may result in double the bandwidth charge without the CDN’s performance benefit.

If we include docs for DigitalOcean, we should recommend that people use _either_ presigned URLs or the CDN, but not both.

Issue #1073, opened 2021-10-13
collectstatic fails


spechtx:
Hello everyone,

I might came across a potential issue with the azure functionality in django-storages in version 1.12.1.

Django's python manage.py collectstatic fails with following error:

`azure.core.exceptions.ServiceRequestError: <urllib3.connection.HTTPSConnection object at 0x00000226A68C26D0>: Failed to establish a new connection: [Errno 11001] getaddrinfo failed`

However the combination of the following works just fine, so i assume its not my settings, that cause the error:
django-storages==1.11.1
azure-storage-blob==2.1.0
azure-storage-common==2.1.0

Thanks :)

pjsier:
Unfortunately I haven't been working with this for a while, but this sounds familiar to the error described here https://stackoverflow.com/a/64600812 where the resources may not exist

PR #1076, opened 2021-10-15
Restore AZURE_CUSTOM_DOMAIN to behave as at 1.11

dimbleby:
Per discussion at https://github.com/jschneier/django-storages/issues/1073#issuecomment-944389156

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 django-storages CHANGELOG
 =========================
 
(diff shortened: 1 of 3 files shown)


jschneier:
@dimbleby So would the accurate statement for the releases notes be "Fix regression where AZURE_CUSTOM_DOMAIN was interpreted as a replacement of `blob.core.windows.net` rather than as a full domain.

Issue #541, opened 2018-07-30
Any plans for Alibaba Cloud Object Storage Service (OSS) backend?


hussaintamboli:
https://github.com/aliyun/aliyun-oss-python-sdk

yywing:
https://help.aliyun.com/apsara/agile-data/v_2_5_0_20200506/oss/insight-developer-guide/s3-api-compatibility-instructions.html

anyidea:
@sww314 i have added usage docs in my PR #1101 .

jschneier:
See comment in #1101

PR #1101, opened 2022-01-09
Alibaba OSS usage docs.

anyidea:
Alibaba OSS is compatible with S3 #541


jschneier:
Sorry, I will accept a docs PR that lists out services compatible with the S3 protocol but don't want to accrete backends.

jschneier:
Sorry, I see this is just docs so will take a look again in the future.

anyidea:
> Sorry, I see this is just docs so will take a look again in the future.

Plz review it

PR #1105, opened 2022-01-27
Added instructions to use S3 backend with Backblaze B2

metadaddy:
Backblaze B2 is a widely used, S3-compatible cloud object storage service. After [answering a question on using B2 with django-storages on Reddit](https://www.reddit.com/r/backblaze/comments/sdb2m7/comment/hud0r20/?utm_source=share&utm_medium=web2x&context=3), I figured I'd drop the same information here.

### DIFF
--- a/docs/backends/backblaze-B2.rst
+++ b/docs/backends/backblaze-B2.rst
@@ -1,0 +1,13 @@
+Backblaze B2
+============
+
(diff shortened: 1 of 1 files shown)


jschneier:
Thanks, this is helpful.

PR #1128, opened 2022-04-21
Dropbox: add support for refresh tokens as new access tokens are short-lived

### LINKED ISSUE #993: Documentation/feature update for Dropbox api v2
The Dropbox API has changed and now supports refresh tokens : https://developers.dropbox.com/oauth-guide
Currently the oAUTH token to be used for Django-storages is only provided as ‘short-lived’ for 4 hours by Dropbox and a new one can be used by sending a refresh token. Documentation or feature update necessary?

### LINKED ISSUE #1018: allow dropbox backend to work with refresh tokens
Hi, I am using the django-storages as part of django-dbbackup, and am configuring the dropbox parameters in my settings.  Everything works fine as expected, but in the dropbox apps info, it is stated that the long-expiration oauth token will be deprecated in the future.

It looks like, according to the dropbox api, the /token/from_oauth1 endpoint can be used in conjunction with your secret key to return an oauth2 shortlived token to allow a call to dropbox function.

I understand that this project is unfortunately not well supported, so I don't expect any changes, but just wanted to point out the potential necessity of making a change before dropbox drop support for the long lived access tokens.

Mtia

dkreeft:
New Dropbox applications do not allow you to generate non-expiring access tokens any longer. This PR implements support for using a refresh token together with the app key and secret to deal with short-lived tokens. Support for access tokens only is still present for legacy purposes.

Closes #993, closes #1018


dkreeft:
@jschneier bugging you as described in the `README`

jschneier:
@dkreeft thanks for the ping. Code looks good although tests seem to fail.

Can you also change the docs formatting [to use a list](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/lists.html).

If access tokens are indeed being deprecated it might be worth calling that out more explicitly in the authenticating section.

jschneier:
Fixed by #1159.

dkreeft:
#1159 seems to be an exact copy of my work, besides some additional instructions. @jschneier what made you decide to credit @begoon for "his" work instead of me or to not also include me?

PR #1134, opened 2022-05-08
[azure] fix get_modified_time()

### LINKED ISSUE #1131: AzureStorage get_modified_time  'ContainerClient' object has no attribute 'get_blob_properties'
I suspect this is a bug from a change in Azure's python client. I can't say I'm an expert but it looks like [this](https://github.com/jschneier/django-storages/blob/master/storages/backends/azure_storage.py#L337) line needs be changed from.

```python
        properties = self.client.get_blob_properties(
            self._get_valid_path(name),
            timeout=self.timeout)
```

to

```python
        properties = self.client.get_blob_client(self._get_valid_path(name)).get_blob_properties(
            timeout=self.timeout)
```

Reproduction steps

```
from storages.backends.azure_storage import AzureStorage
storage = AzureStorage()
storage.get_modified_time("your-file")

AttributeError: 'ContainerClient' object has no attribute 'get_blob_properties'
```

I could submit a pull request if desired. It might be good for another set of eyes to confirm this is the right approach to fix.

dimbleby:
Fixes #1131 

fix is basically as suggested, but I've formatted it for consistency with https://github.com/jschneier/django-storages/blob/0ceb9e03b89a00395cad9bd5ce3909cf99221cd1/storages/backends/azure_storage.py#L253-L256

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -334,9 +334,8 @@ class AzureStorage(BaseStorage):
         Returns an (aware) datetime object containing the last modified time if
         USE_TZ is True, otherwise returns a naive datetime in the local timezone.
         """
-        properties = self.client.get_blob_properties(
-            self._get_valid_path(name),
-            timeout=self.timeout)
+        blob_client = self.client.get_blob_client(self._get_valid_path(name))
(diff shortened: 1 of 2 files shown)


dimbleby:
c99f8cafe18387cf5cd546e95b350ba5e3e414a9, I expect

PR #1135, opened 2022-05-13
Fix install instructions

frankjlin16:
The install instruction needs to be updated. The azure package is deprecated and broken down into separate packages. In addition, the new django-storages doesn't work with the old azure package due to this issue (https://stackoverflow.com/questions/50581138/cannot-import-name-blockblobservice). 

I propose this new update to the installation process. I have tested this process and it works perfectly. Feel free to reach out to me if you have any questions.

### DIFF
--- a/docs/backends/azure.rst
+++ b/docs/backends/azure.rst
@@ -28,10 +28,11 @@ Gzipping for static files must be done through Azure CDN.
 Install
 *******
 
-Install Azure SDK::
+Install Azure SDK (in addition to pip installing django-storages)::
 
-  pip install django-storages[azure]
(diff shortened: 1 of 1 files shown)


dimbleby:
You have misunderstood something and your suggestion does not make sense.  

Installing `django-storages[azure]` installs `django-storage` along with azure extras as defined in this project - which is to say, along with `azure-storage-blob`.

https://github.com/jschneier/django-storages/blob/0ceb9e03b89a00395cad9bd5ce3909cf99221cd1/setup.cfg#L41-L42

jschneier:
Closing as invalid.

PR #1146, opened 2022-06-22
IBM COS support

RankoR:
Same as https://github.com/jschneier/django-storages/pull/1009, but:

* Fixed tests
* Fixed formatting
* Add missing settings


RankoR:
Hey @jschneier, could you please approve the workflow?

jschneier:
Hmm there are many services that implement the S3 protocol, I’m not sure if it’s worth adding a backend for each of them.

Is there anything special about IBM per-se that wouldn’t be covered by docs similar to [Digital Ocean](https://github.com/jschneier/django-storages/blob/master/docs/backends/digital-ocean-spaces.rst)

jschneier:
I will accept a PR that adds more documentation about the S3 protocol but don't want to accrete backends / code. Sorry.

PR #1148, opened 2022-06-25
Add variable to use s3 scheme in uri

mjlabe:
- use AWS_S3_USE_S3_URI_SCHEME to update url_protocol to s3
- add appropriate test
- add documentation

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -104,6 +104,9 @@ searches for them:
 ``AWS_S3_ENDPOINT_URL`` (optional: default is ``None``)
     Custom S3 URL to use when connecting to S3, including scheme. Overrides ``AWS_S3_REGION_NAME`` and ``AWS_S3_USE_SSL``. To avoid ``AuthorizationQueryParametersError`` error, ``AWS_S3_REGION_NAME`` should also be set.
 
+``AWS_S3_USE_S3_URI_SCHEME`` (optional: default is ``False``)
+    Custom `s3` URI scheme use when connecting to S3.
+
 ``AWS_S3_ADDRESSING_STYLE`` (optional: default is ``None``)
     Possible values ``virtual`` and ``path``.
 
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -258,7 +258,10 @@ class S3Boto3Storage(CompressStorageMixin, BaseStorage):
         # Backward-compatibility: given the anteriority of the SECURE_URL setting
         # we fall back to https if specified in order to avoid the construction
         # of unsecure urls.
-        if self.secure_urls:
+        # If the AWS_S3_USE_S3_URI_SCHEME variable is True, use s3 scheme for the uri.
+        if self.use_s3_uri_scheme:
+            self.url_protocol = "s3:"
+        elif self.secure_urls:
             self.url_protocol = 'https:'
 
         self._bucket = None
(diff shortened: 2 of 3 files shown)


storages/backends/s3boto3.py
```diff
@@ -258,7 +258,10 @@ def __init__(self, **settings):
         # Backward-compatibility: given the anteriority of the SECURE_URL setting
         # we fall back to https if specified in order to avoid the construction
         # of unsecure urls.
-        if self.secure_urls:
+        # If the AWS_S3_USE_S3_URI_SCHEME variable is True, use s3 scheme for the uri.
+        if self.use_s3_uri_scheme:
+            self.url_protocol = "s3:"
```
jschneier, line 263: Can you use single quotes here, that is the repo style.

docs/backends/amazon-S3.rst
```diff
@@ -104,6 +104,9 @@ searches for them:
 ``AWS_S3_ENDPOINT_URL`` (optional: default is ``None``)
     Custom S3 URL to use when connecting to S3, including scheme. Overrides ``AWS_S3_REGION_NAME`` and ``AWS_S3_USE_SSL``. To avoid ``AuthorizationQueryParametersError`` error, ``AWS_S3_REGION_NAME`` should also be set.
 
+``AWS_S3_USE_S3_URI_SCHEME`` (optional: default is ``False``)
+    Custom `s3` URI scheme use when connecting to S3.
```
jschneier, line 108: Can you change this to `Use an 's3' URI scheme for connections.`

There's nothing really custom about it.

tests/test_s3boto3.py
```diff
@@ -756,6 +756,12 @@ def test_override_settings(self):
             storage = s3boto3.S3Boto3Storage()
             self.assertEqual(storage.location, 'foo2')
 
+    def test_s3_uri_scheme_override(self):
+        with override_settings(AWS_S3_USE_S3_URI_SCHEME=True):
+            storage = s3boto3.S3Boto3Storage()
+            self.assertEqual(storage.use_s3_uri_scheme, 's3')
```
jschneier, line 762: Need to fix this test.

docs/backends/amazon-S3.rst
jschneier: Also, this only matters if custom domain is enabled, so that should be mentioned as well.

tests/test_s3boto3.py
```diff
@@ -756,6 +756,12 @@ def test_override_settings(self):
...
+            self.assertEqual(storage.url_protocol, 's3')
```
jschneier, line 763: Since is it required for custom domain to be enabled we should add a test that constructs a connection and validate that it is of the correct format in all cases.

jschneier:
Actually, can't you just set `AWS_S3_URL_PROTOCOL = 's3:'`?

mjlabe:
Can you?! I couldn't find it in the docs. I only saw `secure` in the code. I will test and close if so.

jschneier:
Yeah, need to do a scrub on settings and document what isn't documented. Secure vs protocol is very confusing and likely needs to be condensed down.

mjlabe:
I haven't had a change to test, but if that's the case I'll try to update the docs a little instead.

mjlabe:
I remember the issue now looking:
```
if self.secure_urls:
    self.url_protocol = 'https:'
    ...

'secure_urls': setting('AWS_S3_SECURE_URLS', True),
...
'url_protocol': setting('AWS_S3_URL_PROTOCOL', 'http:'),
```

One would have to set both:
```
AWS_S3_URL_PROTOCOL = "s3:"
AWS_S3_SECURE_URLS = False
```

Not a huge deal, but I didn't like it at the time. In hindsight, I'll just update the docs.

mjlabe:
Closing and resolving [Issue 1151](https://github.com/jschneier/django-storages/issues/1150) with [PR-1151](https://github.com/jschneier/django-storages/pull/1151).

Issue #1157, opened 2022-07-19
Dropbox backend url() function throws ApiError when remote file is missing


begoon:
When Django `FileField` or `ImageField` use django-storage Dropbox backend, and if the remote file associated with the field is missing (for example, manually deleted from the storage), the `url()` function throws `ApiError` exception.

The problem is related to Django Admin. The `change` view (when you edit the record) calls the `url()` function and fails reporting about the exception. There is no way to catch this exception to make the view working. The view does not work, and it is impossible to edit the record to clean up the broken field or image field.

```
ApiError at /admin/.../18/change/
ApiError('f165b9076e58427ab4605bbccce7b3d0', GetTemporaryLinkError('path', LookupError('not_found', None)))
Request Method:	GET
Request URL:	http://127.0.0.1:10000/admin/.../18/change/
Django Version:	4.0.5
Exception Type:	ApiError
Exception Value:	
ApiError('f165b9076e58427ab4605bbccce7b3d0', GetTemporaryLinkError('path', LookupError('not_found', None)))
Exception Location:	.../.venv/lib/python3.9/site-packages/dropbox/dropbox_client.py, line 351, in request
Python Executable:	.../.venv/bin/python
Python Version:	3.9.5
Python Path:	
['/...',
 '',
 '/Users/alexander/.pyenv/versions/3.9.5/lib/python39.zip',
 '/Users/alexander/.pyenv/versions/3.9.5/lib/python3.9',
 '/Users/alexander/.pyenv/versions/3.9.5/lib/python3.9/lib-dynload',
 '/Users/alexander/github/eden-aid/.venv/lib/python3.9/site-packages']
Server time:	Tue, 19 Jul 2022 19:40:01 +0000
```

The solution is to silent the `ApiException` in the `url()` function similar to the `exists()` function.

PR #1158, opened 2022-07-19
Ignore ApiError exception in url()

begoon:
This PR fixes https://github.com/jschneier/django-storages/issues/1157 by silencing `ApiError` in the Dropbox the `url()` function.


jschneier:
Should it return `None` instead of `''`?

begoon:
Yes. "None" is more appropriate. Fixed.

PR #1159, opened 2022-07-20
Add refresh token support

begoon:
This PR adds `DROPBOX_APP_KEY`, `DROPBOX_APP_SECRET` and `DROPBOX_OAUTH2_REFRESH_TOKEN` settings.

The PR addresses https://github.com/jschneier/django-storages/issues/1018, https://github.com/jschneier/django-storages/issues/993.

PR #1167, opened 2022-08-06
[dropbox] restore backwards compat for parameter order

jschneier:


jschneier:
Regression in #1159. Test failures were due to additional parameters being added in a non backwards-compatible way.

PR #1173, opened 2022-09-07
Use django_guid to fetch and apply an Azure correlationId

jasonzio:
Set the correlationId for Azure storage calls to match the django_guid set elsewhere in the app.

- try to import django_guid
- if the import succeeds, call get_guid
- if a non-empty result comes back, apply the value as the correlation_id when talking to storage

The django_guid module is optional, so I can't depend on it being available. Even if it is available, the app might not have called set_guid in the current context.


jasonzio:
Sorry, this was premature.

Issue #1097, opened 2021-12-12
Azure Blob Storage: "TypeError: '<' not supported between instances of 'int' and 'NoneType'"


AdmiralGT:
django-storages version: 1.12.3
Django version: 3.2.10
azure-storage-blob version: 12.9.0
azure-core version: 1.21.1

I'm serving a simple FileResponse and get the following error

```
2021-12-12T13:36:51.739891275Z Traceback (most recent call last):
2021-12-12T13:36:51.739908176Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/workers/sync.py", line 136, in handle
2021-12-12T13:36:51.739920476Z     self.handle_request(listener, req, client, addr)
2021-12-12T13:36:51.739931676Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/workers/sync.py", line 182, in handle_request
2021-12-12T13:36:51.739943076Z     resp.write_file(respiter)
2021-12-12T13:36:51.739953676Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/http/wsgi.py", line 385, in write_file
2021-12-12T13:36:51.739964877Z     if not self.sendfile(respiter):
2021-12-12T13:36:51.739975477Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/http/wsgi.py", line 355, in sendfile
2021-12-12T13:36:51.739986577Z     if not util.has_fileno(respiter.filelike):
2021-12-12T13:36:51.739997177Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/util.py", line 569, in has_fileno
2021-12-12T13:36:51.740008278Z     if not hasattr(obj, "fileno"):
2021-12-12T13:36:51.740019178Z   File "/tmp/8d9bd73cfc52ab6/antenv/lib/python3.8/site-packages/django/core/files/utils.py", line 38, in <lambda>
2021-12-12T13:36:51.740031578Z     fileno = property(lambda self: self.file.fileno)
2021-12-12T13:36:51.740042278Z   File "/tmp/8d9bd73cfc52ab6/antenv/lib/python3.8/site-packages/django/core/files/utils.py", line 38, in <lambda>
2021-12-12T13:36:51.740053978Z     fileno = property(lambda self: self.file.fileno)
[...]

PR #1176, opened 2022-09-22
[azure] Use `AZURE_CUSTOM_DOMAIN` only for retrieving blob URLs, and use storage URL for other operations

### LINKED ISSUE #1116: Azure - Authentication error when `AZURE_CUSTOM_DOMAIN` set to Azure CDN
Ever since the Azure backend was updated to the new `azure-storage-blob` library in `v1.12`, using django-storages with `AZURE_CUSTOM_DOMAIN` set results in Authentication errors when uploading files. For me this only happens with Akamai CDNs, but as reported by others below it affects other CDN types as well.

`v1.12` changed how `AZURE_CUSTOM_DOMAIN` is used with BlobServiceClient. In `v1.11` and earlier, the custom domain was only used to get blob URLs. All other operations like uploading, streaming, getting metadata were being done by making requests to the actual storage account endpoint (`https://<accountname>.blob.core.windows.net`) even if a custom domain was specified.

In `v1.12`, the behaviour changed so that the custom domain endpoint is used for all storage operations. This uncovered several different upstream issues causing various storage requests to fail with auth errors.

### 1. Auth error due to MAC signature mismatch when `AZURE_CUSTOM_DOMAIN` set to Akamai CDN
Upstream issue: https://github.com/Azure/azure-sdk-for-python/issues/26381

Uploads fail with the following error:

```
ClientAuthenticationError: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
...
authenticationerrordetail:The MAC signature found in the HTTP request '5dC3N7RcRW9V...' is not the same as any computed signature. Server used following string to sign: 'PUT


1

application/octet-stream






x-ms-blob-type:BlockBlob
x-ms-client-request-id:xxxx
x-ms-date:Mon, 21 Feb 2022 07:34:35 GMT
x-ms-version:2020-10-02
Content: <?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed 
correctly including the signature.
RequestId:xxx
Time:2022-02-21T07:34:35.1058206Z</Message><AuthenticationErrorDetail>The MAC signature found in the HTTP request '5dC3N7RcRW9V...' is not the same as any computed 
signature. Server used following string to sign: 'PUT
```

### 2. Forbidden ClientAuthenticationError when `AZURE_CUSTOM_DOMAIN` set to Microsoft CDN
Upstream issue: https://github.com/Azure/azure-sdk-for-python/issues/23640

Uploads fail with the following error (different than above):
```
azure.core.exceptions.ClientAuthenticationError: Operation returned an invalid status 'Forbidden'
ErrorCode:AuthenticationFailed
```

JeffreyCA:
Fixes #1116, #1097

This restores the previous behaviour that was in 1.11 (with old Azure backend) so that the custom domain is used only for retrieving the blob URLs. All other operations like uploads go through the actual storage account URL. I updated the docs to clarify this too.

In 1.12, we switched to the new storage SDK and also changed it so that all storage operations go through the custom domain instead (not sure if intentional or not). Doing so uncovered several upstream issues that affect storage operations when connecting directly to the custom domain. More details in #1116.

I also made a small change to simplify the `exists()` method a bit.

### DIFF
--- a/docs/backends/azure.rst
+++ b/docs/backends/azure.rst
@@ -144,7 +144,7 @@ The following settings are available:
 
 ``AZURE_CUSTOM_DOMAIN``
 
-    The custom domain to use. This can be set in the Azure Portal. For
+    The custom domain to use for generating URLs for files. For
     example, ``www.mydomain.com`` or ``mycdn.azureedge.net``.
 
 ``AZURE_CONNECTION_STRING``
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -260,11 +260,7 @@ class AzureStorage(BaseStorage):
 
     def exists(self, name):
         blob_client = self.client.get_blob_client(self._get_valid_path(name))
-        try:
-            blob_client.get_blob_properties()
-            return True
-        except ResourceNotFoundError:
-            return False
+        return blob_client.exists()
 
     def delete(self, name):
         try:
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -123,7 +123,9 @@ class AzureStorage(BaseStorage):
     def __init__(self, **settings):
         super().__init__(**settings)
         self._service_client = None
+        self._custom_service_client = None
         self._client = None
+        self._custom_client = None
         self._user_delegation_key = None
         self._user_delegation_key_expiry = datetime.utcnow()
 
@@ -150,11 +152,11 @@ class AzureStorage(BaseStorage):
             "api_version": setting('AZURE_API_VERSION', None),
         }
 
-    def _get_service_client(self):
+    def _get_service_client(self, use_custom_domain):
         if self.connection_string is not None:
             return BlobServiceClient.from_connection_string(self.connection_string)
 
-        account_domain = self.custom_domain or "{}.blob.{}".format(
+        account_domain = self.custom_domain if self.custom_domain and use_custom_domain else "{}.blob.{}".format(
             self.account_name,
             self.endpoint_suffix,
         )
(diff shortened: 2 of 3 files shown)


dimbleby:
If I understand correctly - which I may not, please put me right

- there are bugs such that using custom domains / content delivery networks don't work properly
- this MR works around that by bypassing those things and going direct to the storage account

Isn't that basically the wrong thing to do, albeit driven by the very pragmatic wish to have something that actually works?

If so it seems unsatisfactory to leave this project committed to doing the wrong thing: and to accept configuration making it look as though it does one thing when in fact it does another.  Sooner or later there'll be bug reports saying "I configured use of a custom domain, but that is not being respected"...

Perhaps it would be better simply to note the upstream bugs in the docs somewhere, and recommend that people don't configure things that way until they are fixed?

JeffreyCA:
This PR basically reverts the behaviour to what it was like in `v1.11` and earlier: using two service clients, a regular client that uses the storage endpoint and a "custom" client that uses the custom domain (if it exists, and falls back to storage endpoint otherwise). The custom client is only used to retrieve the blob URL, and the regular client is used for all other operations, which to me makes more sense than what the current behaviour is.

For reference this is what we were doing in the old version:

https://github.com/jschneier/django-storages/blob/3fbad31ed17ac34ab8885e077dab546aeeea01e6/storages/backends/azure_storage.py#L266-L282

If we keep the current implementation, there aren't any good workarounds for people who want to use Azure CDNs. Using Microsoft CDNs just doesn't work at all, and for Akamai CDNs you'd have to set `AZURE_OVERWRITE_FILES` to `True` so that it doesn't send a particular HTTP header.

jschneier:
Happy to merge this as-is. Can you explain how you tested this change? I don't think the unit tests properly cover everything unfortunately.

JeffreyCA:
Thanks for reviewing, I added a few more tests.

I used Fiddler to verify that the storage endpoint is used for uploads and other operations, and that the custom domain is used for generating the file URLs. I tested with all the different types of Azure CDNs and did not get any auth errors. Also verified the user delegation logic still works using an Azure VM w/ managed identity.

PR #1177, opened 2022-09-26
S3: Ensure that upload_fileobj doesn't close the file-object.

Hatell:
Work-around to issue when S3 save closes file-object.

https://github.com/boto/s3transfer/issues/80

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -9,6 +9,7 @@ from tempfile import SpooledTemporaryFile
 from urllib.parse import parse_qsl
 from urllib.parse import urlencode
 from urllib.parse import urlsplit
+from io import BytesIO
 
 from django.contrib.staticfiles.storage import ManifestFilesMixin
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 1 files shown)


jschneier:
This bug is an old friend.

In general we don't want to read the full file into memory which is why the proposed solution was never adopted.

However, the compromise was that if using `ManifestFilesMixin` we would add a workaround which you can see [here](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L580-L591).

Can you tell me how to reproduce this issue? If it was when attempting to use `ManifestFilesMixin` or the like maybe the docs need to be enhanced.

Hatell:
Hello, this issue is happening when we save a file in storage before handling it, so it won't be lost if something goes wrong. We don't want re-read it from storage because the file is in memory already.

Changing the backend to this `storages.backends.s3boto3.S3ManifestStaticStorage` could resolve our issue. I need to test this but it seems that it do correct job.

I think better solution would be a wrapper that prevents closing the file. Something like this:
```
def _save(...):
    .
    .
    .
    def do_nothing(*args, **kwargs):
        pass

    orginal_close = content.close
    content.close = do_nothing
    try:
        obj.upload_fileobj(content, ExtraArgs=params, Config=self._transfer_config)
    finally:
        content.close = orginal_close
    return cleaned_name
```

jschneier:
> I think better solution would be a wrapper that prevents closing the file. Something like this:
> 
> ```
> def _save(...):
>     .
>     .
>     .
>     def do_nothing(*args, **kwargs):
>         pass
> 
>     orginal_close = content.close
>     content.close = do_nothing
>     try:
>         obj.upload_fileobj(content, ExtraArgs=params, Config=self._transfer_config)
>     finally:
>         content.close = orginal_close
>     return cleaned_name
> ```

This is a great solution and I can't believe I never thought of it.

jschneier:
Opened that as a solution in https://github.com/jschneier/django-storages/pull/1303

PR #1178, opened 2022-09-28
[Azure] Supply guid from django_guid as the correlationId on storage calls

jasonzio:
If the django-guid package is installed and `django_guid.set_guid()` has been called, fetch that guid and apply it as the correlationId for all Azure storage API calls. If the module isn't `import`able or no guid has been set, don't supply a correlationId.

While the test environment requires the django-guid package, the runtime environment need not have that package installed.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -22,6 +22,21 @@ from storages.utils import setting
 from storages.utils import to_bytes
 
 
+try:
+    from django_guid import get_guid
+except ImportError:
+    def get_guid():
+        return None
+
+
+def _make_headers():
+    headers = {}
+    correlation_id = get_guid()
+    if correlation_id is not None:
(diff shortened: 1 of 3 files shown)


storages/backends/azure_storage.py
```diff
@@ -24,6 +24,20 @@
...
+try:
+    from django_guid import get_guid
+except ImportError:
+    def get_guid():
+        return None
+
+
+def _make_headers():
+    headers = {}
+    correlation_id = get_guid()
```
hartungstenio, line 36: Is committing to django-guid for this something we want? What if people have other means of generating a Correlation ID (django-log-request-id or OpenTelemetry).

Maybe simple adding a `make_headers` method to the Azure Storage, and allowing users to override it using any solution they want, would be a simpler solution, without committing to any individual package.

Issue #1190, opened 2022-10-27
wrong usage of LibCloudStorage._get_object in LibCloudStorage._read


engAmirEng:
I bumped into this bug when I was trying to make **LibCloudStorage** work with django's **ManifestFilesMixin** but that doesn't matter and it should be fixed regardless.

exact version of what it is now:
```
def _get_object(self, name):
    """Get object by its name. [Return None if object not found"""
    clean_name = self._clean_name(name)
    try:
        return self.driver.get_object(self.bucket, clean_name)
    except ObjectDoesNotExistError:
        return None

def _read(self, name):
    obj = self._get_object(name)
    # TOFIX : we should be able to read chunk by chunk
    return next(self.driver.download_object_as_stream(obj, obj.size))
```
my recommendation:
```
def _read(self, name):
    obj = self._get_object(name)
    if obj is None:
        raise FileNotFoundError(f"{name} does not exist.")
    # TOFIX : we should be able to read chunk by chunk
    return next(self.driver.download_object_as_stream(obj, obj.size))
```

and if you are curious about the exact trigger of the bug:
```
class ManifestFilesMixin(HashedFilesMixin):
    def read_manifest(self):
        try:
            with self.manifest_storage.open(self.manifest_name) as manifest:
                return manifest.read().decode()
        except FileNotFoundError:
            return None
```

PR #1191, opened 2022-10-27
raise appropriate exception in LibCloudStorage._read (#1190)

engAmirEng:
prevent accessing size attribute on None value.


jschneier:
Yikes. Okay, this needs a bunch of cleanup in general.

Can you change it so `get_object` doesn't catch `ObjectDoesNotExistError` and then make it so:

`_read` catches it and re-raises `FileNotFoundError` like you did
`delete` catches & silences it (i.e `pass`)
`exists` catches & returns `False` (`True` if it didn't raise)
`size` & `url` do not catch

jschneier:
Thanks for opening!

jschneier:
Thanks this looks good, can you please merge in master or rebase to get the tests passing.

PR #1193, opened 2022-10-28
Add GCloud parameters as kwargs for storage.url

fao89:
It enables using kwargs, such as `content_type` https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.blob.Blob#google_cloud_storage_blob_Blob_generate_signed_url

Ref: https://github.com/jschneier/django-storages/pull/1071

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -297,7 +297,7 @@ class GoogleCloudStorage(CompressStorageMixin, BaseStorage):
         created = blob.time_created
         return created if setting('USE_TZ') else timezone.make_naive(created)
 
-    def url(self, name):
+    def url(self, name, parameters=None):
         """
         Return public url or a signed url for the Blob.
         This DOES NOT check for existance of Blob - that makes codes too slow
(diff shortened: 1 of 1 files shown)


ipanova:
@jschneier PTAL

ipanova (approved):
looks right to me! thanks!

jschneier:
This looks generally fine, it doesn't allow people to override `expiration`, `version` or sometimes `bucket_bound_hostname`, can you update to fix that? I think the last one is okay but the first two should be override-able.

fao89:
> This looks generally fine, it doesn't allow people to override `expiration`, `version` or sometimes `bucket_bound_hostname`, can you update to fix that? I think the last one is okay but the first two should be override-able.

@jschneier I've addressed your comment, could you please take another look?

jschneier:
Looks good, thanks.

fao89:
@jschneier do you have an idea when it will be released?

ipanova:
@jschneier PTAL https://github.com/jschneier/django-storages/pull/1193#issuecomment-1309251640

PR #1194, opened 2022-11-09
Improve SFTPStorage

vainu-arto:
* Move seekability test to a separate helper function
* Stop calling content.open() from SFTPStorage._save
* Require at least version 1.10.0 of paramiko
* Use SFTPClient.putfo() in SFTPStorage._save()

This produces a fairly major difference in performance, the original version needs over 10x the time it takes for the modified version to transfer a 380M test file. Transferring said file with the current version also requires about 2 gigabytes of memory, which drops to 50 megabytes after the change.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -46,7 +46,7 @@ google =
 libcloud =
     apache-libcloud
 sftp =
-    paramiko
+    paramiko >= 1.10.0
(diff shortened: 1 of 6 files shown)


jschneier:
Looks great, thanks.

PR #1197, opened 2022-12-01
check if file exists before deleting

jmelloy:
S3 was failing on first run because the files didn't exist yet, so check that they exist first.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -460,7 +460,11 @@ class S3Boto3Storage(CompressStorageMixin, BaseStorage):
 
     def delete(self, name):
         name = self._normalize_name(self._clean_name(name))
-        self.bucket.Object(name).delete()
+        try:
+            if self.exists(name):
+                self.bucket.Object(name).delete()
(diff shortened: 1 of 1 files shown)


jschneier:
Thanks, rather than running `exists`, which is another query people pay for, can you just catch the exception?

jmelloy:
Done

storages/backends/s3boto3.py
```diff
@@ -460,7 +460,10 @@ def _save(self, name, content):
 
     def delete(self, name):
         name = self._normalize_name(self._clean_name(name))
-        self.bucket.Object(name).delete()
+        try:
+            self.bucket.Object(name).delete()
+        except Exception as e:
+            print(e)
```
jschneier, line 466: Can you check for whatever specific error code there might be and just silence that and re-raise otherwise.

If you check the handling in `.exists()` and similar you will see what I mean.

jschneier:
Also, will need to rebase / merge due to conflicts.

jschneier:
Fixed in #1201.

PR #1206, opened 2023-01-01
Fix code block

pamelafox:
There's a code block that's currently not rendering as one due to a missing new line, so that's the main fix of this PR. Also fixed a django capitalization while I was in there.


jschneier:
Thanks!

PR #1209, opened 2023-01-13
Add local storage class with file overwriting capabilities

### LINKED ISSUE #1208: Local storage class with file overwrite
Hi @jschneier, I'd like to propose a companion [local storage class](https://gist.github.com/niccolomineo/2b034f97571ec8106ea41248b4e5156e) mimicking the `*_OVERWRITE_FILES` behavior for remote storages.

Would you accept a PR with such an addition?

niccolomineo:
Closes #1208

### DIFF
--- a/storages/base.py
+++ b/storages/base.py
@@ -1,5 +1,5 @@
 from django.core.exceptions import ImproperlyConfigured
-from django.core.files.storage import Storage
+from django.core.files.storage import FileSystemStorage, Storage
 
 
 class BaseStorage(Storage):
@@ -22,3 +22,9 @@ class BaseStorage(Storage):
 
     def get_default_settings(self):
         return {}
+
+
+class FileSystemOverwriteStorage(FileSystemStorage):
+    def get_available_name(self, name, max_length=None):
+        self.delete(name)
+        return super().get_available_name(name, max_length)


pauloxnet (approved):

pauloxnet (approved):

pauloxnet:
@jschneier can you take a look at this PR, please?

craigds (changes requested):
(i'm not a maintainer of this project, but)

There's a more efficient implementation of this - instead of calling delete(), just override `OS_OPEN_FLAGS`, and tell `get_available_name` to just return the original name as-is.

```
    # Don't throw errors if the file already exists when saving.
    # https://manpages.debian.org/bullseye/manpages-dev/open.2.en.html#O_EXCL
    OS_OPEN_FLAGS = FileSystemStorage.OS_OPEN_FLAGS & ~os.O_EXCL
```

niccolomineo:
> (i'm not a maintainer of this project, but)
> 
> There's a more efficient implementation of this - instead of calling delete(), just override `OS_OPEN_FLAGS`, and tell `get_available_name` to just return the original name as-is.
> 
> ```
>     # Don't throw errors if the file already exists when saving.
>     # https://manpages.debian.org/bullseye/manpages-dev/open.2.en.html#O_EXCL
>     OS_OPEN_FLAGS = FileSystemStorage.OS_OPEN_FLAGS & ~os.O_EXCL
> ```

I would avoid this: you're changing the class behaviour from outside.

jschneier:
I think this should be a request to upstream can you open a ticket on the tracker? In general I think I would fix this by overriding `FileSystemStorage`.

craigds:
'upstream': django?

> I would definitely avoid this, as you're changing / controlling the class behaviour from outside.

i wasn't suggesting monkeypatching anything - `OS_OPEN_FLAGS` is an attribute on Django's FileSystemStorage which can be overridden by the django-storages subclass.

jschneier:
Yes opening a ticket on Django.

PR #1210, opened 2023-01-24
Update sftp.rst

jmbreiner:
Please add. I was not able to get working till I realized I needed STATICFILES_STORAGE.


jschneier:
Thanks for the PR but I've gone with #1239 instead to cover the new Django nuances and the like.

PR #1220, opened 2023-02-18
[gcloud] remove deprecated GS_CACHE_CONTROL support

jschneier:


devfemibadmus:
so now we do?

```python
GS_OBJECT_PARAMETERS = {
    'CacheControl': 'max-age=86400',
    'prefix': 'my_folder/',
}
```

PR #1225, opened 2023-02-26
fixes issue:1216

santhosh-reddy03:
Please review the changes related to this issue, I have only updated the changes related to url functionality in AzureStorage.
I have tested by importing the module into my django application, haven't used test cases as they were not using connection_string.

Please let me know if they are any improvements/changes required.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -1,4 +1,5 @@
 import mimetypes
+import re
 from datetime import datetime
 from datetime import timedelta
 from tempfile import SpooledTemporaryFile
@@ -128,6 +129,9 @@ class AzureStorage(BaseStorage):
         self._custom_client = None
         self._user_delegation_key = None
         self._user_delegation_key_expiry = datetime.utcnow()
+        if self.connection_string and (not self.account_name or not self.account_key):
+            self.account_name = re.search(r'AccountName=(\w+);', self.connection_string)[1]
+            self.account_key = re.search(r'AccountKey=(.*);', self.connection_string)[1]
 
     def get_default_settings(self):
         return {


santhosh-reddy03:
I have updated the isort changes and pushed again.

jschneier:
Thanks for the PR. This would definitely fix the issue (a test is missing though).

More generally: is there a use case for setting `AZURE_ACCOUNT_NAME` & `AZURE_ACCOUNT_KEY` if you are already setting `AZURE_CONNECTION_STRING`? If not, then probably it makes sense to parse these variables out in the constructor rather than in the `url()` method.

Maybe @alibeyram can weigh in?

santhosh-reddy03:
When I was reproducing the issue, it was showing `account_key` was missing, but since we are able to set both `account_name` and `account_key` from connection_string, I m updating both.

@jschneier I have now updated the changes inside `__init__` function, Please check.

jschneier:
@santhosh-reddy03 the indentation is wrong causing the lint tests to fail. Can you please fix.

jschneier:
@santhosh-reddy03 failed again due to trailing whitespace.

jschneier:
Merged, thanks.

jschneier:
@santhosh-reddy03 where did the regex you used come from btw? the one in the issue is different. did you try generating connection strings with various characters?

santhosh-reddy03:
No, for storage `account_name`, its alpha numerical's and for `account_key` they have not provided any specifications so I was considering all the characters till the next separation `;`. Will deep dive and see if there are any specifications or format for account_keys.

santhosh-reddy03:
We can also use `find` method, if there are any issues with regex `string[string.find('AccountName'): string.find(';', string.find('AccountName'))].replace("AccountName=", '')`

PR #1226, opened 2023-03-28
Make it easy to customise expiration for cloudfront

### LINKED ISSUE #1222: Customisable expiration for s3 backend
I would like to be able to modify the behavior in terms of generating pre-signed urls.
Now the generation of pre-signed URLs is based on the current time. 
This leads to the fact that the user refreshing the page will receive each time a new link and the browser cache in his case does not work.
In order to override this behavior, it's essentially copying a lot of code, which I'd like to avoid.

https://github.com/jschneier/django-storages/blob/4ad22c3a23048cb5f44c11d4f38d75b8173fec22/storages/backends/s3boto3.py#L580-L582

I want to override this in order to have discrete values of expiration time, acting over some period of time.

It would be nice to have something like this to be able to modify behavior

```python
...
def get_expiration(expire):
    return datetime.utcnow() + timedelta(seconds=expire)
...
expiration = self.get_expiration(expire)
```

image

Mogost:
Fix #1222


jschneier:
Hi thanks for the PR.

Instead of this solution how about one where `expire` can be any of `int|datetime|timedelta`? Would that solve your problem?

Mogost:
Hi. That doesn't help me in any way.
```
expiration = datetime.utcnow() + timedelta(seconds=expire)
```
If we're talking about this expression, I have a problem with the part `datetime.utcnow()` not in `timedelta(seconds=expire)`.

My goal is for links to be generated at discrete time intervals.
Like:
1:00 -> 1:30
1:05 -> 1:30 
1:15 -> 1:45
1:29 -> 1:45 
1:30 -> 2:00
In this case, my link is valid at least 15 minutes. Maximum validity is 30 min.
Now it works like 
1:00 -> 1:30
1:01 -> 1:31
And it doesn't give any chance for the cache to work.

PR #1227, opened 2023-03-28
Add open method on S3Boto3StorageFile with test

rabbit-aaron:


skrat:
I don't like this. `open` method on any kind of file handle is a non-sense. There's no (hopefuly) API out there that would do something like this. File object / value / handle, per standard practices, is a result (return value) of opening some path, URL, etc. Once it's opened (in certain mode), it can be read, written and closed. Seeking is a relic from the past, but ok, it can be useful, and it can be implemented on top of a buffer.

NOTE: your __init__ -> __initialize adds nothing, solves nothing.

rabbit-aaron:
> I don't like this. `open` method on any kind of file handle is a non-sense. There's no (hopefuly) API out there that would do something like this. File object / value / handle, per standard practices, is a result (return value) of opening some path, URL, etc. Once it's opened (in certain mode), it can be read, written and closed. Seeking is a relic from the past, but ok, it can be useful, and it can be implemented on top of a buffer.
> 
> NOTE: your `__init__` -> `__initialize` adds nothing, solves nothing.


```File object / value / handle, per standard practices, is a result (return value) of opening some path, URL, etc```

This is following its super class's design. As for what you've described, it was storage.open, which is completely different.

I believe Django do this to support `instant.file_field.open`

calling `__init__` in `open` looks weird, thus the renaming.

jschneier:
Hi and thanks for the PR. I decided to go another way (see #1321)

PR #1228, opened 2023-04-04
docs: update STORAGES syntax for Django 4.2

paulschreiber:
The current example (a) does not match the [Django docs](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES) and (b) crashes.

Other pages likely need updating as well.


jschneier:
Obsoleted by #1229, I merged that one since it covered all backends in addition to the README.

Issue #1207, opened 2023-01-12
Make django-storages compatible with the new Django's settings `STORAGES`


pauloxnet:
In Django 4.2 the new settings ``STORAGES`` has been added.
https://github.com/django/django/commit/1ec3f0961fedbe01f174b78ef2805a9d4f3844b1

The ``DEFAULT_FILE_STORAGE`` and ``STATICFILES_STORAGE`` settings are deprecated in Django 4.2 and will be removed Django 5.1.
https://github.com/django/django/commit/32940d390a00a30a6409282d314d617667892841

philgyford:
It looks like the new docs are wrong. They say to use the new `STORAGES` setting like this:

```python
    STORAGES = {"default": "storages.backends.s3boto3.S3Boto3Storage"}
```

But it should be:

```python
    STORAGES = {"default": {"BACKEND": "storages.backends.s3boto3.S3Boto3Storage"}}
```

Django docs: https://docs.djangoproject.com/en/4.2/ref/settings/#storages

pauloxnet:
Thanks for reporting.
Can you open a PR with the fix?

philgyford:
When I get a chance. Currently trying to fix a couple of other "I upgraded to Django 4.2" bugs!

PR #1229, opened 2023-04-05
Fix STORAGES settings in docs for Django 4.2

philgyford:
Changed anything like

```python
STORAGES = {"default": "storages.backends.gcloud.GoogleCloudStorage"}
```

to

```python
STORAGES = {"default": {"BACKEND": "storages.backends.gcloud.GoogleCloudStorage"}}
```

And added a bit to the README about the settings for Django >= 4.2 (not sure if the wording is how you would do it, so obviously feel free to change it!).

For #1207


jschneier:
Thank you for the comprehensive update.

PR #1230, opened 2023-04-05
Fix doc for Django >= 4.2 storages setting

mounirmesselmeni:
The setting requires a dict with backend to be defined or it will fail, see https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES


jschneier:
Obsoleted by #1229, I merged that one since it also covered the README. Thanks for opening!

PR #1236, opened 2023-04-14
add support for Django 4.2

jschneier:


allanlewis:
When can we expect a release supporting Django 4.2? v1.13.2 only supports Django 4.1.

PR #1240, opened 2023-04-19
Implement new Django time metadata interface where possible

### LINKED ISSUE #1233: Missing implementations of the Django interface for time metadata due to missing timezone information
The various backends feature two sets of methods for accessing time metadata:

- `get_(accessed|created|modified)_time`: Official Django interface. The returned `datetime` must be timezone aware if Django setting `USE_TZ` is `True`, else naive *in the local timezone*. This means the remote timezone must be known in order to be able to either (if `USE_TZ=True`) attach timezone information to the `datetime` or else (if `USE_TZ=False`) first shift the `datetime` from remote timezone to local timezone (and then remove the timezone information). Since not all backends know the remote timezone, this is not always implemented. Instead, there is also the following.
- `(accessed|created|modified)_time`: `django-storages` methods. The returned `datetime` is always naive. However, it is unclear in which timezone that `datetime` is. Those backends that know the remote timezone first shift this `datetime` to the local timezone (and then remove the timezone information), whereas the others leave it without timezone information in the unknown remote timezone.

Availability (✅ = implemented, ❌ = not implemented):

| [backend](https://github.com/jschneier/django-storages/tree/master/storages/backends) | `accessed_time` | `created_time` | `modified_time` | [`get_accessed_time`](https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.Storage.get_accessed_time) |  [`get_created_time`](https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.Storage.get_created_time) | [`get_modified_time`](https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.Storage.get_modified_time) |
|-|-|-|-|-|-|-|
| [Apache Libcloud](https://github.com/jschneier/django-storages/blob/master/storages/backends/apache_libcloud.py) |❌|❌|❌|❌|❌|❌|
| [Azure](https://github.com/jschneier/django-storages/blob/master/storages/backends/azure_storage.py) |❌|❌|✅|❌|❌|✅|
| [Dropbox](https://github.com/jschneier/django-storages/blob/master/storages/backends/dropbox.py) |✅|❌|✅|❌|❌|❌|
| [FTP](https://github.com/jschneier/django-storages/blob/master/storages/backends/ftp.py) |❌|❌|✅|❌|❌|❌|
| [Google Cloud](https://github.com/jschneier/django-storages/blob/master/storages/backends/gcloud.py) |❌|❌|✅|❌|✅|✅|
| [S3 Boto3](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py) |❌|❌|✅|❌|❌|✅|
| [SFTP](https://github.com/jschneier/django-storages/blob/master/storages/backends/sftpstorage.py) |✅|❌|✅|❌|❌|❌|

Options:
1. Fully implement the official Django interface. I suppose this is not possible due to missing timezone information for some backends.
2. Partially implement the official Django interface. For example, only raise `NotImplementedError` if `USE_TZ` is `True` but timezone information is missing. However, I doubt that conforms to the interface considering the Django documentation explicitly states that the naive `datetime` must be in the local timezone, which means the remote timezone must be known to be able to shift the naive `datetime` to the local timezone.
3. Warn about this peculiarity by explaining *why*  the method is not implemented. For example:
   ```py
   def get_modified_time(name):
       raise NotImplementedError(
           "SFTP does not provide timezone information. "
           "Use modified_time for a naive datetime instead."
   )
    ```
   This would not only inform users what is going on but document for developers that this is intended behavior.
4. Allow providing the (somehow known) remote timezone information (say, through the constructor) and only raise `NotImplementedError` if timezone information is missing.

TimoGuenther:
Closes #1233.

This PR normalizes the distribution of implementations of Django's newer `get_*_time` and its older `*_time` interfaces across the various storage backends. It does not add any new specific functionality, e.g., modified time on some backend that previously did not have this functionality.

On a technical level, this PR introduces a new `TimeStorageMixin` containing time metadata logic. This simplifies storage backends and their tests because they no longer have to deal with the specifics of the Django interface (i.e., converting timezones based on the `USE_TZ` setting). To do so, each subclass implements a custom `_*_time` interface that can return a naive or aware datetime, and `TimeStorageMixin` handles the rest.

Subclasses of `TimeStorageMixin` automatically support both the newer and the older interface if timezone information is present, else only the older `*_time` interface, by raising `NotImplementedError` where appropriate.

To ensure the new interface is definitely supported, subclasses may set `_default_timezone`. Azure, Dropbox, Google Cloud and S3 Boto3 all assume UTC datetime values by default, even if the underlying library should somehow fail to return aware datetime values. FTP and SFTP do not have a default timezone, but end users are free to subclass the storage and provide it if known. Apache Libcloud does not feature any time metadata methods.

Availability (✅ = implemented, ❌ = not implemented, ⚠️ = only old interface implemented but new interface easily implemented by providing a default timezone):

| [backend](https://github.com/jschneier/django-storages/tree/master/storages/backends) | [`get_accessed_time`](https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.Storage.get_accessed_time) |  [`get_created_time`](https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.Storage.get_created_time) | [`get_modified_time`](https://docs.djangoproject.com/en/4.2/ref/files/storage/#django.core.files.storage.Storage.get_modified_time) |
|-|-|-|-|
| [Apache Libcloud](https://github.com/jschneier/django-storages/blob/master/storages/backends/apache_libcloud.py) |❌|❌|❌|❌|❌|❌|
| [Azure](https://github.com/jschneier/django-storages/blob/master/storages/backends/azure_storage.py) |❌|❌|✅|
| [Dropbox](https://github.com/jschneier/django-storages/blob/master/storages/backends/dropbox.py) |✅|❌|✅|
| [FTP](https://github.com/jschneier/django-storages/blob/master/storages/backends/ftp.py) |❌|❌|⚠️|
| [Google Cloud](https://github.com/jschneier/django-storages/blob/master/storages/backends/gcloud.py) |❌|✅|✅|
| [S3 Boto3](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py) |❌|❌|✅|
| [SFTP](https://github.com/jschneier/django-storages/blob/master/storages/backends/sftpstorage.py) |⚠️|❌|⚠️|

Breaking changes:

- The implementations of the old interface may no longer return the same naive datetime where a default timezone is set, e.g., a naive datetime in the current timezone instead of a naive datetime in UTC. This was never specified by Django (hence the deprecation), but some end users might depend on it. Use the new interface instead and, if required, manually apply `make_naive` to its return value.

### DIFF
--- a/docs/backends/azure.rst
+++ b/docs/backends/azure.rst
@@ -91,7 +91,7 @@ The following settings should be set within the standard Django
 configuration file, usually `settings.py`.
 
 Set the default storage (i.e: for media files) and the static storage
-(i.e: fo static files) to use the azure backend::
+(i.e: for static files) to use the Azure backend::
 
     # django < 4.2
     DEFAULT_FILE_STORAGE = 'storages.backends.azure_storage.AzureStorage'
@@ -155,7 +155,7 @@ The following settings are available:
 
 ``AZURE_ENDPOINT_SUFFIX``
 
-    Defaults to ``core.windows.net``. Use ``core.chinacloudapi.cn`` for Azure.cn accounts.
+    Defaults to ``core.windows.net``. Use ``core.chinacloudapi.cn`` for azure.cn accounts.
 
 ``AZURE_CUSTOM_DOMAIN``
 
--- a/docs/backends/dropbox.rst
+++ b/docs/backends/dropbox.rst
@@ -12,13 +12,13 @@ which can be done for you automatically by doing::
 Settings
 --------
 
-To use DropBoxStorage set::
+To use DropboxStorage set::
 
     # django < 4.2
-    DEFAULT_FILE_STORAGE = 'storages.backends.dropbox.DropBoxStorage'
+    DEFAULT_FILE_STORAGE = "storages.backends.dropbox.DropboxStorage"
 
     # django >= 4.2
-    STORAGES = {"default": {"BACKEND": "storages.backends.dropbox.DropBoxStorage"}}
+    STORAGES = {"default": {"BACKEND": "storages.backends.dropbox.DropboxStorage"}}
 
 Two methods of authenticating are supported:
 
@@ -43,14 +43,14 @@ Please set the following variables accordingly:
 
 The refresh token can be obtained using the `commandline-oauth.py`_ example from the `Dropbox SDK for Python`_.
 
-``DROPBOX_ROOT_PATH`` (optional, default ``'/'``)
+``DROPBOX_ROOT_PATH`` (optional, default ``"/"``)
    Path which will prefix all uploaded files. Must begin with a ``/``.
 
 ``DROPBOX_TIMEOUT`` (optional, default ``100``)
    Timeout in seconds for requests to the API. If ``None``, the client will wait forever.
    The default value matches the SDK at the time of this writing.
 
-``DROPBOX_WRITE_MODE`` (optional, default ``'add'``)
+``DROPBOX_WRITE_MODE`` (optional, default ``"add"``)
    Sets the Dropbox WriteMode strategy. Read more in the `official docs`_.
 
 Obtain the refresh token manually
--- a/docs/backends/oracle-cloud.rst
+++ b/docs/backends/oracle-cloud.rst
@@ -1,0 +1,32 @@
+Oracle Cloud
+=============
+
+Oracle cloud provides S3 compatible object storage. To use it follow the instructions in the :doc:`Amazon S3 docs <amazon-S3>` on how to configure *DEFAULT_FILE_STORAGE* and *STATICFILES_STORAGE* and set the following
+configurations on settings.py
+
+- Create a `Customer Secret Key`_
+- Use generated key as ``AWS_SECRET_ACCESS_KEY``
+- And the value in the *Access Key* column as ``AWS_ACCESS_KEY_ID``
+- Set ``AWS_STORAGE_BUCKET_NAME`` with your bucket name
+- Set ``AWS_S3_REGION_NAME`` with the current region
+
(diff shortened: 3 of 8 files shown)


cclauss:
I suddenly got quite interested in this pull request when I read the following note from my colleague this morning about our attempt to upgrade from Django v4.1 to v4.2...
> I haven't seen any breakages in displaying images from S3, but when trying to download a resource, it raises a 500 error: '`S3Storage`' object has no attribute '`modified_time`'.

@TimoGuenther Would it be possible to please rebase this PR given all the recent changes to the repo?
@jschneier After it is rebased, could you please approve the GitHub Actions workflow so we can see if the tests pass?

jschneier:
> I suddenly got quite interested in this pull request when I read the following note from my colleague this morning about our attempt to upgrade from Django v4.1 to v4.2...
> 
> > I haven't seen any breakages in displaying images from S3, but when trying to download a resource, it raises a 500 error: '`S3Storage`' object has no attribute '`modified_time`'.
> 
> @TimoGuenther Would it be possible to please rebase this PR given all the recent changes to the repo? @jschneier After it is rebased, could you please approve the GitHub Actions workflow so we can see if the tests pass?

The method `modified_time` was removed from Django several years ago so I removed support for it in this past version, properly it should have been migrated to `get_modified_time`.

jschneier:
Just opened #1298 to mention the removal.

PR #1246, opened 2023-05-16
enables passing `security_token` to constructor

smirolo:
Without this line, `BaseStorage.__init__` raises an `ImproperlyConfigured` exception which prevents passing security_token as a parameter to `S3Boto3Storage.__init__`.

We use django-storages to manage a multi-tenant / multi-S3-bucket / multi-aws-account service. As a result, multiple `S3Boto3Storage` classes get instantiated with the proper information (access_key, secret_key and security_token).

This line fixes our problem. I am not sure if there was a reason why `security_token` is not
listed in default settings and/or if as a result it breaks some other use cases.


jschneier:
Thanks. I guess with this we can purge out line 259 and probably a bunch of stuff around there. Can you just start with removing that and I will send another cleanup PR.

jschneier:
Never mind, it doesn't matter so I will take this now.

PR #1249, opened 2023-05-19
check if underlying file object is closed in closed method

jordancrane:
Currently it appears that `S3Boto3StorageFile` uses the `closed` property defined by `FileProxyMixin`, which checks the condition `not self.file or self.file.closed` ([source](https://github.com/django/django/blob/main/django/core/files/utils.py#L54)). Since `self.file` in `S3Boto3StorageFile` is a property that delegates to `_get_file`, this check reopens the file. Thus `closed` always returns `False`.

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -131,6 +131,10 @@ class S3Boto3StorageFile(CompressedFileMixin, File):
     @property
     def size(self):
         return self.obj.content_length
+    
+    @property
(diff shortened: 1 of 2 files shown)


jschneier:
Thanks, can you fix the linting issue.

Also, let's add the beginnings of a test class for `S3Boto3StorageFile` that covers this fix.

jschneier:
Thanks!

PR #1252, opened 2023-05-22
Add FileSystemOverwriteStorage

craigds:
Improved implementation of #1209

# Background: 

This ticket is going in loops:

Refer https://code.djangoproject.com/ticket/28144

I guess django-storages apparently _did_ once have this class, and it was removed. 

Per that above ticket, Django later added a fix to make it _possible_ to override the not-overwriting behaviour (an `OS_OPEN_FLAGS` attribute)

Looks like there was (a LONG time ago) some discussion about including this feature in Django: https://code.djangoproject.com/ticket/4339 (it was closed wontfix due to the trivial nature of a custom implementation)

So, that puts us back here - we need somewhere to store that custom implementation, and django-storages seems like the right project for the job...

I've added a test and some basic docs (I don't think much is required, since there's no special configuration etc, unlike the other backends)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,17 @@
 django-storages CHANGELOG
 =========================
 
+UNRELEASED
+**********
+
+Filesystem
+----------
+
+Add new ``FileSystemOverwriteStorage`` backend - similar to Django's existing filesystem storage,
+but overwrites existing files instead of always saving with unique names. (`#1252`_)
+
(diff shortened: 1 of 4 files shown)


craigds:
any chance of a review here?

jschneier:
I'm happy to accept this but am a bit concerned about drift from core.

As an example `get_available_name` has already drifted because `validte_file_name` is not called.

jschneier:
This one seems likely to land into upstream so going to close here.

See: https://github.com/django/django/pull/18020

PR #1253, opened 2023-05-24
Close SSH connection of SFTP storage

### LINKED ISSUE #1234: Unclosed SFTP connection
Consider a one-off use of `SFTPStorage` as you might for example see in a Celery task:

```py
from storages.backends.sftpstorage import SFTPStorage

SFTPStorage(...).listdir("")
```

Unfortunately, the underlying SSH connection might stay open and continue using resources. [Paramiko's documentation of `close`](https://docs.paramiko.org/en/latest/api/client.html#paramiko.client.SSHClient.close) warns about this:

> Paramiko registers garbage collection hooks that will try to automatically close connections for you, but this is not presently reliable. Failure to explicitly close your client after use may lead to end-of-process hangs!

Workaround:

```py
from paramiko.util import ClosingContextManager
from storages.backends.sftpstorage import SFTPStorage

class ClosingSFTPStorage(SFTPStorage, ClosingContextManager):
    def close():
        self._ssh.close()

with ClosingSFTPStorage(...) as sftp:
    sftp.listdir("")
```

TimoGuenther:
Closes #1234 by making the SFTP storage close the underlying SSH connection after being used as a context manager.

This is done by adding Paramiko's convenient `ClosingContextManager` to the SFTP storage and implementing its hook for closing. I could have implemented the context manager manually, but Paramiko is required for SFTP, anyway.

In doing so, this bumps the minimum Paramiko version from 1.10.0 to 1.15.0, which is the version that [introduced `ClosingContextManager`](https://github.com/paramiko/paramiko/commit/979838040f9ff2f52e369812be87f0c7c9c1a0db) almost 9 years ago.

### DIFF
--- a/docs/backends/sftp.rst
+++ b/docs/backends/sftp.rst
@@ -71,3 +71,18 @@ static files) to use the sftp backend::
 .. _`paramiko SSHClient.connect() documentation`: http://docs.paramiko.org/en/latest/api/client.html#paramiko.client.SSHClient.connect
 
 .. _`Python os.chmod documentation`: http://docs.python.org/library/os.html#os.chmod
+
+
+Standalone Use
+--------------
+
+If you intend to construct a storage instance not through Django but directly,
+use the storage instance as a context manager to make sure the underlying SSH
+connection is closed after use and no longer consumes resources.
+
+.. code-block:: python
+
+    from storages.backends.sftpstorage import SFTPStorage
+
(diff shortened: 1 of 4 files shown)


GT4C:
Hello, I'm trying to use this Feature in my project but i can´t implement the ClosingContextManager in my models.py
Can you give me an example of how to use it?

PR #1254, opened 2023-05-24
Fix subheading weight

TimoGuenther:
Before this change, `References` used to show up in the list of backends at the documentation landing page.


jschneier:
Thanks.

PR #1257, opened 2023-06-27
ci: Drop Python 3.7 which is end-of-life

cclauss:
https://devguide.python.org/versions/ -- **On hold** until Django 3.2 Long Term Support ends in April 2024.
* https://docs.djangoproject.com/en/4.2/releases/4.2


jschneier:
Sorry, Django 3.2 will be supported another ~6 months which includes Python 3.7

Issue #917, opened 2020-08-18
Uploading .gz files with Content-Type: application/gzip and no Content-Encoding


normangilmore:
This is regarding `S3Boto3Storage.get_object_parameters` which was added in 1.9, but since the next major version is breaking changes anyway, I have a use-case for revising it.

This is somewhat of a follow-up to the double compression issue #451 which was addressed in Django-Storages 1.9.

My use case is that I want to store files with the filenames like "hello.csv.gz" and I want them to be downloaded from S3 as archives, not decompressed by the user-agent. I am compressing them as I stream out the CSV, so they are already compressed when they reach Django-Storages.

To download as archives, one must set Content-Type: application/gzip and NOT set Content-Encoding.

What I have finally figured out is that mimetypes.guess_type recommends the wrong Content-Type and Content-Encoding
for .gz, if you want to download it as an archive.

> mimetypes.guess_type('hello.csv.gz')
('text/csv', 'gzip')
mimetypes.guess_type('hello.gz')
(None, 'gzip')

This is no good if guess_type is being used to set Content-Type and Content-Encoding headers, and you want the file to download as-is without the browser de-compressing it or modifying the extension.

To download as-is, as an archive, guess_type should work the same as it does for .zip
> mimetypes.guess_type('hello.csv.zip')
('application/zip', None)

So for this use-case, .gz should return:
> mimetypes.guess_type('hello.csv.gz')
('application/gzip', None)

It is actually easy to force mimetypes.guess_type to do the right thing in this case:
> mimetypes.encodings_map
{'.gz': 'gzip', '.Z': 'compress', '.bz2': 'bzip2', '.xz': 'xz'}
del mimetypes.encodings_map['.gz']
mimetypes.guess_type('hello.csv.gz')
('application/gzip', None)

Clearly this would be a sketchy patch. I guess the Pythonic way would be
to instantiate the MimeTypes class with your desired data.
See also:
https://www.iana.org/assignments/media-types/media-types.xhtml
https://tools.ietf.org/html/rfc6713

n.b. The non-standard .gzip extension returns:
> mimetypes.guess_type('hello.csv.gzip')
(None, None)

DjangoStorages uses mimetypes.guess_type to set Content-Type and Content-Encoding headesr and we need to intercept this wrong guess_type if we want our .csv.gz files to download properly as archives.

Version 1.9.1 of DjangoStorages introduces a method `get_object_parameters`
The method can be overridden to return parameters that are merged with the 
[...]

PR #1261, opened 2023-07-14
override mimetypes guess for gzip in S3Boto3Storage

dennisvang:
Basically, gzip files that have been uploaded to S3 using django-storages are extracted automatically upon subsequent download.

This behavior is not always desired (is it ever?) but currently there is no way to prevent this on the django-storages side.

The proposed fix simply prevents the automatic extraction altogether by overriding the mimetype, but that may not be ideal.

That's why I'm hoping this PR will jumpstart the discussion on issue #917

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -473,6 +473,10 @@ class S3Boto3Storage(CompressStorageMixin, BaseStorage):
         params = {}
 
         _type, encoding = mimetypes.guess_type(name)
+        # override the mimetypes guess for .gz, see #917
+        if str(name).endswith('.gz'):
+            _type = 'application/gzip'
+            encoding = None
         content_type = getattr(content, 'content_type', None)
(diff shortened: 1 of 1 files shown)


jschneier:
Another option I came up with is the following:

[enc.patch](https://github.com/jschneier/django-storages/files/12447042/enc.patch)

jschneier:
That is, if `ContentType` is set do not touch `ContentEncoding` at all.

jschneier:
I quite like this solution because it mirrors how `ACL` is handled below, there are no special cases with unconditional overrides now.

jschneier:
Okay I put that up here: https://github.com/jschneier/django-storages/pull/1281

dennisvang:
> Okay I put that up here: #1281

@jschneier Thanks, this looks like a very nice solution. A clear improvement on the [previous proposal][1].

[1]: https://github.com/jschneier/django-storages/issues/917#issuecomment-683535791

Issue #1271, opened 2023-08-03
Add CloudFront settings to OPTIONS parameter of STORAGES in Django 4.2


PetrDlouhy:
With the introduction of Django 4.2, the storages.backends.s3boto3.S3Boto3Storage backend now accepts additional options for configuration. However, I noticed that there are currently no options to configure CloudFront settings directly through these storage options.

Currently it is possible to set `AWS_CLOUDFRONT_KEY_ID` and `AWS_CLOUDFRONT_KEY` which would have effect to all storages.

In my specific use case, I require the cloudfront_key_id and cloudfront_key to be set for the CloudFrontSigner. I used to configure this manually in the constructor of my custom S3Boto3Storage subclass, but I think it would be beneficial to have this as part of the official library.

I propose to extend the options for S3Boto3Storage to accept cloudfront_key_id and cloudfront_key. If these options are provided, S3Boto3Storage would create and store an instance of CloudFrontSigner with these credentials.

Example:

```python
'staticfiles': {
    'BACKEND': 'storages.backends.s3boto3.S3Boto3Storage',
    'OPTIONS': {
        'access_key': 'xxxxx',
        'secret_key': 'xxxxx',
        'bucket_name': 'my_bucket',
        'custom_domain': None,
        'default_acl': 'public-read',
        'endpoint_url': 'https://s3.amazonaws.com/my_bucket',
        'location': 'static',
        'region_name': 'us-west-2',
        'cloudfront_key_id': 'my_cloudfront_key_id',
        'cloudfront_key': 'my_cloudfront_key',
    }
}
```

I would appreciate it if you could consider adding this feature to the library. It would allow a more flexible configuration of S3Boto3Storage and CloudFront in Django 4.2 and later versions. I am willing to contribute to this feature if needed.

PetrDlouhy:
For now I am overriding the `S3Boto3Storage` like this:

```python
from botocore.signers import CloudFrontSigner
from storages.backends.s3boto3 import S3Boto3Storage


class CloudFrontS3Boto3Storage(S3Boto3Storage):
    def __init__(self, *args, **kwargs):
        cloudfront_key_id = kwargs.pop("cloudfront_key_id", None)
        cloudfront_key = kwargs.pop("cloudfront_key", None)

        super().__init__(*args, **kwargs)

        if cloudfront_key_id and cloudfront_key:
            self.cloudfront_signer = CloudFrontSigner(
[...]

PR #1274, opened 2023-08-14
S3Boto3Storage: add support for Cloudfront settings through OPTIONS

PetrDlouhy:
Add support for setting "cloudfront_key_id" and "cloudfront_key" through OPTIONS
 
 Addresses issue #1271

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -256,6 +256,9 @@ class S3Boto3Storage(CompressStorageMixin, BaseStorage):
     config = None
 
     def __init__(self, **settings):
+        cloudfront_key_id = settings.pop("cloudfront_key_id", None)
+        cloudfront_key = settings.pop("cloudfront_key", None)
+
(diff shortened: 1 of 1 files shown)


jschneier:
I see, the problem is that we're running logic in `get_default_settings` which is probably a mistake. Also, I believe this would break support of `AWS_CLOUDFRONT_KEY_ID` etc.

I think we should instead:

1. Remove the checking logic from within `get_default_settings` and move it to `__init__` just before the signer creation you added
2. Remove the creation of `cloudfront_signer` and stop returning it in the dict
3. Add `cloudfront_key_id` and `cloudfront_key` as return values in the dict
4. Update the creation logic to use `self.cloudfront_key` and `self.cloudfront_signer`

What do you think?

jschneier:
Oh I see, it doesn't break the current flow it just overrides it.

jschneier:
Yeah on second look i think this change looks good, thanks.

PR #1281, opened 2023-08-27
[s3] do not override a returned ContentType from get_object_parameters

### LINKED ISSUE #917: Uploading .gz files with Content-Type: application/gzip and no Content-Encoding

jschneier:
Closes #917

### DIFF
--- a/storages/backends/s3boto3.py
+++ b/storages/backends/s3boto3.py
@@ -481,17 +481,16 @@ class S3Boto3Storage(CompressStorageMixin, BaseStorage):
         return self.bucket.Object(name).content_length
 
     def _get_write_parameters(self, name, content=None):
-        params = {}
+        params = self.get_object_parameters(name)
 
-        _type, encoding = mimetypes.guess_type(name)
-        content_type = getattr(content, 'content_type', None)
-        content_type = content_type or _type or self.default_content_type
+        if 'ContentType' not in params:
+            _type, encoding = mimetypes.guess_type(name)
+            content_type = getattr(content, 'content_type', None)
+            content_type = content_type or _type or self.default_content_type
 
-        params['ContentType'] = content_type
-        if encoding:
-            params['ContentEncoding'] = encoding
-
-        params.update(self.get_object_parameters(name))
+            params['ContentType'] = content_type
+            if encoding:
+                params['ContentEncoding'] = encoding
 
         if 'ACL' not in params and self.default_acl:
             params['ACL'] = self.default_acl
--- a/tests/test_s3boto3.py
+++ b/tests/test_s3boto3.py
@@ -195,6 +195,25 @@ class S3Boto3StorageTests(TestCase):
             Config=self.storage.transfer_config
         )
 
+    def test_content_type_set_explicitly(self):
+        name = "test_file.gz"
+        content = ContentFile("data")
+
+        def get_object_parameters(name):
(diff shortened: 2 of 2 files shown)


normangilmore:
@jschneier @dennisvang - clean solution, wish I would have proposed it!

PR #1295, opened 2023-09-05
Lint Python code with ruff instead of flake8 and isort

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 600 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,6 @@ jobs:
 
       - name: Quality tests
         env:
-          TOXENV: flake8
+          TOXENV: ruff
         run: |
           tox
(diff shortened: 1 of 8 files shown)


jschneier:
How is the Interop with black? Taking that hit is on my list of things to do as well (double quotes and all)

jschneier:
Thanks.

PR #1302, opened 2023-09-14
Move S3 auth & cloudfront_signer config validation to init

gerrod3:
Moved the validation in `get_default_settings` to be inside `__init__` and check the final instantiated values, not just the values from settings. So even if the values in settings are incorrect, you can still create a S3 class by passing in the right values.

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -275,17 +275,23 @@ class S3Storage(CompressStorageMixin, BaseStorage):
     """
 
     default_content_type = "application/octet-stream"
-    # If config provided in init, signature_version and addressing_style settings/args
-    # are ignored.
+    # If config provided in subclass, signature_version and addressing_style
+    # settings/args are ignored.
     config = None
 
     def __init__(self, **settings):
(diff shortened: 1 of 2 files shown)


gerrod3:
@jschneier Wdyt about this refactor?

jschneier:
I almost pushed something like this myself. Can you make it explicit what situation you are solving for?

gerrod3:
I want to make sure that if you are instantiating your own storage classes and providing every possible value for it, it shouldn't be blocked by what is configured in settings. In our project, Pulp, we have a multi-tenancy feature that allows users to create their own domains with their own custom storage backends separate from the default storage. [0] These domains must not use or be affected by the default settings.

[0] https://docs.pulpproject.org/pulpcore/workflows/domains-multi-tenancy.html

terencehonles:
This change seems reasonable, but we were relying on `cloudfront_signer=None` which this change forces it to be truthy and a falsy value allowed disabling the signing but still serving via cloudfront. I fixed this in https://github.com/jschneier/django-storages/pull/1326

Issue #1022, opened 2021-06-09
S3 storage backend's `size` method raise `botocore.exceptions.ClientError` instead of FileNotFoundError


nl-emmanuellodovice:
When you call `default_storage.size(path_that_does_not_exists)` with `S3Boto3Storage` as the default storage backend, it will raise `botocore.exceptions.ClientError` instead of `FileNotFoundError`.

jschneier:
Let's say I wrap this but now some other client error is encountered, what should happen?

nl-emmanuellodovice:
Thanks for checking this. I am not a boto expert but I was thinking may be we can check the response for the exception and see if the code is 404 and only then raise the `FileNotFoundError`.

Something like
```
try:
    size = default_storage.size("does_not_exists.file")
except ClientError as e:
    if e.response.get("Error", {}).get("Code", None) == '404':
        raise FileNotFoundError("does_not_exists.file does not exist.")
    else:
        raise
```

PR #1309, opened 2023-09-28
Raise not exists error when attempting to get size for unexisting S3 file

sevdog:
As of now, attempting to get the file size from `S3Storage` always raises a `botocore.ClientError`, which is vague and may not fall under existing file-management related error handling.

Since may other methods in S3 storage already implement this kind of logic I have put the same in the `S3Storage.size` method. Now if the S3 error is a "not found" a more explanatory `FileNotFoundError` is raised instead.


jschneier:
Closes #1022

jschneier:
Thanks

PR #1313, opened 2023-09-29
[s3] raise ImproperlyConfigured if no bucket name is set

### LINKED ISSUE #1140: Not setting bucket name causes confusing error
I'm learning django-storages by starting with zero configuration and adding what's needed one bit at a time to see how it alters behavior.

When I configure credentials but not a default bucket, then make a request to the server, it fails with a `ValueError`.

This stops happening as soon as I either set `settings.AWS_STORAGE_BUCKET_NAME` or set `bucket_name` in my storage subclass.

It may be worth adding an explicit check for required settings in order to provide a relevant error message informing the user of where they're falling short.

Traceback:  https://gist.github.com/odigity/d73cf6338a5bb7908399cbe58d84572d

jschneier:
Closes #1140

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -288,6 +288,8 @@ class S3Storage(CompressStorageMixin, BaseStorage):
         super().__init__(**settings)
 
         check_location(self)
+        if not self.bucket_name:
+            raise ImproperlyConfigured("Setting a bucket name is required.")
 
         if (self.access_key or self.secret_key) and self.session_profile:
             raise ImproperlyConfigured(
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -956,6 +956,11 @@ class S3StorageTests(TestCase):
                 access_key="foo", secret_key="boo", session_profile="moo"
             )
 
+    def test_bucket_name_required(self):
+        with override_settings(AWS_STORAGE_BUCKET_NAME=None):
+            with self.assertRaises(ImproperlyConfigured):
+                s3.S3Storage()
+
 
 class S3StaticStorageTests(TestCase):
     def setUp(self):


banagale:
This change broke my testing setup in CI. 

I'm not disputing that this is a correct change to the package, only that it was a breaking change for how I was loose with s3 variable definitions in my CI.

A while back I made a choice on this project not to use S3 when running unit tests.

On my my local machine, running tests manually, I defined the bucket.

But in CI I was setting `AWS_STORAGE_BUCKET_NAME` to None and getting away with it.

Looking at it again, I see I've had to make a number of small hacks to handle not using S3 there. 

With this new error, it seems like maybe its time to clear the cruft and just let CI have at a valid `django-storages` with s3 setup.

Are people reading / writing to S3 in their unit tests? 

Could be I'm not configuring django-storages for non-s3 usage in my CI run.

mike667:
@jschneier Hello, I have same issue as @banagale. ImproperlyConfigured error broke my CI. I fixed it by set random bucket name in my CI, but is there a way to move this check in another place?
Thanks

banagale:
@mike667 Glad I'm not the only one. Can you give more detail on the context of how you're using `django-storages`?

Also, curious if you have a reason for needing S3 in your CI, is it part of unit testing?

This particular issue popped up due to a reliance on the package from a wagtail installation I did recently.

jschneier:
Hi in general I would expect that CI and Test would swap out the backend but I’ll see what I can do.

banagale:
> Hi in general I would expect that CI and Test would swap out the backend but I’ll see what I can do.

I think swapping out the backend for CI and test is the right way to go about it.

So, I don't know if a change is needed. The docs clearly say bucket names are required.

It could be that adding guidance on config for unit testing in the docs would be enough. Right now there are no results for Unit Tests searching the docs.  If helpful, I could file an issue on this and look at providing a PR.

jschneier:
I am reverting this commit and will release a new version as soon as this and the other pr land.

PR #1318, opened 2023-09-30
add explicit dependency on azure-core

dimbleby:
this project directly imports from azure-core, and should express that dependency.  1.13.0 is the version that introduced `azure.core.utils.parse_connection_string`

this is pedantry really, because azure-storage-blob already pulls azure-core as a sub-dependency


jschneier:
Thanks

PR #1326, opened 2023-10-16
[s3] fix disabling cloudfront signing with cloudfront_signer=None

terencehonles:
This change fixes disabling cloudfront signing by specifying "None" (or another falsy value). This allows multiple storage configs that are all distributed via cloudfront, but only some of the configs need to be signed.

x-ref: https://github.com/jschneier/django-storages/pull/1302


jschneier:
Thanks, can you add a note to the docs about this piece of functionality.

jschneier:
Looks great, thanks.

PR #1334, opened 2023-11-01
[s3] fix disabling cloudfront signing with class variable

terencehonles:
This change fixes setting ``cloudfront_signer=None`` as a class variable to disable cloudfront signing (or passing a custom implementation), and is a follow up to https://github.com/jschneier/django-storages/pull/1326


terencehonles:
I tried to quickly look through the documentation to confirm this pattern is expected, but it brings the behavior inline with  https://github.com/jschneier/django-storages/pull/1326 and also how this code previously behaved. An alternative for us is to overload the `__init__` and specify this manually, but hopefully this change is reasonable.

PR #1336, opened 2023-11-08
Fix `AWS_S3_*` environment variables lookup

### LINKED ISSUE #1335: `AWS_S3_*` environment variables are ignored
It is a regression probably from 705bdb7300642f0c42d823e4e35eb20356579389 in v1.14.

Now these variables are loaded only from Django settings, not from env.

https://github.com/jschneier/django-storages/blob/d0dce62a8a35bd4c75f77d389781424d4f705c07/storages/backends/s3.py#L355-L358

https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#authentication-settings

I think `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` still work because boto3 load them.

iurisilvio:
Fix #1335

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -26,6 +26,7 @@ from storages.utils import check_location
 from storages.utils import clean_name
 from storages.utils import get_available_overwrite_name
 from storages.utils import is_seekable
+from storages.utils import lookup_env
 from storages.utils import safe_join
 from storages.utils import setting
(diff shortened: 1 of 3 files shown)


jschneier:
Thanks for the patch, please fix the lint errors.

iurisilvio:
Fixed!

Issue #941, opened 2020-10-02
signed urls do not work in Cloud Run with django-storages


sww314:
Using Cloud Run the default credentials do not allow signing of urls. 
The error is swallowed in most use cases and the file filed just returns a None in serializers or in django admin. 

The error is confusing because everything works fine on the upload side. 
Even worse, if you use the same service account and run your container locally - it works fine. Since the credential is provided in the different manner. 

To recreate:
- setup Django project in Cloud Run with Media objects in stored using GCS.
- Create file in Django admin (the file field comes back as None, but the file is in the GCS bucket)
- On editing a file in Django admin, the error is displayed: 

```
Exception Type: | AttributeError
-- | --
you need a private key to sign credentials.the credentials you are currently using <class 'google.auth.compute_engine.credentials.Credentials'> just contains a token. see https://googleapis.dev/python/google-api-core/latest/auth.html#setting-up-a-service-account for more details.
/usr/local/lib/python3.8/site-packages/google/cloud/storage/_signing.py, line 51, in ensure_signed_credentials
```

I am still trying to figure out the best work around, but I wanted to add this in case anyone else runs into the error. 
This maybe a documentation update or a change to not swallow the error.

Prikers:
Just run on the very same issue with Google App Engine (standard). Took me a while as well to understand the error. Uploading a file works just fine so I thought it should not be a credential issue, but retrieving the file causes the error (at url signing invocation).

@sww314 did you find a nice work around / hack for this issue? Could I store the credentials in a file in Cloud Storage and add an environment variable to the `app.yaml` to pass the link to this file?

sww314:
I ended up storing the credentials in my container. 
I think solution may work as well, but I have not had a chance to try it yet:
https://stackoverflow.com/questions/64234214/how-to-generate-a-blob-signed-url-in-google-cloud-run

robcharlwood:
:heavy_plus_sign: :one: - I think we are experiencing the same issue.

danfairs:
Our workaround is as follows:

```
import datetime

from django.core.cache import cache
from django.utils.deconstruct import deconstructible

import google.auth
import google.auth.compute_engine
import google.auth.transport.requests
from djangoapps.core.context import get_context
# We'll take it on the chin if this moves
from google.cloud.storage.blob import _quote
from storages.backends.gcloud import GoogleCloudStorage
from storages.utils import clean_name


@deconstructible
[...]

PR #1340, opened 2023-11-27
fix(cred): Use in built library instead of forced cred

dark0dave:
Closes https://github.com/jschneier/django-storages/issues/941

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -61,6 +61,7 @@ For development use cases, or other instances outside Google infrastructure:
 
 Alternatively, you can use the setting ``credentials`` or ``GS_CREDENTIALS`` as described below.
 
+It is also now possible to use workload identity by providing the service account via ``GS_SA_SIGNING_EMAIL``.
 
 Settings
 ~~~~~~~~
@@ -219,3 +220,12 @@ Settings
   It supports `timedelta`, `datetime`, or `integer` seconds since epoch time.
 
   Note: The maximum value for this option is 7 days (604800 seconds) in version `v4` (See this `Github issue  <https://github.com/googleapis/python-storage/issues/456#issuecomment-856884993>`_)
+
+``sa_email`` or ``GS_SA_SIGNING_EMAIL``
+
+  default: ``''``
+
+  This is the signing email if it is not fetched from the credentials. Or if you wish to sign the signed urls with a different service_account.
+
+  As above please note that, Default Google Compute Engine (GCE) Service accounts are
+  `unable to sign urls <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ dropbox = [
   "dropbox>=7.2.1; python_version<'3.12'",
 ]
 google = [
-  "google-cloud-storage>=1.27",
+  "google-cloud-storage>=2.14",
 ]
 libcloud = [
   "apache-libcloud",
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -1,4 +1,5 @@
 import gzip
+import logging
 import io
 import mimetypes
 from datetime import timedelta
@@ -20,6 +21,8 @@ from storages.utils import setting
 from storages.utils import to_bytes
 
 try:
+    from google import auth
+    from google.auth.transport import requests
     from google.cloud.exceptions import NotFound
     from google.cloud.storage import Blob
     from google.cloud.storage import Client
@@ -34,6 +37,7 @@ except ImportError:
 
 CONTENT_ENCODING = "content_encoding"
 CONTENT_TYPE = "content_type"
+_LOGGER = logging.getLogger(__name__)
 
 
 class GoogleCloudFile(CompressedFileMixin, File):
@@ -142,12 +146,19 @@ class GoogleCloudStorage(BaseStorage):
             # roll over.
             "max_memory_size": setting("GS_MAX_MEMORY_SIZE", 0),
             "blob_chunk_size": setting("GS_BLOB_CHUNK_SIZE"),
+            "sa_email": setting("GS_SA_SIGNING_EMAIL")
         }
 
     @property
     def client(self):
+        credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
+        credentials.refresh(requests.Request())
+        if not hasattr(credentials, "service_account_email"):
+            credentials.service_account_email = self.sa_email
+        _LOGGER.debug(f"Signing email: {credentials.service_account_email}")
+        self.credentials = credentials
         if self._client is None:
-            self._client = Client(project=self.project_id, credentials=self.credentials)
+            self._client = Client(project=project_id, credentials=self.credentials)
         return self._client
 
     @property
@@ -318,19 +329,20 @@ class GoogleCloudStorage(BaseStorage):
                 storage_base_url=self.custom_endpoint,
                 quoted_name=_quote(name, safe=b"/~"),
             )
+        elif not self.custom_endpoint:
+            return blob.generate_signed_url(**self.signed_url_extra())
         else:
-            default_params = {
-                "bucket_bound_hostname": self.custom_endpoint,
-                "expiration": self.expiration,
-                "version": "v4",
-            }
-            params = parameters or {}
+            return blob.generate_signed_url(
+                api_access_endpoint=self.custom_endpoint,
+                **self.signed_url_extra()
+            )
 
-            for key, value in default_params.items():
-                if value and key not in params:
-                    params[key] = value
-
-            return blob.generate_signed_url(**params)
+    def signed_url_extra(self):
+        return {
+            "service_account_email": self.credentials.service_account_email,
+            "access_token": self.credentials.token,
+            "credentials": self.credentials,
+        }
 
     def get_available_name(self, name, max_length=None):
         name = clean_name(name)
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -1,5 +1,4 @@
 import gzip
-import logging
 import io
 import mimetypes
 from datetime import timedelta
@@ -37,7 +36,6 @@ except ImportError:
 
 CONTENT_ENCODING = "content_encoding"
 CONTENT_TYPE = "content_type"
-_LOGGER = logging.getLogger(__name__)
 
 
 class GoogleCloudFile(CompressedFileMixin, File):
@@ -151,14 +149,13 @@ class GoogleCloudStorage(BaseStorage):
 
     @property
     def client(self):
-        credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
-        credentials.refresh(requests.Request())
-        if not hasattr(credentials, "service_account_email"):
-            credentials.service_account_email = self.sa_email
-        _LOGGER.debug(f"Signing email: {credentials.service_account_email}")
-        self.credentials = credentials
         if self._client is None:
-            self._client = Client(project=project_id, credentials=self.credentials)
+            project_id, credentials = self.project_id, self.credentials
+            if project_id is None and credentials is None:
+                credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
+                if not hasattr(credentials, "service_account_email"):
+                    credentials.service_account_email = self.sa_email
+            self._client = Client(project=project_id, credentials=credentials)
         return self._client
 
     @property
@@ -329,20 +326,16 @@ class GoogleCloudStorage(BaseStorage):
                 storage_base_url=self.custom_endpoint,
                 quoted_name=_quote(name, safe=b"/~"),
             )
-        elif not self.custom_endpoint:
-            return blob.generate_signed_url(**self.signed_url_extra())
         else:
-            return blob.generate_signed_url(
-                api_access_endpoint=self.custom_endpoint,
-                **self.signed_url_extra()
-            )
-
-    def signed_url_extra(self):
-        return {
-            "service_account_email": self.credentials.service_account_email,
-            "access_token": self.credentials.token,
-            "credentials": self.credentials,
-        }
+            params = {
+                "service_account_email": self.credentials.service_account_email,
+                "access_token": self.credentials.token,
+                "credentials": self.credentials,
+                "expiration": self.expiration,
+            }
+            if self.custom_endpoint:
+                params["api_access_endpoint"] = self.custom_endpoint
+            return blob.generate_signed_url(**params)
 
     def get_available_name(self, name, max_length=None):
         name = clean_name(name)


storages/backends/gcloud.py
```diff
@@ -123,7 +127,7 @@ def get_default_settings(self):
             "location": setting("GS_LOCATION", ""),
             "default_acl": setting("GS_DEFAULT_ACL"),
             "querystring_auth": setting("GS_QUERYSTRING_AUTH", True),
-            "expiration": setting("GS_EXPIRATION", timedelta(seconds=86400)),
+            "expiration": setting("GS_EXPIRATION", timedelta(seconds=31536000)),
```
jschneier, line 130: Why is this default being updated?

storages/backends/gcloud.py
```diff
@@ -142,12 +146,19 @@ def get_default_settings(self):
             # roll over.
             "max_memory_size": setting("GS_MAX_MEMORY_SIZE", 0),
             "blob_chunk_size": setting("GS_BLOB_CHUNK_SIZE"),
+            "sa_email": setting("GS_SA_SIGNING_EMAIL")
         }
 
     @property
     def client(self):
+        credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
```
jschneier, line 154: This now completely eliminates `self.project_id` as a setting, is that always valid?

jschneier (changes requested):
Thanks, have a couple questions to start; am interested in getting this in for sure.

jschneier:
Please also update the documentation once the design is finalized

storages/backends/gcloud.py
dark0dave: Yes, otherwise the credentials might not match the project your request from.

jschneier: Mostly I'm asking about backwards compat here.

storages/backends/gcloud.py
```diff
@@ -142,12 +146,19 @@ def get_default_settings(self):
...
+        credentials.refresh(requests.Request())
+        if not hasattr(credentials, "service_account_email"):
+            credentials.service_account_email = self.sa_email
+        _LOGGER.debug(f"Signing email: {credentials.service_account_email}")
```
jschneier, line 158: This will log every time this uncached property is touched.

dark0dave: sure removed.

storages/backends/gcloud.py
dark0dave: Should work fine. Basically you fetch creds from gcp server based on google auth patterns.

dark0dave: auth.default does not accept project anyways

dark0dave:
> Please also update the documentation once the design is finalized

Can do.

storages/backends/gcloud.py
```diff
@@ -34,6 +37,7 @@
 
 CONTENT_ENCODING = "content_encoding"
 CONTENT_TYPE = "content_type"
+_LOGGER = logging.getLogger(__name__)
```
jschneier, line 40: Please remove the logger definition & import if it's unused.

dark0dave: done.

storages/backends/gcloud.py
jschneier: Don't suppose you have any documentation link handy? Just trying to think through all the ramifications other than the obvious one that we are adding a network call.

You can see that as originally written we cached the client. With this update we do the work but don't update the cached item, which is inconsistent.

dark0dave:
@jschneier thank you for taking the time to review this it is much appreciated. We are using my fork in prod and I would prefer to use the mainline package. Is there anything further needed.

Thanks again.

storages/backends/gcloud.py
dark0dave: I can do a check if its empty, and if its provided we can use it. I don't want to break backwards compat.

dark0dave: So if its not provided now it will use the one from the envs

dark0dave: > Don't suppose you have any documentation link handy? Just trying to think through all the ramifications other than the obvious one that we are adding a network call.

The docs for this were a mess. I had to read source. Will link.

dark0dave: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/_default.py#L541

dark0dave: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/_default.py#L575 <- this is the relevant line

storages/backends/gcloud.py
```diff
@@ -142,10 +144,18 @@ def get_default_settings(self):
...
             "blob_chunk_size": setting("GS_BLOB_CHUNK_SIZE"),
+            "sa_email": setting("GS_SA_SIGNING_EMAIL")
         }
 
     @property
     def client(self):
+        credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
+        if self.project_id is None:
+            self.project_id = project_id
+        credentials.refresh(requests.Request())
```
jschneier, line 155: Do we need to refresh the credentials every time we access `client`? Basically I'm wondering if we can follow the example already there where we only create the client once. The only reason not to guard against this in what you've got is the refresh. Without it we end up with something such as:

```python
if self._client is None:
    project_id, credentials = self.project_id, self.credentials
    if project_id is None and credentials is None:
       credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
       if not hasattr(credentials, "service_account_email"):
           credentials.service_account_email = self.sa_email
      self._client = Client(project=project, credentials=credentials)
```

storages/backends/gcloud.py
```diff
@@ -142,10 +144,18 @@ def get_default_settings(self):
...
+        if not hasattr(credentials, "service_account_email"):
+            credentials.service_account_email = self.sa_email
```
jschneier, line 157: What determines if the `hasattr` check fails?

storages/backends/gcloud.py
```diff
@@ -142,10 +144,18 @@ def get_default_settings(self):
...
+            "sa_email": setting("GS_SA_SIGNING_EMAIL")
         }
 
     @property
     def client(self):
+        credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
+        if self.project_id is None:
+            self.project_id = project_id
+        credentials.refresh(requests.Request())
+        if not hasattr(credentials, "service_account_email"):
```
jschneier, line 156: What determines if the `hasattr` check fails? Also, if it fails is it an error not to see `sa_email`?

storages/backends/gcloud.py
```diff
@@ -318,19 +328,24 @@ def url(self, name, parameters=None):
                 storage_base_url=self.custom_endpoint,
                 quoted_name=_quote(name, safe=b"/~"),
             )
+        elif not self.custom_endpoint:
```
jschneier, line 331: We don't need the new method or this chain, might be better written as follows. Before that, I am wondering if we need to always include the new parameters such as "access_token" in all of the requests in all cases?

```python
else:
    params = {
            "service_account_email": self.credentials.service_account_email,
            "access_token": self.credentials.token,
            "credentials": self.credentials,
            "expiration": self.expiration,
    }
    if self.custom_endpoint:
        params["api_access_endpoint"] = self.custom_endpoint
    return blob.generate_signed_url(**params)
```

jschneier: I see from the docs that if we don't pass the credentials the client tries to obtain it anyway which means we probably should pass it in. The other two I'm less sure about though since we've already got the credentials there.

storages/backends/gcloud.py
dark0dave: Happy to add this. I just wanted to refresh the client if it has been a long time since the user called it.

docs/backends/gcloud.rst
```diff
@@ -219,3 +219,12 @@ Settings
...
   Note: The maximum value for this option is 7 days (604800 seconds) in version `v4` (See this `Github issue  <https://github.com/googleapis/python-storage/issues/456#issuecomment-856884993>`_)
+
+``sa_email`` or ``GS_SA_SIGNING_EMAIL``
+
+  default: ``''``
+
+  This is the signing email if it is not fetched from the credentials. Or if you wish to sign the signed urls with a different service_account.
+
+  As above please note that, Default Google Compute Engine (GCE) Service accounts are
+  `unable to sign urls <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
```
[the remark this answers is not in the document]
dark0dave: Will update

storages/backends/gcloud.py
dark0dave: Sure I have updated. Yeah I think we should always pass it if we can.

storages/backends/gcloud.py
dark0dave: In some situations (workload identity) if you try to use the default service account the service account email can be missing. In this situation its best to provide the sa_email so you can sign urls.

dark0dave:
@jschneier  I have updated

storages/backends/gcloud.py
```diff
@@ -319,17 +326,14 @@ def url(self, name, parameters=None):
...
-                "version": "v4",
             }
-            params = parameters or {}
-
-            for key, value in default_params.items():
-                if value and key not in params:
-                    params[key] = value
-
+            if self.custom_endpoint:
+                params["api_access_endpoint"] = self.custom_endpoint
```
jschneier, line 336: This set of changes reverts using the `parameters` bit of `url()`. Sorry if there is no test currently covering this.

storages/backends/gcloud.py
jschneier: Sorry, you are saying that this *must* be set in some situation? Can you add that to the documentation.

jschneier (changes requested):

pyproject.toml
```diff
@@ -54,7 +54,7 @@ dropbox = [
   "dropbox>=7.2.1; python_version<'3.12'",
 ]
 google = [
-  "google-cloud-storage>=1.27",
+  "google-cloud-storage>=2.14",
```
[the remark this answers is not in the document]
dark0dave: No. Buts generally good to bump the library.

storages/backends/gcloud.py
```diff
@@ -142,13 +144,18 @@ def get_default_settings(self):
...
         }
 
     @property
     def client(self):
         if self._client is None:
-            self._client = Client(project=self.project_id, credentials=self.credentials)
-        return self._client
+            project_id, credentials = self.project_id, self.credentials
+            if project_id is None and credentials is None:
+            credentials, project_id = auth.default(scopes=['https://www.googleapis.com/auth/cloud-platform'])
```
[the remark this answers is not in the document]
dark0dave: Fixed.

pyproject.toml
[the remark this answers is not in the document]
jschneier: It’s good to bump your application requirements but as a library I’m trying to maintain as much flexibility for users as possible. If we need a new feature then bumping is worthwhile.

A major version bump in particular could induce work and compat issues to users.

Yohannes-B:
Any news on this? I am also trying to implement this. The current state of the PR looks sound. Is there anything missing that needs to be added/reviewed? Would love to help with any new asks if any. Currently our code is in MVP stage and we have a class that inherits from `GoogleCloudStorage` with the logic in this PR

jschneier:
> Any news on this? I am also trying to implement this. The current state of the PR looks sound. Is there anything missing that needs to be added/reviewed? Would love to help with any new asks if any. Currently our code is in MVP stage and we have a class that inherits from `GoogleCloudStorage` with the logic in this PR

Yeah I actually have been working on this one a bit myself but could very much use the help, especially for testing with Google Cloud. There are a few things:

1. As mentioned in my comment this change breaks the `parameters` support for `url`
2. We shouldn't bump the library required version for no reason
3. The parameters passed in `url()` were changed, I'd like confirmation that this changes NOTHING about backwards compat
4. I don't feel we have enough documentation about e.g when SA_EMAIL must be set, it'd be good to get some comprehensive testing using Google Cloud
5. We should change the variable name to `GA_SA_EMAIL` to be consistent with `sa_email`

yohannes15:
@jschneier what is the protocol here when someone has made PR already and I want to add some of the finishing touches. Do I fork @dark0dave:master and PR to that branch after adding my changes. Or should I fork django-storages:master and add this logic and rest of implementation there as a new separate PR?

Btw prev comment was from my work account. Going to be deleting that comment and communicating using this.

yohannes15:
>     1. As mentioned in my comment this change breaks the `parameters` support for `url`
> 
>     2. We shouldn't bump the library required version for no reason
> 
>     3. The parameters passed in `url()` were changed, I'd like confirmation that this changes NOTHING about backwards compat
> 
>     4. I don't feel we have enough documentation about e.g when SA_EMAIL must be set, it'd be good to get some comprehensive testing using Google Cloud
> 
>     5. We should change the variable name to `GA_SA_EMAIL` to be consistent with `sa_email`

1. Agree the parameters support shouldn't be removed. I still see the value in it even with the current state of the PR. There are kwargs that aren't in the GoogleCloudStorage settings that are needed such as content_type/api_access_endpoint... I think I have a solution that is backwards compatible with that.
2. Agree
3. Agree
4. According to the generate_signed_url functions docs and tests, my thinking is that the GS_SA_EMAIL (sa_email) will be what service_account_email would be. In 
 Here is the source code i'm referring to:
- https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/_signing.py#L279 
- https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/_signing.py#L456
It looks like both v2 and v4 both use the same logic in terms of checking for service_account_email and access_token before defaulting to credentials.signer_email
- https://googleapis.dev/python/google-auth/latest/reference/google.auth.credentials.html#google.auth.credentials.Signing
- https://github.com/googleapis/python-storage/pull/356 
I will look at the tests and try to see what tests would be good to make sure of backwards compatibility.
5. Agree

yohannes15:
https://github.com/googleapis/google-auth-library-python/issues/50

This issue which has been open for ages is really the concern I think here that is making our logic ugly. There are different complaints about why google doesn't automatically infer credentials when signing urls. https://github.com/googleapis/google-auth-library-python/issues/50#issuecomment-1139904099

storages/backends/gcloud.py
```diff
@@ -319,17 +327,14 @@ def url(self, name, parameters=None):
                 quoted_name=_quote(name, safe=b"/~"),
             )
         else:
-            default_params = {
-                "bucket_bound_hostname": self.custom_endpoint,
+            params = {
+                "service_account_email": self.credentials.service_account_email,
+                "access_token": self.credentials.token,
```
yohannes15, line 332: currently self.credentials throws a attribute error here as well since self.credentials hasn't been changed. the credentials that were set in the client @property doesn't flow over here. Might have to either set the self.credentials in there, or use self._client._credentials. 1st one seems ugly in setting a instance variable in an other property, 2nd option uses private instance variable. There is probably a 3rd option but that might involve some additional clean up

jschneier:
> @jschneier what is the protocol here when someone has made PR already and I want to add some of the finishing touches. Do I fork dark0dave:master and PR to that branch after adding my changes. Or should I fork django-storages:master and add this logic and rest of implementation there as a new separate PR?

Generally a new PR. As the maintainer I can push to it but generally only do that for minor things.

For (1) above, I added tests in #1408 so that when this is rebased, they will break.

For the rest I am happy to accept the findings, just mentioning what I think are the issues with the current version & why it hasn't merged. I trust you to make sure it's all buttoned up in the end!

jschneier:
Superseded by #1427.

PR #1358, opened 2024-02-27
Azure: raise FileNotFoundError when requested blob is missing from storage

alxbridge:
This PR implements behaviour in `AzureStorage._open` similar to that [already found in](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3.py#L475) `S3Storage._open`, raising a `FileNotFoundError` if the blob doesn't exist in the storage.

### DIFF
--- a/tests/test_azure.py
+++ b/tests/test_azure.py
@@ -502,3 +502,13 @@ class AzureStorageTest(TestCase):
                 "https://foo_name.blob.core.windows.net",
                 credential={"account_name": "foo_name", "account_key": "foo_key"},
             )
+
+    def test_missing_file_handling(self):
+        client_mock = mock.MagicMock()
+        client_mock.exists.side_effect = [True, False]
+        self.storage._client.get_blob_client.return_value = client_mock
+
+        file_that_exists = self.storage._open('file-that-exists')
+        self.assertIsInstance(file_that_exists, azure_storage.AzureStorageFile)
+
+        self.assertRaises(FileNotFoundError, self.storage._open, 'file-that-does-not-exist')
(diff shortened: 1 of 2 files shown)


jschneier:
Thanks, makes sense to fix this.

Any idea about doing the same kind of RAII-style handling in init of `AzureStorageFile` rather than calling `.exists()`? Might be able to save an API call that way (which costs money).

tests/test_azure.py
```diff
@@ -502,3 +502,13 @@ def test_blobserviceclient_no_custom_domain(self):
                 "https://foo_name.blob.core.windows.net",
                 credential={"account_name": "foo_name", "account_key": "foo_key"},
             )
+
+    def test_missing_file_handling(self):
+        client_mock = mock.MagicMock()
+        client_mock.exists.side_effect = [True, False]
+        self.storage._client.get_blob_client.return_value = client_mock
+
+        file_that_exists = self.storage._open('file-that-exists')
```
jschneier, line 511: Can you change this to use `open()` as that is the public API. Also to use a `with` context handler.

alxbridge:
> Thanks, makes sense to fix this.
> 
> Any idea about doing the same kind of RAII-style handling in init of `AzureStorageFile` rather than calling `.exists()`? Might be able to save an API call that way (which costs money).

I'm not sure I understand the existing code / Azure SDK well enough to say, tbh. It looks like where that's done in the `S3File` class, it's conditional on the file being opened in read mode, but the init for `AzureStorageFile` isn't currently passed similar information on the intended mode for the file.

tests/test_azure.py
alxbridge: I've just pushed commits to make these changes

jschneier (changes requested):
This will break since #1422. My suggestion is to do something similar to S3 and reference `self.file` in the constructor of AzureFile since that will properly try to download the file.

This version I think also will break if a file is opened for writing.

PR #1359, opened 2024-03-04
Set a default size when using `SpooledTemporaryFile`

RealOrangeOne:
Leaving `0` as the default means the file never gets rolled over to disk, resulting in higher memory usage than expected (or needed).

S3 and GCS both used `0` for their default value. Now, they use `FILE_UPLOAD_MAX_MEMORY_SIZE` (2.5MB by default). This is significantly better than `0`, but still allows configuration as needed.

Azure already set a default of 2MB, which is better, but now brought in line with S3 and GCS. Dropbox previously didn't allow configuration, so an additional setting was added.

### DIFF
--- a/docs/backends/amazon-S3.rst
+++ b/docs/backends/amazon-S3.rst
@@ -91,7 +91,7 @@ Settings
 
 ``max_memory_size`` or ``AWS_S3_MAX_MEMORY_SIZE``
 
-  Default: ``0`` i.e do not roll over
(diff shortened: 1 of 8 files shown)


jschneier:
This is a breaking change but I think a very reasonable one to follow upstream, thanks.

jschneier:
Everything should go green if you rebase now

RealOrangeOne:
@jschneier over to you :smile:

PR #1362, opened 2024-03-11
Don't look at user's AWS config when running tests

craigds:
If you set up a different config file locally and then run tests, the tests easily fail, and who knows what it's doing with user's credentials :/


jschneier:
Thanks, makes sense. Please rebase.

Issue #1363, opened 2024-03-15
[Feature] Distinct SFTP_BASE_URL and [Bugfix] sftp.exists() works incorrectly when using root_path


Toruitas:
[Feature]
I'm using Django Storages for both S3 and SFTP, so the fact they both use MEDIA_URL was problematic. 

I've added a new setting SFTP_BASE_URL which will be preferred. If it doesn't exist, MEDIA_URL will be used. 

```
# sftpstorage.py
def get_default_settings(self):
        return {
            ...
            "base_url": setting("SFTP_BASE_URL") or setting("MEDIA_URL"),
        }
```

And to set it in 
```
# django settings.py
SFTP_BASE_URL = os.getenv('SFTP_BASE_URL')  # use this first
MEDIA_URL= os.getenv('MEDIA_URL')  # use this second
```

[Bug]

Additionally, in implementing the SFTP storage with a root path, the SFTPStorage.exists() would always prepend the root path to what it was checking.

If the root path was `uploads`, it would check for the existence of `uploads/uploads` rather than just `uploads` when going through the recursive `_mkdir()` calls during a `_save()`.

This has been fixed by removing the use of self._remote_path(name) within SFTPStorage.exists(), and just returning True if name is falsy.

While in my manual testing in a Django project this works fine both with and without a root path, **I have not managed to get tox to run**. If someone would help me getting tox up and running, I can add a test for it.

Toruitas:
@jschneier  Additional details about the bugfix. I've just added a couple prints at the start of sftpstorage.py's exists method. This shows the process as it starts with _save() and the recursive _mkdir() calls.

I hadn't tried the pre-fix code without the root_path earlier, but I see there's a recursion error there, too. 

As I see it, there are 2 possible places to address the issue. Either in exists() or in _mkdir(). 


```

def exists(self, name):
        print(f"exists: Checking if '{name}' exists")
        print(f"exists: Statting {self._remote_path(name)}")
        try:
            self.sftp.stat(self._remote_path(name))
            return True
        except FileNotFoundError:
            return False
```

Scenario 1: 
root_path: `""` Using the default here
file_path: `file.txt` Not placing in a directory. 
sftp server initial directory: `/`

Outputs:
```
exists: Checking if 'file.txt' exists
exists: Statting file.txt
exists: Checking if '' exists
exists: Statting
exists: Checking if '' exists
exists: Statting
exists: Checking if '' exists
exists: Statting
Eventual recursion error. 
```

[...]

PR #1364, opened 2024-03-15
SFTP_BASE_URL and sftp.exists fix. #1363

Toruitas:
PR for issue: https://github.com/jschneier/django-storages/issues/1363

Adds: SFTP_BASE_URL
Fixes: SFTPStorage.exists()

Plus a bit of logic so we only check for a parent directory's existence if there's a parent path provided.

I haven't managed to get tox to run, so if I could get some help with that, I can write a test or two for this.

### DIFF
--- a/docs/backends/sftp.rst
+++ b/docs/backends/sftp.rst
@@ -117,7 +117,7 @@ Settings
 
   Absolute path of know host file, if it isn't set ``"~/.ssh/known_hosts"`` will be used.
 
-``base_url``
+``base_url`` or ``SFTP_BASE_URL``
 
   Default: Django ``MEDIA_URL`` setting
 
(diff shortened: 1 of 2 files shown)


jschneier:
Hi and thanks for the PR. If you want to separate out the `SFTP_BASE_URL` change I will merge that one straightaway.

I'm not sure I follow the other fixes, can you expand a bit more on what's going on? I looked at the ticket as well and have a glimpse but don't think I've managed to wrap my head all the way around it.

Toruitas:
I've added some more details to the ticket https://github.com/jschneier/django-storages/issues/1363

Toruitas:
And here's the PR for just the SFTP_BASE_URL https://github.com/jschneier/django-storages/pull/1368

jschneier:
Took #1372 over this.

Issue #1255, opened 2023-06-02
`collectstatic` is extremely slow with S3 Manifest


catgirlinspace:
hiya! I'm deploying my app on DigitalOcean App Platform. As part of the build step there, I run `python manage.py collectstatic --noinput`. Before switching to storing my static assets in a S3-compatible bucket this was instant, but now takes over 10 minutes when using the S3 manifest. The S3-compatible service I am using is Backblaze B2, although I wouldn't expect this to matter. 

## Configuration for my static files

`storages.py`
```py
class StaticStorage(S3ManifestStaticStorage):
    bucket_name = 'splashcat-static'
    custom_domain = 'static.splashcat.ink'
```

`settings.py`
```py
if not DEBUG:
    STORAGES = global_settings.STORAGES | {
        "default": {"BACKEND": "storages.backends.s3boto3.S3Boto3Storage"},
        "staticfiles": {"BACKEND": "splashcat.storages.StaticStorage"}
    }
```

DuncanTormey:
@catgirlinspace I am experiencing this too, did you find a solution?

DuncanTormey:
I ended up doing this. 
```
python -m cosmos.manage collectstatic --no-input & # background the static collection
gunicorn cosmos.cosmos_site.wsgi --config cosmos/gunicorn_conf.py
```

catgirlinspace:
@DuncanTormey hi, sorry for the late reply. my workaround for this was to just use collectstatic and serve it without S3 and then stick a CDN  in front.

davidmir:
+1

chorvitz-medicircle:
+1, it is making S3ManifestStaticStorage unusable.

txels:
Hey everyone... I have made some attempts to improve collectstatic via parallelisation until I realised the problem was that django storages is quite inefficient for that use case. Find a more detailed explanation in https://github.com/django/django/pull/19935#issuecomment-3649209685

Here is how I've sped up the process (from 2m to 20s after an initial run) in my static storage subclass. YMMV, let me know if that is helping your case as well:

```python
import logging
import os
from uuid import uuid4

from storages.backends.s3 import S3ManifestStaticStorage as S3StaticStorage
from storages.backends.s3 import S3Storage

logger = logging.getLogger(__name__)


DAYS = 3600 * 24


class StaticStorage(S3StaticStorage):
    location = "static"
    default_acl = "public-read"
    object_parameters = {
[...]

PR #1367, opened 2024-03-16
Cache S3 key metadata to avoid unnecessary calls especially with collectstatic

kupsum:
Relates to https://github.com/jschneier/django-storages/issues/1255

When running `collectstatic` S3 key check first the existence and the last modification time.  With this change the latter call is avoided. Previously last modification fetched the whole object and didn't use the metadata. This cause a lot of unnecessary traffic and thus making `collectstatic` slow.

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -3,6 +3,7 @@ import os
 import posixpath
 import tempfile
 import threading
+import typing
 import warnings
 from datetime import datetime
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


davidmir:
Is there any chance we can move this issue forward? :)

jschneier:
Does this work in a multi threaded environment or does the cache need a lock?

LeaveMyYard:
+1 for this PR, will really increase a speed of my deployment :)

@jschneier, from their docs cachetools are not thread-safe (https://cachetools.readthedocs.io/en/stable/)

PR #1368, opened 2024-03-18
SFTP_BASE_URL

Toruitas:
This PR adds SFTP_BASE_URL so that it can be used alongside other kinds of storages.


jschneier:
Thanks 👍

PR #1371, opened 2024-03-22
Explain how to use with Azurite (Azure Storage Emulator)

### LINKED ISSUE #1306: Using Azurite emulator of Azure storage
Hi, what is the best way to run django-storages with Azurite storage emulator?

After an update from a very old version (1.9.1), I have discovered that the AZURE_EMULATED_MODE setting is now ignored and the connection to the emulator no longer works out of the box.

Even when setting up the necessary custom domain and credentials, the client uses a hard-coded value `use_custom_domain=False`:
https://github.com/jschneier/django-storages/blob/c7662485b02890efe967ae216a648762f7d8a172/storages/backends/azure_storage.py#L184C1-L188C36
```python
class AzureStorage(BaseStorage):
    ...
    @property
    def service_client(self):
        if self._service_client is None:
            self._service_client = self._get_service_client(use_custom_domain=False)
        return self._service_client
```
There is custom_client which hard-codes use_custom_domain=True and that logic then works also with the emulator.

Based on that, I have two alternatives which work:
To override the client property individually:
```python
class AzureStaticStorage(AzureStorage):
    client = AzureStorage.custom_client
    ...
```
Or globally:
```python
if AZURE_EMULATED_MODE:
    from storages.backends.azure_storage import AzureStorage
    AzureStorage.client = AzureStorage.custom_client
```

This does not feel like the ideal solution.

Is there a reason for abandoning the emulated storage flag or not determining the use_custom_domain based on whether the custom domain is provided?

Otherwise, what would be the best way to tell Django to use the custom_client by default instead of client?

Thanks for the great library!

tonybaloney:
This PR explains how to use Azurite, the free local storage emulator for Azure Storage.

It also closes #1306

### DIFF
--- a/docs/backends/azure.rst
+++ b/docs/backends/azure.rst
@@ -147,6 +147,33 @@ Settings
 
   The api version to use.
 
+Using with Azurite (previously Azure Storage Emulator)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Azurite is a local emulator for Azure Storage accounts that emulates the API for Azure Blob storage and enables local testing and development without an Azure account, free of charge.
+
+To use the Azure Storage Emulator, you download and install it from the `Azurite page <https://learn.microsoft.com/azure/storage/common/storage-use-azurite>`_.
+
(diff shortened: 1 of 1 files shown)


jschneier:
Thanks!

PR #1372, opened 2024-03-22
[sftp] Check full path for existance in ._save() and ._mkdir()

### LINKED ISSUE #1363: [Feature] Distinct SFTP_BASE_URL and [Bugfix] sftp.exists() works incorrectly when using root_path

pjpetersik:
This pull request should fix #1363 and the recursion problem mentioned in the issue. 

The problem was related to the fact that the `.exists(self, name)` function assumes that the input is the `name` and not the full path. However, in `._save()` and `._mkdir()` a full path instead of the name was passed to the `.exists()`. 

To not interfere with the interface of the `.exists()` function, I introduced a `._exists(self, path)` function that uses the full path instead of the name as input.

I am aware that this is sort of a duplicate to what PR #1364 targets but I also wanted to "throw this solution into the ring" ;).

### DIFF
--- a/storages/backends/sftpstorage.py
+++ b/storages/backends/sftpstorage.py
@@ -118,7 +118,7 @@ class SFTPStorage(ClosingContextManager, BaseStorage):
         """Create directory, recursing up to create parent dirs if
         necessary."""
         parent = posixpath.dirname(path)
-        if not self.exists(parent):
+        if not self._path_exists(parent):
             self._mkdir(parent)
         self.sftp.mkdir(path)
 
@@ -134,7 +134,7 @@ class SFTPStorage(ClosingContextManager, BaseStorage):
             content.seek(0, os.SEEK_SET)
         path = self._remote_path(name)
         dirname = posixpath.dirname(path)
-        if not self.exists(dirname):
+        if not self._path_exists(dirname):
             self._mkdir(dirname)
 
         self.sftp.putfo(content, path)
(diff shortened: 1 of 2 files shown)


jschneier:
Thanks this makes sense and also maintains the proper `.exists()` functionality. Can you make a few updates:

1. Let's add a comment to `_exists` & `exists` to make it clear why both exists and let's move `._exists()` next to `.exist()`. I also want to change `_exists` to `_full_path_exists` or something.
2. Add a test or two that shows this is fixed

pjpetersik:
Thanks for the feedback. I added your suggestions. 

The important change in the test cases is the addition of the `root_path="root"` keyword in the `setUp()` method as well as the following assert
https://github.com/jschneier/django-storages/blob/52ff727f256de57e37191ce87b910398b014e766/tests/test_sftp.py#L98

Without the changes in `sftpstorage.py` this assert would fail since `mock_sftp.stat.call_args_list[0][0]` would return `('root/root/bar',)` because of the recursion bug.

jschneier:
Looks great, thanks for updating!

Toruitas:
Thanks for this!

PR #1381, opened 2024-04-21
[s3] Fix newline handling for text-mode files

### LINKED ISSUE #1351: `S3Boto3Storage.open(name, "r")` doesn't handle newlines properly
With S3Boto3Storage, opening a file in text mode (`"r"`) doesn't properly convert windows newlines (`"\r\n"`) to `"\n"` as you'd expect. Neither does it handle `\r` as `\n`

jschneier:
Rebased version of #1352 

Closes #1351

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -1,3 +1,4 @@
+import io
 import mimetypes
 import os
 import posixpath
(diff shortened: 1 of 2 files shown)


skim618:
Hello - I will have a look at the test failures sometime today/tomorrow

jschneier:
@skim618 Any updates?

skim618:
Sorry for the delay. Looking - it seems the issue is related to the changes made in python 3.11 for `SpooledTemporaryFile`.

`SpooledTemporaryFile` introduced new changes in 3.11 (https://docs.python.org/3/library/tempfile.html)
It now fully implements the `io.BufferedIOBase` and `io.TextIOBase` abstract base classes allowing the file
to be readable in the mode that it was specified (without accessing the underlying `._file` object). 

It seems like wrapping the underlying `._file` is forbidden for versions 3.11 and higher (results in the `ValueError`)
Having two have separate methods for different versions isn't the prettiest fix but should suffice.

PR #1386, opened 2024-04-25
[s3] Add ``client_config`` configuration setting

### LINKED ISSUE #1337: Threaded parallel uploads on S3
I'm uploading a lot of small files. I can thread the uploads using `ThreadPool` and it works great.
If I use more than 10 threads, I have the warning `Connection pool is full, discarding connection: my.s3.server.url. Connection pool size: 10 `.
I found 2 issues about this problem (https://github.com/jschneier/django-storages/issues/1106 and https://github.com/jschneier/django-storages/issues/610).

I tried to use `TransferConfig`, without success.
```python
AWS_S3_TRANSFER_CONFIG = TransferConfig(
    use_threads=True,
    max_concurrency=70,
)
```
I tried with `use_threads=False` and  `use_threads=True`.

What should I do to increase the max number of threads?

jschneier:
Closes #1337

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -35,7 +35,7 @@ try:
     import boto3.session
     import s3transfer.constants
     from boto3.s3.transfer import TransferConfig
-    from botocore.client import Config
+    from botocore.config import Config
     from botocore.exceptions import ClientError
     from botocore.signers import CloudFrontSigner
 except ImportError as e:
@@ -318,8 +318,17 @@ class S3Storage(CompressStorageMixin, BaseStorage):
         self._bucket = None
         self._connections = threading.local()
 
-        if not self.config:
-            self.config = Config(
+        if self.config is not None:
+            warnings.warn(
+                "The 'config' class property is deprecated and will be "
+                "removed in a future version. Use AWS_S3_CLIENT_CONFIG "
+                "to customize any of the botocore.config.Config parameters.",
+                DeprecationWarning,
+            )
+            self.client_config = self.config
+
+        if self.client_config is None:
+            self.client_config = Config(
                 s3={"addressing_style": self.addressing_style},
                 signature_version=self.signature_version,
                 proxies=self.proxies,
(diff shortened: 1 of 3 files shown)


storages/backends/s3.py
```diff
@@ -318,8 +318,17 @@ def __init__(self, **settings):
...
+        if self.config is not None:
+            warnings.warn(
+                "The 'config' class property is deprecated and will be "
+                "removed in a future version. Use AWS_S3_CLIENT_CONFIG "
+                "to customize any of the botocore.config.Config parameters.",
+                DeprecationWarning,
+            )
+            self.client_config = self.config
+
+        if self.client_config is None:
```
mgzwarrior, line 330: I'm thinking that there is an unintended logical change here that makes `self.client_config` a required field, defaulted to `None`

jschneier: See the logic [here](https://github.com/jschneier/django-storages/blob/master/storages/base.py#L6-L11).

mgzwarrior: I don't think I am following.  The `get_default_settings()` method [here](https://github.com/jschneier/django-storages/blob/8d7681fa22ae1532f03f144202340a8d124d3ccb/storages/base.py#L24) seems to return `{}` by default.

In our case, we have overridden this method as such:
```
    def get_default_settings(self):
        return {
            "access_key": setting(f'{self.setting_prefix}_ACCESS_KEY_ID'),
            "secret_key": setting(f'{self.setting_prefix}_SECRET_ACCESS_KEY'),
            ...,
        }
```

However, we don't have a setting for `client_config` because this is new.  This seems to be the root cause of the error: `object has no attribute 'client_config'`.  One solution I could try is to add a dummy attribute to the above method and default to `None`, but I shouldn't have to if the setting is not required, right?

jschneier: I don't provide backwards compat guarantees on `get_default_settings` as that is not a public API. If you want this to reliably work I'd recommend something like

```python
settings = super().get_default_settings()
...your changes here
```

PR #1388, opened 2024-04-28
Fix Incorrect URL Format in S3Storage.url Method

kudah99:
This PR should  solve  the issue with the ` url ` method of the S3Storage class generates URLs with an incorrect format by omitting the colon (":") after the protocol.

**PR solution** - The string formatter in the `url `method has been corrected to include the colon after the protocol. The previous formatter `"{}//{}/{}{}"` was changed to `"{}://{}/{}{}"` to properly format the URL.


jschneier:
Opened #1390 to clarify the docs so closing this one out.

PR #1394, opened 2024-05-04
Drop AWS_S3_ env variables in docs not working

gionn:
Spent 40 minutes to figure out the reason for an `Invalid token` raised when trying to upload a file on Scaleway object storage.


jschneier:
Sorry the fix is in master, release now.

gionn:
ref https://github.com/jschneier/django-storages/issues/1336

PR #1403, opened 2024-05-14
Fixing the problem that arises when using --clear in python manage.py collectstatic

Frodothedwarf:
There is a ongoing issue with python manage.py collectstatic --clear

Due to Django trying to walk through every dir we just return all the files from the top dir.
That way we don't get a error when Django is trying to walk the files.

Tested with --clear and without --clear.

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -268,6 +268,15 @@ class AzureStorage(BaseStorage):
         return super().get_available_name(name, max_length)
 
     def exists(self, name):
+        if not name:
+            try:
+                self.service_client.get_container_client(
+                    self.azure_container
+                )
+                return True
+            except:
+                return False
+
         blob_client = self.client.get_blob_client(self._get_valid_path(name))
         return blob_client.exists()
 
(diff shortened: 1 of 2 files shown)


Frodothedwarf:
@jschneier can you please look into this pr?

storages/backends/azure_storage.py
```diff
@@ -268,6 +268,15 @@ def get_available_name(self, name, max_length=_AZURE_NAME_MAX_LEN):
         return super().get_available_name(name, max_length)
 
     def exists(self, name):
+        if not name:
+            try:
+                self.service_client.get_container_client(
+                    self.azure_container
+                )
```
jschneier, line 275: I believe this can be replaced with `self.client`.

I think it's probably better to just `return True` here `if not name`.

storages/backends/azure_storage.py
```diff
@@ -390,16 +399,9 @@ def list_all(self, path=""):
...
-        dirs = set()
-        for name in self.list_all(path):
-            n = name[len(path) :]
-            if "/" in n:
-                dirs.add(n.split("/", 1)[0])
-            else:
-                files.append(n)
-        return list(dirs), files
+        
+        return [], self.list_all(path)
```
jschneier, line 407: So, while technically true, in practice all of these services use "/" as a directory separator.

What is the issue we are trying to solve here and what is the solution?

jschneier (changes requested):
Thanks for the PR, added a few comments. Please also add tests.

storages/backends/azure_storage.py
Frodothedwarf: The issue here is that Django tries to "walk" the directories, so first top level, next folder level and so on.
Azure does not support this type of walk, since they don't support listing directories, only blobs.
The only purpose for Djangos walk is to find all the files, but since we can just find all files with the list_all(), we just return all files and none directories.

storages/backends/azure_storage.py
Frodothedwarf: You are right, True should be returned if not name.

Frodothedwarf:
Changes has been implemented, and tests has been updated @jschneier

Frodothedwarf:
@jschneier workflow failed, whitespaced fixed.
https://github.com/jschneier/django-storages/pull/1403/commits/c8bb98e64c3e7627e7f65fc2a61f128bbbd60d20

jschneier (approved):

PR #1406, opened 2024-05-24
[s3] Fix wrong variable name

Mhs-220:
Use correct variables according to the code, for `ACCESS_KEY` and `SECRET_KEY`


jschneier (approved):
Thanks!

Issue #1413, opened 2024-06-27
AzureStorage does not allow creation of signed URLs for upload


CorentinGoodays:
**Summary**
The `AzureStorage` class in the `django-storages` module does not currently support generating signed URLs for uploads due to the hardcoded `BlobSasPermissions(read=True)` in the `url()` method. This limitation restricts users from creating SAS tokens with upload permissions.

**Environment**
django-storages version: All

**Steps to Reproduce**
Attempt to generate a signed URL for upload using the `AzureStorage` class.
Observe that the generated URL only has read permissions.

**Expected Behavior**
Users should be able to generate signed URLs with different permissions, including upload permissions.

**Actual Behavior**
The `url()` method only generates URLs with read permissions, as `BlobSasPermissions(read=True)` is hardcoded.

**Proposed Solution**
Allow the `url()` method to accept a `mode` parameter to specify different permissions. Here is a possible implementation:

```
diff --git a/storages/backends/azure_storage.py b/storages/backends/azure_storage.py
index 69b0213..435bc9a 100644
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -309,9 +309,10 @@ class AzureStorage(BaseStorage):
     # azure expects time in UTC
     return datetime.utcnow() + timedelta(seconds=expire)

-    def url(self, name, expire=None, parameters=None):
+    def url(self, name, expire=None, parameters=None, mode="r"):
     name = self._get_valid_path(name)
     params = parameters or {}
+    permission = BlobSasPermissions.from_string(mode)

     if expire is None:
         expire = self.expiration_secs
@@ -326,7 +327,7 @@ class AzureStorage(BaseStorage):
         name,
         account_key=self.account_key,
         user_delegation_key=user_delegation_key,
-        permission=BlobSasPermissions(read=True),
+        permission=permission,
         expiry=expiry,
         **params,
     )
```
**Testing**
I have tested this solution in my environment and confirmed that it allows generating signed URLs with different permissions, including upload permissions. Using these generated URLs, I successfully uploaded a file to Azure Storage.

PR #1414, opened 2024-06-27
[azure] Allow creation of signed URLs for upload

CorentinGoodays:
Allow creation of signed URLs for upload in AzureStorage (See #1413)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,11 +15,13 @@ Azure
 -----
 
 - Fix ``collectstatic --clear`` (`#1403`_)
+- Add ``mode`` kwarg to ``.url()`` to support creation of signed URLs for upload (`#1414`_)
 
 .. _#1399: https://github.com/jschneier/django-storages/pull/1399
 .. _#1381: https://github.com/jschneier/django-storages/pull/1381
 .. _#1402: https://github.com/jschneier/django-storages/pull/1402
 .. _#1403: https://github.com/jschneier/django-storages/pull/1403
+.. _#1414: https://github.com/jschneier/django-storages/pull/1414
 
 
 1.14.3 (2024-05-04)
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -309,9 +309,10 @@ class AzureStorage(BaseStorage):
         # azure expects time in UTC
         return datetime.utcnow() + timedelta(seconds=expire)
 
-    def url(self, name, expire=None, parameters=None):
+    def url(self, name, expire=None, parameters=None, mode="r"):
         name = self._get_valid_path(name)
         params = parameters or {}
+        permission = BlobSasPermissions.from_string(mode)
 
         if expire is None:
             expire = self.expiration_secs
@@ -326,7 +327,7 @@ class AzureStorage(BaseStorage):
                 name,
                 account_key=self.account_key,
                 user_delegation_key=user_delegation_key,
-                permission=BlobSasPermissions(read=True),
+                permission=permission,
                 expiry=expiry,
                 **params,
             )
--- a/tests/test_azure.py
+++ b/tests/test_azure.py
@@ -164,7 +164,9 @@ class AzureStorageTest(TestCase):
         fixed_time = make_aware(
             datetime.datetime(2016, 11, 6, 4), datetime.timezone.utc
         )
+
         with mock.patch("storages.backends.azure_storage.datetime") as d_mocked:
+            # Implicit read permission
             d_mocked.utcnow.return_value = fixed_time
             self.assertEqual(
                 self.storage.url("some blob", 100),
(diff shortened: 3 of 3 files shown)


jschneier:
Thanks for the PR, looks okay. Two comments:

1. Is there a library version minimum for the `from_string` api?
2. Can you please add or update a test case that covers this change

CorentinGoodays:
According to the [release notes](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/CHANGELOG.md#1200b4-2019-10-08), the `BlobSasPermissions.from_string` method was introduced in version `12.0.0b4`:
```
New features:
...
AccountSasPermissions, BlobSasPermissions, ContainerSasPermissions now have method from_string which takes parameters as a string.
```
This version was released a few weeks before version `12.0.0`, so from my understanding, there is no need to change the requirements of `django-storages`.

Regarding the test case, I believe the modification I'm proposing is already covered by this existing test:
https://github.com/jschneier/django-storages/blob/master/tests/test_azure.py#L157

This test currently only checks the functionality with read permissions, but there is no functional difference between a signed URL with read, write, or any other permissions. The only difference resides in the query string of the resulting URL. There's is a `sp` parameter which contains the permission string (like `r`, or `w` etc), and the `sig` parameter which contains the actual signature. Since the query string is not included in the mock implementation, I am not certain what additional aspects could be tested.

jschneier:
Sounds good on point (1).

For point (2), we should update the test or tweak it such that we confirm this is covered both for the default case and at least for the write case. I want the test to break if someone accidentally reverts parts of your patch.

CorentinGoodays:
I updated the [test](https://github.com/jschneier/django-storages/pull/1414/commits/8f894f15230810972428aa82755ef65f98ba8801) to cover both the default case and the write case.

Initially, I was hoping I could do this:
```
            generate_blob_sas_mocked.assert_called_once_with(
                self.account_name,
                self.container_name,
                "some blob",
                account_key=self.account_key,
                user_delegation_key=None,
-               permission=mock.ANY,
+               permission=BlobSasPermission(read=True),
                expiry=fixed_time + timedelta(seconds=100),
            )
...
            generate_blob_sas_mocked.assert_called_with(
                self.account_name,
                self.container_name,
                "some blob",
                account_key=self.account_key,
                user_delegation_key=None,
-               permission=mock.ANY,
+               permission__str=BlobSasPermission(write=True),
                expiry=fixed_time + timedelta(seconds=100),
            )

```
But sadly Microsoft has not implemented the `__eq__` method for the [BlobSasPermissions](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/azure/storage/blob/_models.py#L913) class, so the check is not working because it's comparing two different instances :/

I resorted to this instead:
```
            called_args, called_kwargs = generate_blob_sas_mocked.call_args
            self.assertEqual(str(called_kwargs["permission"]), "r")
...
            called_args, called_kwargs = generate_blob_sas_mocked.call_args
            self.assertEqual(str(called_kwargs["permission"]), "w")

```
Are you OK with this approach?

Would you prefer to have a separate test case for the write permission?

jschneier:
Ah that’s too bad. Is there a property on the class we can check?

CorentinGoodays:
There's a `_str` field whose value is based on the boolean values (read, write, etc) passed to the constructor. That's actually what I'm checking by converting the permission to a string here:
```
self.assertEqual(str(called_kwargs["permission"]), "w")
```
An alternative could be to dynamically add an `__eq__` method to the `BlobSasPermissions` class, like this:
```
# Dynamically add __eq__ method to BlobSasPermissions
def blob_sas_permissions_eq(self, other):
    if not isinstance(other, BlobSasPermissions):
        return NotImplemented
    return self._str == other._str

BlobSasPermissions.__eq__ = blob_sas_permissions_eq
```
Then this code would work as expected:
```
            generate_blob_sas_mocked.assert_called_with(
                self.account_name,
                self.container_name,
                "some blob",
                account_key=self.account_key,
                user_delegation_key=None,
                permission=BlobSasPermission(write=True),
                expiry=fixed_time + timedelta(seconds=100),
            )

**Comment by jschneier** [comment]
Thanks for updating, looks great!

**Comment by jschneier** [comment]
Can you just fix the lint error? Good to merge after that.

**Comment by jschneier** [comment]
Also if you wouldn't mind adding to the CHANGELOG file, that'd be appreciated!

```patch
diff --git a/CHANGELOG.rst b/CHANGELOG.rst
index 6cdc419..0b0085f 100644
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,11 +15,13 @@ Azure
 -----
 
 - Fix ``collectstatic --clear`` (`#1403`_)
+- Add ``mode`` kwarg to ``.url()`` to support creation of signed URLs for upload (`#1414`_)
 
 .. _#1399: https://github.com/jschneier/django-storages/pull/1399
 .. _#1381: https://github.com/jschneier/django-storages/pull/1381
 .. _#1402: https://github.com/jschneier/django-storages/pull/1402
 .. _#1403: https://github.com/jschneier/django-storages/pull/1403
+.. _#1414: https://github.com/jschneier/django-storages/pull/1414
 ```

**Comment by CorentinGoodays** [comment]
All right, I updated the changelog and addressed the lint error (though I couldn't reproduce it using tox, only when running Black 23.3.0 manually).

I'm not clear whether you're satisfied with this way of doing:
```
called_args, called_kwargs = generate_blob_sas_mocked.call_args
self.assertEqual(str(called_kwargs["permission"]), "w")
```
or if you would prefer adding the `__eq__` method to `BlobSasPermissions` like this?
```
# Dynamically add __eq__ method to BlobSasPermissions
def blob_sas_permissions_eq(self, other):
    if not isinstance(other, BlobSasPermissions):
        return NotImplemented
    return self._str == other._str

BlobSasPermissions.__eq__ = blob_sas_permissions_eq

...
            generate_blob_sas_mocked.assert_called_with(
                self.account_name,
                self.container_name,
                "some blob",
                account_key=self.account_key,
                user_delegation_key=None,
                permission=BlobSasPermission(write=True),
                expiry=fixed_time + timedelta(seconds=100),
            )
```
I think the second approach is nice as it makes the test case more readable. Additionally, if Microsoft ever implements the `__eq__` method, we can easily remove our custom implementation.

Please let me know which approach you prefer.

PR #1417, opened 2024-07-05
Caches Cloudfront Signers

scuml:
This adds a class-level cache layer in front of cloudfront signers.  

Was debugging slowness on a django repo with 50 s3 storage fields, and the expensive `load_pem_private_key` was called for every field implementing the S3Storage class, even though the id/key was the same for each field.  This added a full 3 seconds to server start/restart time.

Caching seems desirable for all users of the library, will provide an instant speed benefit for current users, and will keep django apps snappy and responsive no matter how many filefields they add to their django project.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ S3
 - Pull ``AWS_SESSION_TOKEN`` from the environment (`#1399`_)
 - Fix newline handling for text mode files (`#1381`_)
 - Do not sign URLs when ``querystring_auth=False`` e.g public buckets or static files (`#1402`_)
+- Cache CloudFront Signers (`#1417`_)
 
 Azure
 -----
@@ -22,6 +23,7 @@ Azure
 .. _#1402: https://github.com/jschneier/django-storages/pull/1402
 .. _#1403: https://github.com/jschneier/django-storages/pull/1403
 .. _#1414: https://github.com/jschneier/django-storages/pull/1414
+.. _#1417: https://github.com/jschneier/django-storages/pull/1417
 
 
 1.14.3 (2024-05-04)
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -376,7 +376,9 @@ class S3Storage(CompressStorageMixin, BaseStorage):
     def get_cloudfront_signer(self, key_id, key):
         cache_key = f"{key_id}:{key}"
         if cache_key not in self.__class__._signers:
-            self.__class__._signers[cache_key] = _cloud_front_signer_from_pem(key_id, key)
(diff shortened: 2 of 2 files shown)


jschneier:
Thanks this makes sense. Should this be removed in `__getstate__` and `__setstate__`?

scuml:
Good idea to check that:

Since it is declared and always treated as a class var and not an instance var, it never shows up in __dict__, so it would not need to be handled in `__getstate__` and `__setstate__`

```python
>>> class A():
...     class_var = None
...     instance_var = None
...     def __init__(self):
...             self.__class__.class_var = 1
...             self.instance_var = 2
...             print(self.__dict__)
...
>>> A()
{'instance_var': 2}
```

jschneier:
Doesn’t line 379 reference as an instance variable?

scuml:
Correct - remedied.

jschneier:
Just the lint failure and then good to merge!

scuml:
Ah - yes, line length is why I removed __class__ on that line during development.
Linted and done.

jschneier:
Huh well that's a lint error I haven't seen before.

scuml:
Python typing was a mistake.... i think this will fix it.

scuml:
Looks like that one throws a number of false positives. 
https://github.com/astral-sh/ruff/issues/5243

jschneier:
> Python typing was a mistake....

Amen.

jschneier:
This is what it wants (verified locally)

```patch
diff --git a/storages/backends/s3.py b/storages/backends/s3.py
index 98b6e94..76dec89 100644
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -4,6 +4,7 @@ import os
 import posixpath
 import tempfile
 import threading
+import typing
 import warnings
 from datetime import datetime
 from datetime import timedelta
@@ -30,8 +31,6 @@ from storages.utils import safe_join
 from storages.utils import setting
 from storages.utils import to_bytes
 
-from typing import Dict
-
 try:
     import boto3.session
     import botocore
@@ -313,7 +312,7 @@ class S3Storage(CompressStorageMixin, BaseStorage):
     # settings/args are ignored.
     config = None
 
-    _signers: Dict[str, str] = {}
+    _signers: typing.ClassVar = {}
 
     def __init__(self, **settings):
         omitted = object()
```

jschneier:
Alternatively we could update it to ignore the error via a comment. Either is fine for me as I don't find this annotation very helpful.

scuml:
Thanks for merging.  My leaning would be to ignore as the classvar type **a)** conveys nothing of value to the user and in a way obfuscates the fact it's a simple dict, and **b)** would avoid confusion later with future contributors wondering why this single type exists, and if they are now obligated to litter the rest of a very readable codebase with types.

PR #1422, opened 2024-07-06
[general] Write overwriting files in terms of .exists()

jschneier:
Follow upstream and write the file overwriting handling in terms of `exists()`.

Reference: https://github.com/django/django/commit/0b33a3abc2ca7d68a24f6d0772bc2b9fa603744e#diff-63fca472a0ca8c04505cb4ec9c132663ac71fc964110afc8ae9218416f37821cR195-R202

### DIFF
--- a/storages/backends/azure_storage.py
+++ b/storages/backends/azure_storage.py
@@ -19,7 +19,6 @@ from django.utils.deconstruct import deconstructible
 
 from storages.base import BaseStorage
 from storages.utils import clean_name
-from storages.utils import get_available_overwrite_name
 from storages.utils import safe_join
 from storages.utils import setting
 from storages.utils import to_bytes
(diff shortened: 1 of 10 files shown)


mikicz:
Btw found it very confusing that this change was not in the changelog!

CorentinGoodays:
I noticed the change where exists() now always returns False when allow_overwrite=True. I’m curious about the rationale behind this decision, as I often rely on the exists() method for troubleshooting storage issues to check whether a blob already exists.

With this new behavior, how would you recommend handling such cases when allow_overwrite=True? Is there an alternative approach you had in mind for checking the existence of files under this setting?

Thanks in advance for the clarification!

stuarta0:
@CorentinGoodays looks like the behaviour should be walked back in django-storages. See:

- https://github.com/jschneier/django-storages/issues/1430
- https://github.com/jschneier/django-storages/issues/1440#issuecomment-2305267349
- https://github.com/django/django/commit/8d6a20b656ff3fa18e36954668a44a831c2f6ddd

jschneier:
Indeed, the intent is to walk back the behavior which is why no new version has been released. If someone wants to submit a patch that fixes this for all versions and doesn't introduce a security vulnerability (following Django's lead most likely) then I would happily merge.

PR #1424, opened 2024-07-06
[ftp] Add FTP_ALLOW_OVERWRITE setting

### LINKED ISSUE #1360: Avoid Overwriting files when uploading with django-ckeditor to ftp server -> possible fix
Hi, 

files that are uploaded in the django-ckeditor5 to a ftp server are overwritten if they share the same name. i added some code to fix this. i just wanted to leave this here in case anyone else has the same problem. i am just an amateur and not really experienced with this whole "pull request" stuff.

So here is a possible fix:

I changed the original _put_file function in ftp.py to:

    from django.utils.crypto import get_random_string
    def _put_file(self, name, content) -> str:
        # Connection must be open!
        try:
            self._mkremdirs(os.path.dirname(name))
            pwd = self._connection.pwd()
            self._connection.cwd(os.path.dirname(name))

            # Rename the new file before saving if there is already a file with an
            # identical name in the directory
            files_in_dir = self.listdir(pwd)[1]
            for file in files_in_dir:
                if file == name:
                    while file == name:
                        splitted_file_str = os.path.splitext(name)
                        file_root = splitted_file_str[0]
                        file_ext = splitted_file_str[1]
                        name = "%s_%s%s" % (file_root, get_random_string(7), file_ext)**

            # Store the file
            self._connection.storbinary(
                "STOR " + os.path.basename(name),
                content.file,
                content.DEFAULT_CHUNK_SIZE,
            )
            self._connection.cwd(pwd)

            return name

....

i also changed the _save func of ftp.py to:

        content.open()
        self._start_connection()
        name = self._put_file(name, content)
        content.close()
        return name

that's it. maybe this is useful for someone.

jschneier:
Closes #1360

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,7 @@ FTP
 ---
 
 - Conform to ``BaseStorage`` interface (`#1423`_)
+- Add ``FTP_ALLOW_OVERWRITE`` setting (`#1424`_)
 
 .. _#1399: https://github.com/jschneier/django-storages/pull/1399
 .. _#1381: https://github.com/jschneier/django-storages/pull/1381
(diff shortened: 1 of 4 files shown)


th3hamm0r:
@jschneier I'm just curious, but shouldn't this be changed to the method overriding `get_available_name()` too, similar to https://github.com/jschneier/django-storages/pull/1484 ?
I think it currently has the same issue as https://github.com/jschneier/django-storages/issues/1430 (?)

PR #1427, opened 2024-07-08
Fix: Generate Signed Urls through a service account by providing service_account_email and access_token

### LINKED ISSUE #941: signed urls do not work in Cloud Run with django-storages

yohannes15:
fixes #941 

- Adds `sa_email` setting in order to set the service_account_email used for signing URL if service_account key file not available in the env. Compute services (Cloud Run, Cloud Function, App Engine, Compute Engine ...) don't have access to the key file but get access tokens through the metadata server. Passing service_account_email and access_token will cause generate_signed_url to use the IAM API to signBlob using the provided service account email and token.

### DIFF
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -57,7 +57,7 @@ In most cases, the default service accounts are not sufficient to read/write and
 present when generating signed urls. The service account private key is unavailable when running on a compute service.
 Compute Services (App Engine, Cloud Run, Cloud Functions, Compute Engine...) fetch `access tokens from the metadata server <https://cloud.google.com/docs/authentication/application-default-credentials>`__ .
 These services do not have access to the service account private key. This means that when trying to sign data in these services,
-you **MUST** use one of the Cloud IAM sign functions (SignBlob, SignJwt) to sign data and directly signing data isn't possible by any means.
+you **MUST** use Cloud IAM sign function (SignBlob) to sign data and directly signing data isn't possible by any means.
 
 Luckily this can be worked around by passing `service_account_email` and `access_token` to the generate_signed_url function.
 When both of those args are provided, generate_signed_url will use the IAM SignBlob API to sign the url and no private key file is needed.
@@ -233,7 +233,7 @@ Settings
 
   default: ``False``
 
-  Signing urls requires a service account key file to be present in the env or IAM SignBlob/JWT API call
+  Signing urls requires a service account key file to be present in the env or IAM SignBlob API call
   through a service account email and access_token. Certain GCP services (ex: Compute services) don't have access to the key file in the env.
   This setting needs to be `True` when running on such services as they fetch access tokens from metadata server instead of having key files
   If using `v4` of generate_signed_url, `google-cloud-storage>=v1.36.1 <https://github.com/googleapis/python-storage/releases/tag/v1.36.1>`_ is required .
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -59,6 +59,23 @@ For development use cases, or other instances outside Google infrastructure:
 #. Ensure the key is mounted/available to your running Django app.
 #. Set an environment variable of GOOGLE_APPLICATION_CREDENTIALS to the path of the json file.
 
+**Note Regarding Authentication**
+
+There is currently a limitation in the GCS client for Python which by default requires a service account private key file to be
+present when generating signed urls. **This is important to realize**. The service account private key is unavailable when running on a compute service.
+Compute Services (App Engine, Cloud Run, Cloud Functions, Compute Engine...) fetch access tokens from a metadata service.
+Those services do not have access to the service account private key. That means you must use one of the IAM sign functions (SignBlob, SignJwt)
+to have Google sign using their managed private key. This means, you can't directly sign data. You must use the Cloud IAM API to do signing.
+
+Luckily this can be worked around by passing `service_account_email` and `access_token` to the generate_signed_url function.
+When both of those args are provided, generate_signed_url will use the IAM service SignBlob API to sign the url and no private key file is needed.
+
+Google also now recommends avoiding service account json key files as they are insecure, risky and hard to manage. This avoids the need for that
+when developing locally.
+
+`GS_SA_EMAIL` will be what is what provided to generate_signed_url param: service_account_email. Note, this service account will need credentials to
+sign and download/upload files as necessary. Read more `here <https://cloud.google.com/storage/docs/access-control/signing-urls-with-helpers>`__
+
 Alternatively, you can use the setting ``credentials`` or ``GS_CREDENTIALS`` as described below.
 
 
@@ -219,3 +236,14 @@ Settings
   It supports `timedelta`, `datetime`, or `integer` seconds since epoch time.
 
   Note: The maximum value for this option is 7 days (604800 seconds) in version `v4` (See this `Github issue  <https://github.com/googleapis/python-storage/issues/456#issuecomment-856884993>`_)
+
+``sa_email`` or ``GS_SA_EMAIL``
+
+  default: ``''``
+
+  This is the service account email to be used for signing the url. Signing urls either requires a service account key file to be present in the env or IAM API call.
+  Compute services (App Engine, Cloud Run, Cloud Functions, Compute Engine...) for example don't have access to the key file in the env. Providing, sa_email, will use
+  the IAM API in order to sign the URL thus avoiding the need for a private service account json key file.
+
+  As above please note that, Default Google Compute Engine (GCE) Service accounts are
+  `unable to sign urls <https://googlecloudplatform.github.io/google-cloud-python/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url>`_.
--- a/docs/backends/gcloud.rst
+++ b/docs/backends/gcloud.rst
@@ -68,7 +68,7 @@ Those services do not have access to the service account private key. That means
 to have Google sign using their managed private key. This means, you can't directly sign data. You must use the Cloud IAM API to do signing.
 
 Luckily this can be worked around by passing `service_account_email` and `access_token` to the generate_signed_url function.
-When both of those args are provided, generate_signed_url will use the IAM service SignBlob API to sign the url and no private key file is needed.
+When both of those args are provided, generate_signed_url will use the IAM SignBlob API to sign the url and no private key file is needed.
 
 Google also now recommends avoiding service account json key files as they are insecure, risky and hard to manage. This avoids the need for that
 when developing locally.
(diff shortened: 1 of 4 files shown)


storages/backends/gcloud.py
```diff
@@ -141,11 +144,20 @@ def get_default_settings(self):
...
 
     @property
     def client(self):
         if self._client is None:
+            if self.project_id is None or self.credentials is None:
+                self.credentials, self.project_id = auth.default(
+                    scopes=['https://www.googleapis.com/auth/cloud-platform']
+                )
+                if not self.credentials.token_state == TokenState.FRESH:
+                    self.credentials.refresh(requests.Request())
```
jschneier, line 158: This is inside the cache check, does it need to be moved outside? Something like (e.g)

```python
    @property
    def client(self):
        if self._client is None:
            if self.project_id is None or self.credentials is None:
                self.credentials, self.project_id = auth.default(
                    scopes=['https://www.googleapis.com/auth/cloud-platform']
                )
                if not hasattr(self.credentials, "service_account_email") and self.sa_email:
                    self.credentials.service_account_email = self.sa_email
            self._client = Client(project=self.project_id, credentials=self.credentials)

        if self.credentials and self.credentials.token_state != TokenState.FRESH:
            self.credentials.refresh(requests.Request())
        return self._client
```

storages/backends/gcloud.py
```diff
@@ -323,12 +335,15 @@ def url(self, name, parameters=None):
                 quoted_name=_quote(name, safe=b"/~"),
             )
         else:
+            params = parameters or {}
             default_params = {
                 "bucket_bound_hostname": self.custom_endpoint,
                 "expiration": self.expiration,
-                "version": "v4",
+                "version": "v4"
```
jschneier, line 342: lint will complain here.

jschneier (changes requested):
One main comment about token freshness.

We also should add at least one test.

The note about authentication is helpful but I'm wondering if it needs to be more deeply integrated into the docs rather than a separate note.

storages/backends/gcloud.py
yohannes15: adjusted

storages/backends/gcloud.py
yohannes15: adjusted for this as well

yohannes15:
I adjusted some existing tests that were failing due to error Default Creds missing by mocking _client. I will do some additional tests for the feature add plus a credentials test when I get time. The test you set up `test_custom_endpoint_with_parameters` is now passing as well.

jschneier:
Curious why you've decided to go in the new direction? The last set of changes seemed easier to understand but I am likely missing some context.

yohannes15:
@jschneier the previous set of changes was easier to understand but we would be forcing all envs that had the service account email to start signing using the IAM Blob API. That API has a rate limit. Thats why I was thinking maybe it is better to make this a setting that has to be provided so we don't break any large service that is potentially has a service account email in the env but is signing with a private key file. Does that make sense? This might be helpful. 
image
https://stackoverflow.com/questions/65169199/google-cloud-storage-signed-urls-is-there-an-upper-limit-on-the-number-of-is
https://cloud.google.com/iam/quotas

yohannes15:
image
This is the code that decides it. It is in the generate_signed_url functions. We shoud probably not pass service_account and email to the function at all times. We probably want to make it a setting (iam_blob_sign (default=False)) and require service_account_email to be provided through the adc or `sa_email`. If service account email missing and iam_blob_sign is True, we throw a attribute error because that doesn't make sense.

jschneier:
Thanks, that makes sense, need to digest a bit.

storages/backends/gcloud.py
```diff
@@ -141,12 +144,22 @@ def get_default_settings(self):
...
     @property
     def client(self):
         if self._client is None:
+            if self.project_id is None or self.credentials is None:
+                self.credentials, self.project_id = auth.default(
+                    scopes=['https://www.googleapis.com/auth/cloud-platform']
+                )
             self._client = Client(project=self.project_id, credentials=self.credentials)
+        if self.credentials and self.credentials.token_state != TokenState.FRESH:
+            self.credentials.refresh(requests.Request())
```
jschneier, line 162: Have a couple questions I'm hoping you can expand upon:

1. Why don't we guard the `auth.default` behind `iam_sign_blob` rather than project_id & credentials?
2. Should it be an `and` check on project_id & credentials?
3. Similar question on guarding `credentials.refresh`

docs/backends/gcloud.rst
```diff
@@ -53,7 +53,16 @@ In most cases, the default service accounts are not sufficient to read/write and
...
 
-For development use cases, or other instances outside Google infrastructure:
+**Note:** There is currently a limitation in the GCS client for Python which by default requires a service account private key file to be
+present when generating signed urls. The service account private key is unavailable when running on a compute service.
+Compute Services (App Engine, Cloud Run, Cloud Functions, Compute Engine...) fetch `access tokens from the metadata server <https://cloud.google.com/docs/authentication/application-default-credentials>`__ .
+These services do not have access to the service account private key. This means that when trying to sign data in these services,
+you **MUST** use Cloud IAM sign function (SignBlob) to sign data and directly signing data isn't possible by any means.
+
+Luckily this can be worked around by passing `service_account_email` and `access_token` to the generate_signed_url function.
+When both of those args are provided, generate_signed_url will use the IAM SignBlob API to sign the url and no private key file is needed.
```
jschneier, line 63: Can you specify which settings need to be set for the storage to use this path.

jschneier:
I agree that this direction looks good, just have a few more questions before we merge it in.

docs/backends/gcloud.rst
yohannes15: Just did

storages/backends/gcloud.py
yohannes15: Yeah I have reverted the changes on the Client property and made the auth default use under the iam_sign_blob condition

storages/backends/gcloud.py
```diff
@@ -330,8 +337,37 @@ def url(self, name, parameters=None):
...
+                default_params["access_token"] = access_token
+
             for key, value in default_params.items():
                 if value and key not in params:
                     params[key] = value
 
             return blob.generate_signed_url(**params)
+
+    def _get_iam_sign_blob_params(self):
+        credentials, _ = auth.default(
```
jschneier, line 352: Should we be calling `auth.default` every time we call `url`?

yohannes15: ah I see okay. I will keep it under the Client property then but with iam_blob_sign condition? I see what you meant.

yohannes15: Just made the adjustment. Lmk what you think. I will add one or two tests around here later tonight or this weekend

jschneier:
Sounds good, definitely missing a few tests. We will bump the minimum required Google SDK library with this so you can remove the version mention.

Yohannes-B:
@jschneier I have added some tests around the new settings initialization and also tests around the generate_signed_url calls with this new branch of code

yohannes15:
@jschneier I have added some tests around the new settings initialization and also tests around the generate_signed_url calls with this new branch of code

jschneier:
Thanks, the code for this looks very good. I think we should reword the authentication docs a bit before merging. My concern is that there is a very large blob of text discussing the limitation and for the mitigation the django-storages settings is mixed in with discussion of kwargs for `generate_signed_url`. I think most users won't read or untangle the details needed to get the project working.

Can you explain here what the requirements are, it sounds like setting `iam_sign_blob` isn't always needed? An example of how we might reword this is to add a `4.` to the list of things to do with a conditional `if` in front of it. Then the reasoning could go into some kind of `.. note` or `.. info` block?

Given that there are now multiple ways to authenticate it also might make sense to make sub-sections so the delineation is more clear.

yohannes15:
@jschneier I have made the changes requested. Let me know what you think

yohannes15:
Just checking to remind you like you said in the README : ). No worries if busy

rwblokzijl:
@jschneier would you mind taking a look at this? Getting this working would be lovely :)

jschneier (approved):

jschneier:
These changes look good. I will be happy to merge if you rebase, there are also some minor nits to language in the documentation I can take a pass at if you prefer. For example:

`if storing private key file isn't viable and would rather generate signed urls using an API`

Is I think missing a few words.

yohannes15:
> These changes look good. I will be happy to merge if you rebase, there are also some minor nits to language in the documentation I can take a pass at if you prefer. For example:
> 
> `if storing private key file isn't viable and would rather generate signed urls using an API`
> 
> Is I think missing a few words.

@jschneier Thank you for looking at this. I just rebased. Feel free to change the verbiage as you wish.

Yohannes-B:
@jschneier hey, question. I see the documentation https://django-storages.readthedocs.io/en/latest/backends/gcloud.html#configuration-settings referring to the new settings we added as part of the feature, but I don't see the code in version 1.14.5

Aurora-at-Othersphere:
> @jschneier hey, question. I see the documentation https://django-storages.readthedocs.io/en/latest/backends/gcloud.html#configuration-settings referring to the new settings we added as part of the feature, but I don't see the code in version 1.14.5. Just a call out

This PR was merged _after_ the 1.14.5 release, however the confusing part is the docs site by default displays 'latest' (i.e. the most recent commit on the master branch) instead of the docs for the latest _release_ (which should be 1.14.5). This means it can get into a state where you are reading documentation with unreleased features.

I have been confused by this for two days, and FWIW your comment finally unlocked for me what was wrong.

robbie-portia:
Thanks so much for sorting this issue - I've been running into this too! Do we know when the next release will be that includes this fix?

jschneier:
I've been waiting to see if anything hugely important pops up but since it hasn't, I've gone ahead and released.

robbie-portia:
Amazing, thank you @jschneier!

PR #1432, opened 2024-07-11
[azure] Add AZURE_CLIENT_OPTIONS setting

daviddavis:
This allows users to customize how many times they want django-storages to retry failed requests to Azure Storage.

https://learn.microsoft.com/en-us/azure/storage/blobs/storage-retry-policy-python#use-the-default-retry-policy

https://learn.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.exponentialretry?view=azure-python#parameters

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,19 @@
 django-storages CHANGELOG
 =========================
 
+X.YY.Z (UNRELEASED)
+*******************
+
+Azure
+-----
+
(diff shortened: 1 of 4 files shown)


jschneier:
Thanks for the PR. Have seen this one before! Can you please add a setting like `AZURE_CLIENT_OPTIONS` so that we don’t end up needing a separate setting for each.

Them we will deprecate the API_ VERSION one.

daviddavis:
@jschneier thanks for the feedback. I've updated the PR.

PR #1446, opened 2024-09-04
Update CHANGELOG to reflect support for Django 5.1

kimihito:
I forgot to update the CHANGELOG in #1444, my apologies. This pull request addresses that oversight.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ django-storages CHANGELOG
 X.YY.Z (UNRELEASED)
 *******************
 
+General
+-------
+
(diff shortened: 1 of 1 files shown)


foarsitter:
According to the PR you made, there where no changes required to support 5.1? Then I would suggest to rephrase the wording so It comes clear that the support is formally and earlier versions may also support Django 5.1.

mgrdcm:
@foarsitter the changes to actually support 5.1 are in #1444 this is just adding an entry to the changeling.

jschneier:
Apologies, I did a pass when I released the most recent version and forgot this PR was opened. Thanks for creating though, always appreciate someone reducing maintainer burden!

PR #1451, opened 2024-09-17
[s3] Pass object parameters to head_object in `exists`

czosel:
When using server-side-encryption with customer-provided keys (SSE-C), the `head_object` call needs the SSE-C key and algorithm - otherwise the call will fail with `botocore.exceptions.ClientError: An error occurred (400) when calling the HeadObject operation: Bad Request` if the file already exists.

From the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html):
> If you encrypt an object by using server-side encryption with customer-provided encryption keys (SSE-C) when you store the object in Amazon S3, then when you retrieve the metadata from the object, you must use the following headers to provide the encryption key for the server to be able to retrieve the object's metadata. The headers are:
> - x-amz-server-side-encryption-customer-algorithm
> - x-amz-server-side-encryption-customer-key
> - x-amz-server-side-encryption-customer-key-MD5

Since `head_object` is similar to `get`, we're reusing the existing logic to infer and filter the needed parameters.

Thank you in advance for your feedback!

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -581,8 +581,11 @@ class S3Storage(CompressStorageMixin, BaseStorage):
 
     def exists(self, name):
         name = self._normalize_name(clean_name(name))
+        params = _filter_download_params(self.get_object_parameters(name))
         try:
-            self.connection.meta.client.head_object(Bucket=self.bucket_name, Key=name)
+            self.connection.meta.client.head_object(
+                Bucket=self.bucket_name, Key=name, **params
+            )
             return True
         except ClientError as err:
             if err.response["ResponseMetadata"]["HTTPStatusCode"] == 404:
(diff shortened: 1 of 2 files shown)


jschneier:
Thanks for this, it looks good. I'll merge as-is, just needs a rebase or similar to pick up CI. I would do this myself but I don't have permissions to push to your branch.

czosel:
@jschneier Awesome, thanks for the review! rebased.

jschneier:
Thanks, just need to fix lint and good-to-go.

czosel:
@jschneier done, sorry I missed that earlier!

coltonpark5:
@jschneier Would it be possible to get a release for this?

PR #1455, opened 2024-09-29
Update digital-ocean-spaces.rst

joedevon:
The vars listed didn't work for Digital Ocean so I wrote the values that worked for me and removed the following from the docs:

region_name
endpoint_url
access_key
secret_key


jschneier:
Sorry but the point is that the env var works but the other version works as well, you can see both options identified in the S3 docs which is referenced from the Digital Ocean docs, no?

PR #1457, opened 2024-10-02
[gcloud] fix double decompression when using gzip

### LINKED ISSUE #1456: Fix GCloud GZIP
Need to remove https://github.com/jschneier/django-storages/blob/master/storages/backends/gcloud.py#L67-L68 per https://stackoverflow.com/questions/67744979/how-to-prevent-gcs-from-automatically-decompressing-objects-when-using-python-sd. Supersedes #1454.

thuibr:
The `self.blob.download_to_file` already performs decompression per this [StackOverflow](https://stackoverflow.com/questions/67744979/how-to-prevent-gcs-from-automatically-decompressing-objects-when-using-python-sd) post. Rather than attempting to decompress the gzip file a second time and receiving an error, allow `self.blob.download_to_file` to decompress it.

An alternative to this would be to use the `raw_download=True` kwarg for `self.blob.download_to_file` if that is desirable.

This closes #1456.

### DIFF
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -62,10 +62,9 @@ class GoogleCloudFile(CompressedFileMixin, File):
             )
             if "r" in self._mode:
                 self._is_dirty = False
+                # This automatically decompresses the file
                 self.blob.download_to_file(self._file, checksum="crc32c")
                 self._file.seek(0)
-            if self._storage.gzip and self.blob.content_encoding == "gzip":
-                self._file = self._decompress_file(mode=self._mode, file=self._file)
         return self._file
 
     def _set_file(self, value):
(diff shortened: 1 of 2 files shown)


thuibr:
@jschneier what are your thoughts on this change as opposed to using `raw_download=True` in the client per https://stackoverflow.com/questions/67744979/how-to-prevent-gcs-from-automatically-decompressing-objects-when-using-python-sd ?

thuibr:
@jschneier I added a test. How do I run the tests in CI?

thuibr:
@jschneier can you merge this?

thuibr:
@jschneier can you look at this?

thuibr:
@jschneier what do I need to do to get this PR looked at and approved?

jschneier:
@thuibr can you confirm this fixes your issue?

thuibr:
@jschneier, I double checked just now that this fixes the issue. I also made a change to the test that I think fixes CI.

thuibr:
@jschneier do you need to run the CI?

markesha:
@jschneier Do you mind making a release for this change?

PR #1473, opened 2024-11-15
Fix #1469 Use a different checksum calculation method to run in FIPS env

### LINKED ISSUE #1469: Use a different checksum calculation method to run in FIPS enabled environment with python3.10+
Python 3.10 and later versions rely on OpenSSL 1.1.1 or newer, which includes FIPS-compliance checks.

MD5 is not an approved algorithm in FIPS mode, so attempting to instantiate [self.blob.download_to_file(self._file)](https://github.com/jschneier/django-storages/blob/master/storages/backends/gcloud.py#L65) will fail when the system is running in FIPS mode.

gcloud Python storage SDK provides a different method [_crc32c_](https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/blob.py#L1167) for calculating checksum, which can be used instead:

```
class GoogleCloudFile(CompressedFileMixin, File):
def _get_file(self):
    if self._file is None:
        self._file = SpooledTemporaryFile(
            max_size=self._storage.max_memory_size,
            suffix=".GSStorageFile",
            dir=setting("FILE_UPLOAD_TEMP_DIR"),
        )
        if "r" in self._mode:
            self._is_dirty = False
            self.blob.download_to_file(self._file, checksum="crc32c")
            self._file.seek(0)
        if self._storage.gzip and self.blob.content_encoding == "gzip":
            self._file = self._decompress_file(mode=self._mode, file=self._file)
    return self._file
```

markesha:
Python 3.10 and later versions rely on OpenSSL 1.1.1 or newer, which includes FIPS-compliance checks.

MD5 is not an approved algorithm in FIPS mode, so attempting to instantiate self.blob.download_to_file(self._file) will fail when the system is running in FIPS mode.

The change configures the `download_to_file` function to use an alternative algorithm provided by gcloud storage SDK - 'crc32c' - for checksum calculation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ dropbox = [
   "dropbox>=7.2.1",
 ]
 google = [
-  "google-cloud-storage>=1.27",
+  "google-cloud-storage>=1.32",
 ]
 libcloud = [
   "apache-libcloud",
--- a/storages/backends/gcloud.py
+++ b/storages/backends/gcloud.py
@@ -62,7 +62,7 @@ class GoogleCloudFile(CompressedFileMixin, File):
             )
             if "r" in self._mode:
                 self._is_dirty = False
-                self.blob.download_to_file(self._file)
+                self.blob.download_to_file(self._file, checksum="crc32c")
                 self._file.seek(0)
             if self._storage.gzip and self.blob.content_encoding == "gzip":
                 self._file = self._decompress_file(mode=self._mode, file=self._file)
--- a/tests/test_gcloud.py
+++ b/tests/test_gcloud.py
@@ -42,7 +42,7 @@ class GCloudStorageTests(GCloudTestCase):
                 self.filename, chunk_size=None
             )
 
-            f.blob.download_to_file = lambda tmpfile: tmpfile.write(data)
+            f.blob.download_to_file = lambda tmpfile, **kwargs: tmpfile.write(data)
             self.assertEqual(f.read(), data)
 
     def test_open_read_num_bytes(self):
@@ -55,7 +55,7 @@ class GCloudStorageTests(GCloudTestCase):
                 self.filename, chunk_size=None
             )
 
-            f.blob.download_to_file = lambda tmpfile: tmpfile.write(data)
+            f.blob.download_to_file = lambda tmpfile, **kwargs: tmpfile.write(data)
             self.assertEqual(f.read(num_bytes), data[0:num_bytes])
 
     def test_open_read_nonexistent(self):


jschneier:
Just making sure, this change is backwards compatible, right?

markesha:
@jschneier the change depends on a specific version of the google storage lib, that I'm locking with the change. All the test python versions were already using the latest one 2.18.2, e.g. see 3.7 logs:

```py3.7-django3.2: apache-libcloud==3.8.0,asgiref==3.7.2,azure-core==1.30.1,azure-storage-blob==12.19.1,bcrypt==4.2.0,boto3==1.33.13,botocore==1.33.13,cachetools==5.5.0,certifi==2024.8.30,cffi==1.15.1,charset-normalizer==3.4.0,coverage==7.2.7,cryptography==43.0.3,Django==3.2.25,django-storages @ file:///home/runner/work/django-storages/django-storages/.tox/.tmp/package/1/django-storages-1.14.4.tar.gz#sha256=fd2920db03cfb63bf4d6f600ee0e6bdb668931505a261c3f0107cda7e5947bc8,dropbox==12.0.2,exceptiongroup==1.2.2,google-api-core==2.23.0,google-auth==2.36.0,google-cloud-core==2.4.1,google-cloud-storage==**2.18.2**,```

I guess it should be compatible.

**Comment by markesha** [comment]
@jschneier the failing test `tests/test_gcloud.py::GoogleCloudGzipClientTests::test_storage_save_gzipped` appears unrelated to the change and also fails on earlier revisions.

```
 AssertionError: expected call not found.
 Expected: upload_from_file(<ANY>, rewind=True, retry=<google.api_core.retry.retry_unary.Retry object at 0x10dad5300>, size=11, predefined_acl=None, content_type='application/javascript')
 Actual: upload_from_file(<ContentFile: Raw content>, rewind=True, retry=<google.api_core.retry.retry_unary.Retry object at 0x10dad5300>, size=11, predefined_acl=None, content_type='text/javascript')
```

**Comment by jschneier** [comment]
@markesha any idea what's going on here? There are no failures now but the tests were never "fixed".

File: pyproject.toml [hunk 1]
```diff
@@ -55,7 +55,7 @@ dropbox = [
   "dropbox>=7.2.1",
 ]
 google = [
-  "google-cloud-storage>=1.27",
+  "google-cloud-storage>=1.31",
```
jschneier, line 58: Can you actually change this to be `>=1.32`. Turns out we already require that due to the retry import.

jschneier (changes requested):
Looks good although I don't understand the test reality.

markesha:
@jschneier I'm not sure what's happening - it still fails locally for me

jschneier:
> @jschneier I'm not sure what's happening - it still fails locally for me

Well, in #1476 I just change it to use CSS which should fix whatever weird mimetype problem we are having.

jschneier (approved):

markesha:
@jschneier Do you mind making a release for this change?

PR #1483, opened 2025-02-06
Defend against settings.AWS_S3_URL_PROTOCOL set to None

magul:
We had a situation, that somebody put in our settings:
```
AWS_S3_URL_PROTOCOL = os.environ.get("AWS_S3_URL_PROTOCOL")
```
here it was falsely assumed, that not setting this environment variable will make `django-storages` to fallback to default `https:` (which didn't happened and we want to fix it in this pull request).

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -429,7 +429,7 @@ class S3Storage(CompressStorageMixin, BaseStorage):
                     "image/svg+xml",
                 ),
             ),
-            "url_protocol": setting("AWS_S3_URL_PROTOCOL", "https:"),
+            "url_protocol": setting("AWS_S3_URL_PROTOCOL") or "https:",
             "endpoint_url": setting("AWS_S3_ENDPOINT_URL"),
             "proxies": setting("AWS_S3_PROXIES"),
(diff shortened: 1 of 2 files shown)


jschneier:
This test passes though, can you be more clear on what happened that you didn't expect.

```python
    def test_url_protocol(self):
        storage = s3.S3Storage()
        self.assertEqual(storage.url_protocol, "https:")
```

magul:
Hi @jschneier 

I've just spotted I made a mistake in descriptio (I've already fixed it) - it always was about `AWS_S3_URL_PROTOCOL`, not `AWS_S3_CUSTOM_DOMAIN`.

so the problem we encounter was, that somebody added to our settings
```python
AWS_S3_URL_PROTOCOL = os.environ.get("AWS_S3_URL_PROTOCOL")
```
and it was widely (mis)understood in our project, that not having `AWS_S3_URL_PROTOCOL` set in settings and having `AWS_S3_URL_PROTOCOL` set to `None` will behave the same way.

In this pull request I make `django-storages` behave the same way regardless whether `AWS_S3_URL_PROTOCOL` is not set and set to `None`.

(I've also write 2 tests, so maybe they will clarify, what I want to achieve here).

PR #1492, opened 2025-02-26
Updated old docs link

gerbermarco:
Updated dead Microsoft docs link for configuring a connection string for an Azure storage account.


jschneier:
Thanks!

PR #1494, opened 2025-03-03
Fix leading slash in path which creates a problem for S3 when URL path has double slashes

mbocevski:
S3 throws an access denied when URL has doubleslashes in it, and having a leading slash in templates causes an access denied error. For example: https://mydjangosite.com//jquery/ajax.js would not work. 

This fix removes a leading slash from the path and can be expanded to remove all double-slashes, however I feel that removing the leading slash is enough for most common scenarios.

### DIFF
--- a/storages/utils.py
+++ b/storages/utils.py
@@ -54,6 +54,11 @@ def clean_name(name):
     if clean_name == ".":
         clean_name = ""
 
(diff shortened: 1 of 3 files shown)


mbocevski:
@jschneier can you check this one out whenever you have a chance? Thanks

PR #1495, opened 2025-04-01
Fix multithreading s3 memory leak

ddelange:
storages.s3 has a memory leak ref https://github.com/boto/boto3/issues/1670

this PR adds a single thread-safe boto3 resource per-storage instead of per-thread, and adds a 1 hour time to live cache for it such that it gets periodically recreated to avoid this memory leak.

- the memory leak on infra level (6 day view):
  image
- [memray --leaks](https://github.com/bloomberg/memray) profile consistently shows a long lived s3 client as the culprit (just a random snapshot from the entire profile):
  image


jschneier:
Thanks for opening. I don't think we can collapse UNSIGNED down like this. The reason that was written at all is that if you want to _upload_ something you need to use a signed version regardless. Maintaining a single storage API that can do both uploading and generating URLs requires hiding the complexity within the class.

storages/backends/s3.py
```diff
@@ -691,10 +674,7 @@ def url(self, name, parameters=None, expire=None, http_method=None):
         params["Bucket"] = self.bucket.name
         params["Key"] = name
 
-        connection = (
-            self.connection if self.querystring_auth else self.unsigned_connection
-        )
```
ddelange, line 696: hi @jschneier :wave:

I didn't see any point in code where `self.querystring_auth` gets altered after init, and I didn't see any other point in code where `self.unsigned_connection` is used, so this seemed like a safe refactor to me. I didn't look at parent classes though.

ddelange: ah, but the same is not true for `connection`! you're right, will adjust :+1:

ddelange: 27fa7c3

ddelange:
This is green now on [my fork](https://github.com/ddelange/django-storages/pull/1) 👍

ddelange:
this has been live since April 2nd and as you can see, we are now consistently below 50% memory 🎉 
```bash
pip install 'django-storages[s3]@https://github.com/jschneier/django-storages/archive/872a66a7f608fee9e2cbd12095d4fb791b57fb6d.zip'
```
image

storages/backends/s3.py
```diff
@@ -347,6 +349,9 @@ def __init__(self, **settings):
                 s3={"addressing_style": self.addressing_style},
                 signature_version=self.signature_version,
                 proxies=self.proxies,
+                max_pool_connections=64,  # thread-safe
+                tcp_keepalive=True,
+                retries={"max_attempts": 6, "mode": "adaptive"},
```
ddelange, line 354: fyi, these values have been empirically chosen using a c6 family EC2 machine talking directly to s3 from the AWS network using [this code](https://gist.github.com/ddelange/13b0f9da3147f3754b9e1e88c13303ba). increasing threads beyond 64 starts to become unstable and does not yield a total throughput increase.

ddelange:
Alright, this last commit took some trial and error. The test that the commit adds was essential, and prompted that change.

ddelange:
Hi @jschneier 👋 

This is ready for review

ddelange:
Hi @jschneier 👋 

This is ready for review

storages/backends/s3.py
```diff
@@ -30,6 +30,21 @@
...
+    raise ImproperlyConfigured(msg) from e
+
+try:
+    from functools import cached_property
+except ImportError:  # python_version<='3.7'
+    try:
+        from backports.cached_property import cached_property
+    except (ImportError, ModuleNotFoundError) as e:
+        msg = "Could not import backports.cached_property. Did you run 'pip install django-storages[s3]'?"  # noqa: E501
+        raise ImproperlyConfigured(msg) from e
```
ddelange, line 46: @jschneier is the plan to support python 3.7 much longer? since it's [end of life](https://devguide.python.org/versions/#unsupported-versions) for almost 2 years now

jschneier: It was supported because I had not yet dropped Django3.2 but that is now done.

storages/backends/s3.py
jschneier: Any idea what the defaults are?

People run on all kinds of hardware and I'm mostly inclined to say this should be configured by the user if they want.

jschneier:
Thanks for this PR. My main hesitation is taking a dependency on a 3rd party library, namely `cachetools` in this case. Generally speaking I prefer not to do that for all the typical reasons.

Is there no other way to solve the problem, even in a less good way?

ddelange:
Hi @jschneier 👋 

I've added a commit to remove dependency on cachetools (and cached_property). ready for review again!

storages/backends/s3.py
ddelange: iirc the default is no retries, no keepalive, 10 conns. the above is the sensible default imo. people can pass their own client_config if the need arises

ddelange:
@jschneier kind reminder :)

ddelange:
Hi @jschneier :wave:

Anything from my side still?

ddelange:
Hi @jschneier :wave:

Kind reminder on this one

ddelange:
@jschneier what can be dome to get this merged?

PR #1502, opened 2025-04-23
[ci] Update CI to use Ubuntu 22.04 for testing

kimihito:
Upgrade GitHub Actions Ubuntu version from 20.04 to 22.04

Ubuntu 20.04 (Focal Fossa) has reached end of support in GitHub Actions, so this PR updates our workflow configurations to use Ubuntu 22.04 (Jammy Jellyfish) instead.

This change ensures our CI/CD pipelines continue to run on a supported environment.

ref: https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


jschneier:
Thanks!

PR #1505, opened 2025-04-28
Drop support for Django3.2 and Django4.1

jschneier:


ddelange:
@jschneier I think you forgot to update `requires-python` in `pyproject.toml`

PR #1512, opened 2025-05-22
Upgrade to AWS ListObjectsV2

cogsea:
This line now fails in django storage (listdir()) due to: 

**An error occurred (NoSuchKey) when calling the ListObjects operation: None**

ListObjects is to be deprecated: [source](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html)

Boto3 shows the new list_objects_v2 as example: [source](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/paginators.html)

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -602,7 +602,7 @@ class S3Storage(CompressStorageMixin, BaseStorage):
 
         directories = []
         files = []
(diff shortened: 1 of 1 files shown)


jschneier:
Yeah so the reason it’s like this is that S3 Compatible service providers didn’t necessarily implement this

PR #1513, opened 2025-05-26
Make sure azure delegation key is valid when received

jgoedeke:
Time difference between system and azure can lead to not yet valid user delegation keys. This will throw an `azure.core.exceptions.HttpResponseError`. Solution can be found here: https://stackoverflow.com/a/79473336


andyp05:
see: https://github.com/jschneier/django-storages/issues/1525
might be better to set 
```
key_expiry_time = now + timedelta(days=7, minutes=-10)
```

PR #1523, opened 2025-06-25
Introduce S3SeekableFile as a S3 file wrapper

Flauschbaellchen:
Previously, whenever an S3 file was opened, it was directly downloaded into a `SpooledTemporaryFile`, independently if it was ever read or not. For files with a bigger size it reduced performance and increased traffic and expenses in cloud environments.

This commit wraps the access to the S3 file into a S3SeekableFile wrapper which reads requested bytes on demand, without loading the file completely into memory or local storage.

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -98,6 +98,54 @@ def _filter_download_params(params):
     }
 
 
+class S3SeekableFile(io.RawIOBase):
+    """
+    Wrapper around an S3 file, providing support for seeking.
+    The position of the pointer is held in memory.
+    For every read only the requested length is loaded from the S3 service.
+    """
+
+    def __init__(self, obj, ExtraArgs):
+        self.s3_object = obj
+        self.size = obj.content_length
+        self.pos = 0
+        self.ExtraArgs = ExtraArgs
+
+    def seek(self, offset, whence=io.SEEK_SET):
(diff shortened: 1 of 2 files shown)


storages/backends/s3.py
```diff
@@ -167,35 +215,24 @@ def closed(self):
...
             if "r" in self._mode:
                 self._is_dirty = False
                 params = _filter_download_params(
                     self._storage.get_object_parameters(self.name)
                 )
-                self.obj.download_fileobj(
-                    self._file, ExtraArgs=params, Config=self._storage.transfer_config
-                )
-                self._file.seek(0)
+                self._file = S3SeekableFile(self.obj, ExtraArgs=params)
```
Flauschbaellchen, line 223: Please note that `Config=self._storage.transfer_config` was removed... I'm not sure if this is needed as the method changed from `download_fileobj` to `get` as well.

Flauschbaellchen:
@jschneier May I bump this PR? I would love to hear from you what you think about this change. It would help a lot when working with large files on S3 if only a subset of data needs to be read. It also reduces the time it takes from opening the file until the first chunk can be processed by the application as it does not need to wait until the full download has been completed.

ben-xo:
following

dralley:
Related: https://github.com/jschneier/django-storages/pull/1569

PR #1536, opened 2025-08-31
Allow `querystring_auth` to be overridden on a per-call basis

brianhelba:
This preserves the existing behavior by default, but adds the ability for individual calls to `S3Storage.url` to specify a different value of `querystring_auth`.

This supports the use case of buckets with mixed public and private content.

### DIFF
--- a/storages/backends/s3.py
+++ b/storages/backends/s3.py
@@ -665,12 +665,14 @@ class S3Storage(CompressStorageMixin, BaseStorage):
         else:
             return make_naive(entry.last_modified)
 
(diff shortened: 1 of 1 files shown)


brianhelba:
I'm not sure how to write a useful test for this. The existing [`test_url_unsigned`](https://github.com/jschneier/django-storages/blob/ca89a94a7462a2423df460e7bfd5f847457042ca/tests/test_s3.py#L717-L720) would be the appropriate place to add some additional testing, but the current test structure doesn't really assert anything useful about whether signing actually occurred.

PR #1548, opened 2026-01-03
d

sarojjpathak:


nikolas:
I think we can safely close this PR, though it is an interesting proposal here.

PR #1555, opened 2026-01-27
Raise generic FileNotFoundError from Azure-specific ResourceNotFoundError for more graceful handling by Django

dericke:
Currently, a missing resource in Azure Storage will raise an Azure `ResourceNotFoundError`. Django doesn't have any handling for this exception, so the whole page crashes. By raising a `FileNotFoundError` from that exception, Django can render the rest of the page. 

I came across this problem while troubleshooting a failed migration with a Wagtail site, where a link to an image in Azure Blob Storage broke. Because the missing resource crashed related pages in Wagtail's edit interface, I couldn't do the quick resolution of removing or fixing the problem resource through the Wagtail interface. When I reproduced the missing resource on my test site using this fork, I was able to fix the missing resource problem.


dericke:
Thinking about it some more; if you agree with the direction of this PR, I should probably add tests to make sure all of the backends raise the same exception for a missing resource. LMK if that's desirable.

dericke:
I added FileNotFoundErrors for the other backends, for consistency. All backends should now raise a FileNotFoundError for a bad path.

PR #1545, opened 2025-12-22
Update supported Django/Python versions and lint fixes

jasongi:
fixes #1544 

I have also cleaned up some lint errors I was getting by:
- prefixing _ at the start of unused variables
- lowercasing characters in unicode escape (in test)
- excluding storages/backends/s3.py from PLC0415 as it has imports inside of functions.

CI passing here: https://github.com/jasongi/django-storages/actions/runs/20428821628

PR #1559, opened 2026-03-17
[ftp] Add get_modified_time

staceybeard:
## Summary
Implemented `get_modified_time` in the `FTPStorage` class, from the `Storage` interface (via `BaseStorage`).

This function returns a file's last updated timestamp as a `datetime` value.

## What's Included

- `storages/backends/ftp.py` — `get_modified_time` function with helper function `_get_file_last_modified_details` (modeled on `_get_dir_details`). This function uses [MLST](https://datatracker.ietf.org/doc/html/rfc3659#section-7) to get last modified data for a given file.

- `tests/test_ftp.py` — Unit tests for success and error cases in the new function.

## Notes

- Unit tests of naive vs timezone-aware datetimes were modeled on those written by `scjody` in `tests/test_gcloud.py`.
- Tests and linting pass when running with changes from #1545.
- From reading the specification at https://datatracker.ietf.org/doc/html/rfc3659#section-2.3, it seems that time values are always expected to be provided in UTC server-side. Because of this, this implementation assumes UTC when reading timestamps from the MLST command. This relates to concerns raised in #1233 about knowing the timezone information for different backends.

### DIFF
--- a/storages/backends/ftp.py
+++ b/storages/backends/ftp.py
@@ -15,6 +15,7 @@
 # class FTPTest(models.Model):
 #     file = models.FileField(upload_to='a/b/c/', storage=fs)
 
+import datetime
 import ftplib
 import io
 import os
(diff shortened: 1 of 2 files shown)


mschoettle:
@jschneier Any chance this can be considered to be included?

PR #1556, opened 2026-02-28
Add Vercel Blob storage backend

jlariza:
## Summary

Adds a new `VercelBlobStorage` backend for [VercelBlob](https://vercel.com/docs/storage/vercel-blob), using the official [`vercel-py`](https://github.com/vercel/vercel-py) SDK.

## What's included

- **`storages/backends/vercel_blob.py`** — full implementation of the DjangoStorage API:
  - `_open`, `_save`, `delete`, `exists`, `listdir`, `size`, `url`, `get_available_name`
  - `VercelBlobFile` wrapping `SpooledTemporaryFile` for lazy reads and buffered writes
  - Both `"public"` and `"private"` access modes
  - URL construction from the bearer token without a network round-trip, with a `head()` fallback
  - Paginated `listdir` using `mode="folded"` to surface virtual directories
  - Token resolution via Django setting → `BLOB_READ_WRITE_TOKEN` → `VERCEL_BLOB_READ_WRITE_TOKEN`
- **`tests/test_vercel_blob.py`** — 59 unit tests (fully mocked, no network calls)
- **`docs/backends/vercel-blob.rst`** — full documentation covering installation, settings, private-blob proxy pattern, URL behaviour, security notes, and known limitations
- **`pyproject.toml`** — new `vercel-blob` optional dependency group (`vercel>=0.5.0`)

## Notes

- The `vercel` package requires **Python 3.10+**. This is documented in the installation note and in the backend's docstring; the rest of django-storages continues to support Python 3.8+.
- Vercel Blob has no presigned/expiring URL support. The docs include a proxy-view pattern for serving private blobs from Django.
- The sync SDK creates a fresh HTTP client per call, so the backend is thread-safe without any additional locking.
- Claude Code was used in the coding process. However, everything was reviewed and corrected by humans when necessary.

## Checklist

- [x] Tests pass (`pytest tests/test_vercel_blob.py` — 59/59)
- [x] No type annotations (matches existing backend style)
- [x] Optional dependency declared in `pyproject.toml`
- [x] Documentation added and linked from `docs/index.rst`
- [x] Tested with Vercel directly. It properly worked for both `default` and `staticfiles` using this configuration

```python
STORAGES = {
    "default": {
        "BACKEND": "storages.backends.vercel_blob.VercelBlobStorage",
        "OPTIONS": {
            "token": "<token>",
            "location": "media",
        },
    },
    "staticfiles": {
        "BACKEND": "storages.backends.vercel_blob.VercelBlobStorage",
[...]

PR #1560, opened 2026-03-20
Add Vercel Blob storage backend

dnwpark:
Adds a storage backend for Vercel Blob (https://vercel.com/docs/storage/vercel-blob).

- Install with `pip install django-storages[vercel]` (adds `requests` dependency).
- New classes:
  - `VercelBlobStorage`: storage backend backed by the Vercel Blob REST API
  - `VercelBlobFile`: lazy read-only file access
- Configurable via Django settings (`VERCEL_BLOB_*`) or `OPTIONS` in `STORAGES`
- Support for public/private ACL, overwrite, random suffix, cache control, and location prefix

Includes unit tests and new docs.

### DIFF
--- a/docs/backends/vercel-blob.rst
+++ b/docs/backends/vercel-blob.rst
@@ -62,7 +62,7 @@ Settings
   A path prefix prepended to all stored file names, e.g. ``"media"`` or
   ``"uploads"``.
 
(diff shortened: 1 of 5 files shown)


dnwpark:
closed in favor of #1556

PR #1569, opened 2026-09-11
Add a streaming, non-buffered API for file access

### LINKED ISSUE #1568: Add an opt-in bounded streaming-read API for remote storage backends
## Problem

`Storage.open()` returns a seekable Django file object. For the django-storages S3, Azure, and GCS backends, opening a file for reading currently materializes the complete remote object into a temporary file before the caller receives data.

This is problematic for applications that must proxy authenticated downloads through their application server. A chunked response loop cannot bound memory or time-to-first-byte if its first read waits for a full download. Configuring `max_memory_size` only changes whether the complete temporary copy is held in RAM or on disk.

## Proposal

Add an optional, context-managed, forward-only streaming-read API to django-storages:

```
    with storage.open_stream(name, start=0, length=None) as stream:
        while chunk := stream.read(256 * 1024):
            do_something(chunk)
```

`open_stream()` would:

- Accept the logical storage name, an optional starting byte offset, and an optional length.
- Return a binary, forward-only reader. It need not implement seeking, fileno(), text-mode behavior, or transparent decompression.
- Avoid a full local spool; normal SDK/provider buffers are acceptable.
- Close the underlying provider response when the context exits, including error and client-disconnect paths.
- Preserve the existing Storage.open() behavior unchanged.

It would be opt-in because existing callers may depend on having a seekable file object, temporary-file backing, text mode, gzip handling, file descriptor etc. which this would not provide.

## Backend implementation direction

- S3: normalize the logical name normally, use get_object() with a Range header, pass the download-compatible subset of get_object_parameters(), and expose the returned StreamingBody.
- Azure: use download_blob(offset=..., length=...) and wrap its incremental chunks() iterator as a forward-only reader.
- GCS: use the Google client progressive/ranged blob-reading API, provided it does not buffer the complete object.

## Compatibility and tests

- Test full and ranged reads, path/location handling, backend-specific request parameters, bounded incremental reads, and cleanup after read/open failures.
- Document that open_stream() is intended for efficient proxying and large sequential downloads, while open() remains appropriate when applications need a conventional seekable file object.

(Assisted by Codex)

dralley:
Storage.open() provides a seekable API backed by buffering the file in memory or on disk. For large or very large files (e.g. 50 GB) this adds extreme overhead when buffering is not needed.

This commit provides an API for streaming files from remote storage without intermediate buffering. Offset + length is supported for range requests.

closes #1568


storages/backends/azure_storage.py
```diff
@@ -241,6 +245,50 @@ def _get_valid_path(self, name):
...
+        )
+        try:
+            yield LimitedStream(ChunkIteratorStream(downloader.chunks()), length)
+        finally:
+            # azure-storage-blob has no public downloader close method. Newer
+            # releases may gain one, so prefer it before closing its transport
+            # response as a compatibility fallback for interrupted reads.
+            close = getattr(downloader, "close", None)
+            if close is not None:
+                close()
```
dralley: This is probably unnecessary, I can drop it if desired.

dralley:
@jasongi @jschneier I know the maintainership situation is a bit strained ATM, could you at least set the CI running?  

We may vendor/fork a copy with our patches in the short-term (until the situation is sorted out) since the issue is a bit acute for us.

